PR #6, opened 2016-06-13
Gradlebuild

rspieldenner:
Start of a conversion to show off what the gradle build would be like before going whole hog.

Other option is to just wrap the make calls and use gradle just for ospackage to create the debs.


rspieldenner:
Let me know if you want to go this direction and if so I'll fix the conflicts.

zli-nflx:
Hi @rspieldenner: many of the VMAF users tend to be more familiar with make rather than Gradle. I feel a compromise is as what you have suggested, to wrap make calls and use grade for ospackage for debs.

li-zhi:
Hi @rspieldenner,

Since we last communicated, the repo has go through some substantial changes. I think it is time to resume our effort to add a CI loop to it. Also, as we talked about, can we elect the option of keeping Makefile and wrapping make calls and use grade for ospackage for debs?

Best,
Zhi

PR #15, opened 2016-08-05
Feature/adm scales

ikatsavounidisNetflix:
- Return all 4 scale values (num, den) from ADM
- Provide additional parameter to control border in ADM; currently, using default value (0.1)


anushmoorthy:
@zli-nflx Looks good to me.

zli-nflx:
Ok! Will look into it tonight!

PR #85, opened 2017-05-23
recommend pip, add ffmpeg2vmaf info

shintaroid:
I tried `apt-get` way of installing python packages: `apt-get install python-numpy` which is mentioned in https://www.scipy.org/install.html , but the version is old so I used *pip* at last. Besides, we need to install *h5c* by *pip* after all. So I believe we should just use *pip* for installing python packages.

*ffmpeg2vmaf* is very useful for easy test mp4 files, so I believe we should add that info.

By the way, there is an error which says `package tk` is missing when running *run_vmaf* after the installation. I solved that by `apt-get install python-tk`,  I am not sure if I should add that into *README.md*

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,24 +31,28 @@ It also requires a number of Python packages:
 
 You will need to install gfortran, freetype, pkg-config, freetype and hdf5 (these can't be compiled from source here):
 - on OSX: `brew install gcc freetype pkg-config homebrew/science/hdf5` (`gcc` has the `gfortran` compiler needed for scipy)
-- on linux: `sudo apt-get install -y pkg-config gfortran libhdf5-dev libfreetype6-dev liblapack-dev`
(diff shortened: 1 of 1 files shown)


shintaroid:
alright, it seems no one likes my idea

li-zhi:
Hi Su,

Sorry didn't attend the pull request earlier.

In fact for the latest codebase, ffmpeg2vmaf won't be needed. Just use
'notyuv' in place of the yuv type, and supply the non-yuv file path (for
both reference and distorted videos). You will see an error about the
ffmpeg path. Follow the instruction to resolve it, then it would work ok.

shintaroid:
Thanks for your attention. I see, so it is getting better.

By the way, I have to say that vmaf is a very nice tool, good job!

li-zhi:
Apology for providing wrong info. I was away from my computer and the memory was illusional. Indeed, ffmpeg2vmaf is the right way to go. Can you re-open the PR and I'll review it.

shintaroid:
Sorry, I accidentally deleted the fork, so I did again. Please refer to the new pull request.

PR #90, opened 2017-06-24
Library interface for vmaf

ashk43712:
Hi, this is a library interface which we created to integrate vmaf with our filter inside ffmpeg. See this source code (https://github.com/ashk43712/FFmpeg/blob/librarification/libavfilter/vf_vmaf.c) or this ffmpeg patch (https://patchwork.ffmpeg.org/patch/4103/) to know how we use it.
Library header is "libvmaf.h" which contains an API compute_vmaf(..). Here we have used threads(mutex/cond) inside ffmpeg using which vmaf pulls the frames from ffmpeg. I have created separate files (inside wrapper/libsrc/ and object files are generated in wrapper/libobj/) to use vmaf as a library and added make install/uninstall and library creation rules in Makefile.
I can add the same interface to convert your code into using the library if you like it.


wrapper/libsrc/combo_lib.c
```diff
@@ -0,0 +1,403 @@
...
+        DArray *vif_den_scale3_array,
+        DArray *vif_array,
+        DArray *psnr_array,
+        DArray *ssim_array,
+        DArray *ms_ssim_array,
+        char *errmsg
+        )
+{
+    double score = 0;
+    double scores[4*2];
```
anushmoorthy: Not sure if we should hard-code array sizes.

wrapper/libsrc/combo_lib.c
```diff
@@ -0,0 +1,403 @@
+/**
```
anushmoorthy: Not sure if we need to copy-paste a whole bunch of code just so that the calling interface can be changed. The typical way to do this is to create a new interface and link to the older one internally without re-writing code.

wrapper/libsrc/vmaf_lib.cpp
```diff
@@ -0,0 +1,779 @@
+/**
```
anushmoorthy: Similar comment to the other file. Much of the this code is a copy-paste. I would split this to a call to existing interface and then population of the new data structure.

wrapper/libsrc/vmaf_lib.cpp
```diff
@@ -0,0 +1,779 @@
...
+    free_array(&ms_ssim_array);
+
+    return result;
+}
+
+
+// static const char VMAFOSS_XML_VERSION[] = "0.3.1";
+//static const char VMAFOSS_XML_VERSION[] = "0.3.2"; // add vif, ssim and ms-ssim scores
+//static const char VMAFOSS_XML_VERSION[] = "0.3.3"; // fix slopes and intercepts to match nflxtrain_vmafv3a.pkl
+static const char VMAFOSS_XML_VERSION[] = "0.3.2"; // fix slopes and intercepts to match nflxall_vmafv4.pkl
```
anushmoorthy: The copy-paste making things dangerous here...the version is now in several places!

Makefile
```diff
@@ -20,3 +20,27 @@ test:
 	@echo hello;
 
 .PHONY: clean $(TARGETS)
+
+
+PREFIX = /usr/local
+
+alib = libvmaf.a
```
li-zhi: Can you move the logic here and below into wrapper/Makefile, and only make calls from this top-level Makefile?

wrapper/Makefile
```diff
@@ -31,11 +33,37 @@ OBJS = \
...
+	$(LIBOBJDIR)/pugixml.o \
+	$(LIBOBJDIR)/combo_lib.o \
+	$(LIBOBJDIR)/vmaf_lib.o \
+    $(LIBOBJDIR)/libvmaf.o \
 
 AVX_OBJS := \
 	$(OBJDIR)/convolution_avx.o \
 
-all: vmafossexec
+all: vmafossexec lib
```
li-zhi: You don't have to include lib in all. It can be called from the top-level Makefile by 'make lib'.

ashk43712:
Hi, added the new interface. read_frame is the callback function using which vmaf can pull data.
I have moved your file related stuff to main.cpp which has also the callback read_frame that pulls the data from files. Same callback can be used by an external program to transfer the data to vmaf. So, no duplication of code this time.
Library header is "libvmaf.h" which contains an API compute_vmaf(..) to call vmaf functions.

Makefile
ashk43712: Hi, done.

wrapper/Makefile
ashk43712: done.

wrapper/src/combo.c
```diff
@@ -50,7 +50,7 @@ int compute_psnr(const float *ref, const float *dis, int w, int h, int ref_strid
 int compute_ssim(const number_t *ref, const number_t *cmp, int w, int h, int ref_stride, int cmp_stride, double *score, double *l_score, double *c_score, double *s_score);
 int compute_ms_ssim(const number_t *ref, const number_t *cmp, int w, int h, int ref_stride, int cmp_stride, double *score, double* l_scores, double* c_scores, double* s_scores);
 
-int combo(const char *ref_path, const char *dis_path, int w, int h, const char *fmt,
+int combo(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, double *score, void *user_data), void *user_data, int w, int h, const char *fmt,
```
li-zhi: Thanks for helping refactor this! This is really helpful.

Makefile
```diff
@@ -18,5 +18,13 @@ clean:
...
 	@echo hello;
+     
+install:
+	cd wrapper; $(MAKE) install; cd ..;
+
+uninstall:
+	cd wrapper; $(MAKE) uninstall; cd ..;     
 
 .PHONY: clean $(TARGETS)
+
```
ashk43712: Hi, done.

ashk43712:
Hi, added one more step for the making of the library (libvmaf.a) using the command make lib. This will create libvmaf.a inside wrapper/ which can then be installed using make install.

li-zhi:
@ashk43712 I am wondering what's the current status of including VMAF as a filter in FFmpeg? Is it going to become part of the main branch, and configurable via ./configure?

rbultje:
I still have to merge his patch for that, but yes.

li-zhi:
Thanks for the message! @rbultje

ashk43712:
@li-zhi  Hi, vmaf is now included as a filter in FFmpeg main branch. It can be configured using:
./configure --enable-libvmaf. The filter lies in vf_libvmaf.c inside libavfilter.

rbultje:
And in case you don't want to read the docs, the way to use it is something like:

ffmpeg -i file -i ref -lavfi libvmaf=/path/to/model.pkl -v info -nostats -f null -

Great work Ashish!

ashk43712:
Thanks @rbultje

li-zhi:
Great work! @ashk43712

ashk43712:
Thanks @li-zhi

li-zhi:
@ashk43712 I am curious, is there an easy to create a test case for libvmaf.a? This will be useful whenever we update VMAF.

ashk43712:
@li-zhi I think it can be done. What all kind of test cases we want for libvmaf.a?

li-zhi:
@ashk43712 I added a test case for libvmaf in the Travis CI loop.

ashk43712:
@li-zhi very cool.

PR #94, opened 2017-07-16
Installation and usage instructions for libvmaf

ashk43712:
Hi, I have added installation and usage instruction for libvmaf so that it becomes easier for users to install and use the library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,11 +110,12 @@ API:
 
 ```
 double compute_vmaf(char* fmt, int width, int height, int (*read_frame)(float *ref_data, float *main_data, float *temp_data,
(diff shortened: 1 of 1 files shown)


li-zhi:
@ashk43712 the structure 'data' and callback function 'read_frame' is not as straightforward for a library user to write their own. Should they be part of libvmaf.h?

li-zhi:
I added some instruction to direct library users to wrapper/src/main.cpp. Hopefully this will make the instruction self-contained.

ashk43712:
@li-zhi thanks for updating the instructions.
structure 'data' is something that a user can create depending on the need and set it before calling vmaf library and use it again in the callback.
Since different users may need different data I think it's a good idea to keep it separate from the library.

PR #97, opened 2017-07-19
pkg-config for libvmaf

ashk43712:
Hi, added **pkg-config** for libvmaf to allow external programs to query libvmaf and to check if it's installed.
Also merged **make lib** with **make** so creation of both **libvmaf.a** and **vmafossexec**  takes place on running **make**.
Linked vmafossexec to libvmaf.a , so vmaf also uses the libvmaf library. Updated readme.
I am ready to revert any changes that are not consistent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,9 @@ install:
   - pip install --upgrade scikit-learn
   - pip install --upgrade h5py
 script:
-  - make lib
+  - make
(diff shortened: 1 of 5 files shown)


li-zhi:
great job @ashk43712

ashk43712:
thanks @li-zhi

PR #132, opened 2017-12-21
Unbreak ptools compilation on MinGW

GyanD:
At present, on MinGW, make fails in /ptools with "Makefile:1: *** missing separator.  Stop.". Use of include directive lets 'make' proceed.

### DIFF
--- a/ptools/Makefile
+++ b/ptools/Makefile
@@ -1,1 +1,1 @@
-Makefile.VMAF
+include Makefile.VMAF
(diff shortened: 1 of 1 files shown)


GyanD:
Title should be 'Unbreak ptools makefile on MinGW'

GyanD:
Windows!#?

This Makefile is a symlink to Makefile.VMAF but would get checked out (silently) as a regular file on Windows. This should be deleted and replaced with a Makefile that includes the target Makefile. Changing core.symlinks for local repo is not a robust workaround as that requires Admin rights.

There are other symlinks in the repo but this particular one leads to a truncated build, so I think it should be replaced.

PR #159, opened 2018-05-06
Refactor compute methods of psnr and ansnr features

jeoliva:
A bit of refactor of psnr/ansnr compute methods. Move pixel format checking out of the read frames loop.


li-zhi:
Thanks @jeoliva !!

PR #167, opened 2018-06-06
Fix thread lock situation

jeoliva:
Fix a thread lock situation detected in vmafossexec (at least when running in MacOS). It happens quite infrequently (1 of 30 approx) but it is there and is easy to reproduce for longer files. This PR modifies `combo_threadfunc` function of `combo.c` to remove `pthread_mutex_unlock` operations that were done over an already unlocked mutex. After applying this change I couldn't reproduce the thread lock anymore.

I took advantage of this change to create a macro for printing progress while keeping the possibility of activating/deactivating it through a preprocessor define (`PRINT_PROGRESS`). Just to keep code easy to read.

### DIFF
--- a/wrapper/src/combo.c
+++ b/wrapper/src/combo.c
@@ -33,6 +33,7 @@
 #include "darray.h"
 #include "adm_options.h"
 #include "combo.h"
+#include "debug.h"
 
 #ifdef MULTI_THREADING
 #include "common/blur_array.h"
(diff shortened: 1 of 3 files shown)


jeoliva:
By the way, forgot to mention: Thanks for the multi thread support!. It, together with the new subsample feature, highly reduce the time that takes calculating VMAF which make it much more usable.

wrapper/src/combo.c
```diff
@@ -242,7 +243,6 @@ void* combo_threadfunc(void* vmaf_thread_data)
         {
 #ifdef MULTI_THREADING
             thread_data->stop_threads = 1;
-            pthread_mutex_unlock(&thread_data->mutex_readframe);
```
li-zhi: Thank you for this! I spotted the same issue and is doing the debugging but you beat me! The weird part is that this issue didn't emerge in Ubuntu when the compiler was g++.

li-zhi:
@jeoliva the credit for multi-threading feature goes to @DonTequila

PR #170, opened 2018-06-09
Refactor unittests to reduce redundant code; Add basic E-VMAF utility.

christosbampis:


python/src/vmaf/testutil.py
```diff
@@ -0,0 +1,41 @@
+__copyright__ = "Copyright 2016-2018, Netflix, Inc."
```
li-zhi: Since it is a utility for test only, consider moving it under `python/test`.

python/src/vmaf/core/quality_runner.py
```diff
@@ -420,6 +420,137 @@ def _remove_result(self, asset):
         vmaf_fassembler = self._get_vmaf_feature_assembler_instance(asset)
         vmaf_fassembler.remove_results()
 
+class EnsembleVmafQualityRunner(VmafQualityRunner):
```
li-zhi: We can put this in for now, but may need to revisit later to refactor it to make the structure better.

PR #179, opened 2018-06-22
Rename conf_interval to enable_conf_interval for clarity; change its …

li-zhi:
…type from bool to int for compute_vmaf(); update version to 1.3.8.


li-zhi:
@kparkins could you review the change?

PR #198, opened 2018-07-21
Json and Xml file assembly.

christosbampis:


python/test/result_test.py
```diff
@@ -244,5 +249,115 @@ def test_load_result_with_none(self):
...
+                'model_filepath': VmafConfig.model_path("vmaf_v0.6.1.pkl"),
+            },
+            optional_dict2=None,
+        )
+        self.runner.run()
+
+        results = self.runner.results
+
+        xml_file_list = []
+        dump_dir = os.getcwd() + '/temp_outputs/'
```
li-zhi: Do not do this. Instead, use VmafConfig.workdir_path(...). Follow example at:
https://github.com/Netflix/vmaf/blob/master/python/test/command_line_test.py#L16

Also, for tests, each test must be 'memoryless', that means each test should not interfere another. Therefore, you must clean up your temp files after creating them. This is best done in the setUp() and tearDown() function. Follow the example above.

python/test/result_test.py
```diff
@@ -244,5 +249,115 @@ def test_load_result_with_none(self):
...
+        with open(xml_file, 'wb') as myfile:
+            myfile.write(results[1].to_xml())
+        xml_file_list.append(xml_file)
+
+        import xml.etree.ElementTree as ET
+        tree_1 = ET.ElementTree(ET.fromstring(results[0].to_xml()))
+        tree_2 = ET.ElementTree(ET.fromstring(XML_Assembler([xml_file_list[0]]).assemble()))
+
+        tree_3 = ET.ElementTree(ET.fromstring(XML_Assembler(xml_file_list).assemble()))
+        tree_4 = ET.ElementTree(ET.fromstring(XML_Assembler([xml_file_list[1], xml_file_list[0]]).assemble()))
```
li-zhi: This is to test if the order of the files in the list changes, the result doesn't change?

Once assembled, will you re-serialize the frame numbers? You should. And make sure to add a test to test it.

python/src/vmaf/tools/ResultAssembler.py
```diff
@@ -0,0 +1,231 @@
...
+
+        return assembly_output
+
+    def create_aggregation_dict(self):
+
+        # TODO fix: only does for aggregation
+        aggregate_dict = OrderedDict()
+        valid_score_keys = sorted([key for key in self.scores_dict if key.endswith("score")])
+        for key in valid_score_keys:
+            aggregate_dict[key] = np.mean(self.scores_dict[key])
```
li-zhi: Here you hard-coded the aggregate method to 'mean'. You shouldn't. See how Result.to_xml() does it at:
https://github.com/Netflix/vmaf/blob/master/python/src/vmaf/core/result.py#L221

python/src/vmaf/tools/ResultAssembler.py
```diff
@@ -0,0 +1,231 @@
...
+__copyright__ = "Copyright 2016-2018, Netflix, Inc."
+__license__ = "Apache, Version 2.0"
+
+
+def print_usage():
+    print("usage: python " + os.path.basename(sys.argv[0]) + " input_file output_file\n")
+    print("input_file contains a list of files for assembly (can be xml or json)")
+    print("output_file contains the assembled file")
+
+class FileAssembler():
```
li-zhi: If these classes go to tools.py, it means they much be very general. However, this is not the case. For example, these classes rely on underlying structure of 'scores', which is specifically related to the Result class. Logically, the assemble functions you are implementing is associated with Result, so they should be moved under Result as 'staticmethod'.

This is a solution, but it may not be the most elegant one. I am thinking what would be the best way to do this. Result already has 'to_json' and 'to_xml' methods, which is to convert a Result object to a file. Similarly, you can create 'from_json' and 'from_xml', which can convert a file to a Result object. Then, to assemble two files, you simply need to write a method to assemble two Result objects into one new Result object. After assembly, you can call 'to_json' or 'to_xml' to convert the new object to a file. This seems to be the most elegant way, since it avoids issues like you have to implement twice the logic associated with calculating the 'aggregate'.

PR #202, opened 2018-07-25
Add desired_yuv_type attribute in asset.

christosbampis:


python/src/vmaf/core/asset.py
```diff
@@ -570,6 +574,23 @@ def dis_yuv_type(self):
...
 
+    @property
+    def desired_yuv_type(self):
+        """
+        for notyuv assets, we want to allow the decoded yuv format to be set by the user
+        this is highly relevant to image decoding, where we would like to select yuv444p
+        this property tries to read desired_yuv_type from asset_dict, if it is there it is set
+        else it default to default_yuv_type
+        """
+        allowed_yuv_types = [y for y in Asset.SUPPORTED_YUV_TYPES if y != 'notyuv']
```
li-zhi: A better way is to convert to a set and exclude notyuv. See https://stackoverflow.com/questions/3462143/get-difference-between-two-lists

python/src/vmaf/core/asset.py
```diff
@@ -570,6 +574,23 @@ def dis_yuv_type(self):
...
+        if 'desired_yuv_type' in self.asset_dict:
+            if self.asset_dict['desired_yuv_type'] in allowed_yuv_types:
```
li-zhi: I would prefer:
```
desired_yuv_type = self.asset_dict['desired_yuv_type']
assert desired_yuv_type in allowed_yuv_types, "Desired YUV format {} is not valid, pick: {}".format(desired_yuv_type, str(allowed_yuv_types))
return desired_yuv_type
```
"one logic a line" is preferred than "nested logic".

python/src/vmaf/core/asset.py
christosbampis: +1 for workfile_yuv_type, I like it a lot!

not sure about supported_yuv_types: there is already .SUPPORTED_YUV_TYPES, but we want to exclude notyuv from it

christosbampis:
fixed all of those!

PR #210, opened 2018-08-15
Add frame differencing FE for VIF.

christosbampis:


feature/src/vif.c
```diff
@@ -394,3 +394,159 @@ int vif(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, i
...
+        }
+        if (ret == 2)
+        {
+            break;
+        }
+
+        // ===============================================================
+        // offset pixel by OPT_RANGE_PIXEL_OFFSET
+        // ===============================================================
+        offset_image(ref_buf, OPT_RANGE_PIXEL_OFFSET, w, h, stride);
```
li-zhi: Not sure if this is the right place to apply this offset. In original VIF, ADM, the offset is -128 is applied before the values are passed into compute_vif or compute_adm to ensure they are 0-mean. Here if you do it before the differencing, it has no effect.

I am also wondering: the output values of the apply_frame_differencing will have range twice of the input values. What would be the consequences when passing the output values to VIF?

feature/src/vif.c
```diff
@@ -394,3 +394,159 @@ int vif(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, i
...
+        if (frm_idx == 0)
+        // pay attention for first frame, frame difference for first frame is not well-defined when subtracting the current frame from the previous one
+        // why not define the frame-difference in the opposite direction? Since quality is mostly affected by recency, it is better to get an
+        // unreliable T-VIF SCORE for earlier frames. Therefore we set the scores to some default value (e.g. 0).
+		{
+		    score = 0.0;
+		    score_num = 0.0;
+		    score_den = 0.0 + 1e-5;
+		    for(int scale=0; scale<4; scale++){
+		    	scores[2*scale] = 0.0;
```
li-zhi: Always add space around *, =, etc.

feature/vmaf.dSYM/Contents/Info.plist
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
li-zhi: What are these?

christosbampis: this is something that slipped through, some random files that sometimes get generated, will remove

feature/src/vif.c
christosbampis: indeed I try to follow this; however, some pieces of the existing code did not follow it

feature/src/vif.c
christosbampis: good question! I think that it might be still ok, since the difference of two zero mean is still zero mean, right?
about the second question, let me think this tomorrow, if it is a problem

PR #225, opened 2018-09-12
Fix implicit declarations of functions

snghojeong:
I'm making VMAF run in iPhone & Android using LLVM.
First of all, I fix implicit declarations of functions to follow C99.


Jinyu-Qin:
How do you compile on Android?

snghojeong:
I modified build script to build VMAF for iOS and Android.

PR #245, opened 2018-10-27
Create perf metric distributions by bootstrapping test indices.

christosbampis:


python/test/routine_test.py
```diff
@@ -226,6 +226,17 @@ def test_test_on_dataset(self):
         self.assertAlmostEqual(test_assets[2].groundtruth, 100, places=4)
         self.assertAlmostEqual(test_assets[3].groundtruth, 80, places=4)
 
+    def test_test_on_dataset_bootstrap_perf_metric(self):
```
li-zhi: Again, since most of the change is in train_test_model, can you put some tests in train_test_model_test?

christosbampis: sure, how about the new test: test_train_bootstrap_perf_metric_libsvmnusvr
updated the commit accordingly (some minor fixes as well e.g. to allow for specifying the number of bootstraps for determining the performance metric list as a keyword arg)

PR #250, opened 2018-11-03
Extract number of bootstrapped models for plotting.

christosbampis:


python/src/vmaf/core/quality_runner.py
```diff
@@ -970,6 +972,7 @@ def predict_with_model(cls, model, xs, **kwargs):
                 'ys_pred_stddev': ys_pred_stddev,
                 'ys_pred_ci95_low': ys_pred_ci95_low,
                 'ys_pred_ci95_high': ys_pred_ci95_high,
+                'num_type_bootstrap_models': num_type_bootstrap_models,
```
li-zhi: Do you really need to pass this out? Can't this be figured out by the length of the list of 'all_models'?

python/src/vmaf/core/train_test_model.py
```diff
@@ -1108,12 +1108,16 @@ def _get_num_models_from_param_dict(cls, param_dict):
             'num_models'] if 'num_models' in param_dict else cls.DEFAULT_NUM_MODELS
         return num_models
 
+    def get_num_type_bootstrap_models(self):
```
li-zhi: Don't think it is a good idea to pass out a string.

christosbampis: Addressed this in the updated PR, we just count the number of models from the size of prediction_all_models.

python/src/vmaf/core/quality_runner.py
christosbampis: Yes, new PR does that.

PR #273, opened 2018-12-10
Output per bootstrap model scores in vmafossexec.

christosbampis:
Changes to the C++ code so that the output of vmafossexec in xml and json format reports the per bootstrap model scores.

### DIFF
--- a/python/src/vmaf/core/quality_runner.py
+++ b/python/src/vmaf/core/quality_runner.py
@@ -701,7 +701,20 @@ class VmafossExecQualityRunner(QualityRunner):
         tree = ElementTree.parse(log_file_path)
         root = tree.getroot()
         scores = []
+
+        num_bootstrap_models = np.int(root.findall('params')[0].attrib['num_bootstrap_models'])
+        bootstrap_model_prefix = root.findall('params')[0].attrib['bootstrap_model_prefix']
+
+        bootstrap_model_list = []
+        for bootstrap_ind in range(num_bootstrap_models):
+            # NOTE: bootstrap model indices start from 1
+            bootstrap_model_list.append('{}{:04d}'.format(bootstrap_model_prefix, bootstrap_ind + 1))
+
+        # augment the feature set with any bootstrap models
+        self.FEATURES += bootstrap_model_list
+
         feature_scores = [[] for _ in self.FEATURES]
+
         for frame in root.findall('frames/frame'):
             scores.append(float(frame.attrib['vmaf']))
             for i_feature, feature in enumerate(self.FEATURES):
--- a/wrapper/src/vmaf.cpp
+++ b/wrapper/src/vmaf.cpp
@@ -279,17 +279,17 @@ void LibsvmNusvrTrainTestModel::populate_and_normalize_nodes_at_frm(size_t i_frm
     }
 }
 
-std::map<VmafPredictionReturnType, double> LibsvmNusvrTrainTestModel::predict(svm_node* nodes) {
+VmafPredictionStruct LibsvmNusvrTrainTestModel::predict(svm_node* nodes) {
 
     double prediction = svm_predict(svm_model_ptr.get(), nodes);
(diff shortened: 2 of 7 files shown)


wrapper/src/vmaf.cpp
```diff
@@ -881,22 +892,41 @@ void BootstrapVmafQualityRunner::_postproc_transform_clip(
...
+        bootstrap_ind_str = bootstrap_ind_strstream.str();
+
+        result.set_scores("vmaf_bootstrap_model_" + bootstrap_ind_str, perModelScore);
+        perModelScore = perModelScore.clear();
+        bootstrap_ind_strstream.str("");
+    }
+
 }
 
 static const char VMAFOSS_DOC_VERSION[] = "1.3.7";
```
li-zhi: Need update the doc version

wrapper/src/vmaf.h
```diff
@@ -63,6 +63,10 @@ class StatVector
 public:
     StatVector() {}
     StatVector(std::vector<double> l): l(l) {}
+    StatVector clear()
```
li-zhi: What does this mean?

christosbampis: We define a public method to the StatVector class, which allows us to erase all the contents of a created StatVector object. This is used when we re-use the StatVector perModelScore, which fills up with predictions from each bootstrap model and then is re-used for the next model.

wrapper/src/vmaf.cpp
christosbampis: done!

python/src/vmaf/core/quality_runner.py
```diff
@@ -701,8 +701,21 @@ def _get_quality_scores(self, asset):
...
+            # NOTE: bootstrap model indices start from 1
+            bootstrap_model_list.append('{}{:04d}'.format(bootstrap_model_prefix, bootstrap_ind + 1))
+
+        # augment the feature set with any bootstrap models
+        self.FEATURES += bootstrap_model_list
+
         feature_scores = [[] for _ in self.FEATURES]
-        for frame in root.findall('frames/frame'):
+
+        for frame_ind, frame in enumerate(root.findall('frames/frame')):
```
li-zhi: why need frame_ind?

wrapper/src/vmaf.cpp
```diff
@@ -823,83 +827,117 @@ std::unique_ptr<LibsvmNusvrTrainTestModel> BootstrapVmafQualityRunner::_load_mod
...
+        }
+        sprintf(char_buffer, "%04d", j + 1);
+        result.set_scores("vmaf_" + std::string(char_buffer), perModelScore);
+        perModelScore.clear();
+    }
+
 }
 
-static const char VMAFOSS_DOC_VERSION[] = "1.3.7";
+static const char VMAFOSS_DOC_VERSION[] = "1.3.10";
```
li-zhi: Actually it should be 1.3.11 now. Could you also update three things: 1) VERSION, 2) CHANGELOG.md, 3) wrapper/libvmaf.pc?

wrapper/src/vmaf.cpp
```diff
@@ -823,83 +827,117 @@ std::unique_ptr<LibsvmNusvrTrainTestModel> BootstrapVmafQualityRunner::_load_mod
...
     _clip_value(model, baggingScore);
     _clip_value(model, ci95LowScore);
     _clip_value(model, ci95HighScore);
     _clip_value(model, scorePlusDelta);
     _clip_value(model, scoreMinusDelta);
+
+    // clip bootstrap model scores
+    int num_models = predictionStruct.vmafMultiModelPrediction.size();
+
+    for (int model_i; model_i < num_models; model_i++)
```
HolyWu: Shouldn't `model_i` be initialized with value 0?

christosbampis: good catch, thanks! I will be updating this together with some other fixes

PR #289, opened 2019-01-05
Optimization of ADM, VIF and buffer management modules (#288)

li-zhi:


feature/src/adm.c
```diff
@@ -48,6 +53,228 @@ static char *init_dwt_band(adm_dwt_band_t *band, char *data_top, size_t buf_sz_o
     return data_top;
 }
 
+#if ADM_OPT_ENABLE
+static char *init_dwt_band_hvd(adm_dwt_band_t *band, char *data_top, size_t buf_sz_one)
```
kylophone: type of `data_top` should be a `float*`. Every call of this function, the data starts as a `float*` get casted to a a `char*` and in the function body gets casted back to a `float*`

feature/src/adm.c
```diff
@@ -48,6 +53,228 @@ static char *init_dwt_band(adm_dwt_band_t *band, char *data_top, size_t buf_sz_o
...
+
+	int scale;
+	int ret = 1;
+	
+	// Code optimized to save on multiple buffer copies 
+	// hence the reduction in the number of buffers required from 35 to 17 
+#define NUM_BUFS_ADM 17 
+	if (SIZE_MAX / buf_sz_one < NUM_BUFS_ADM)
+	{
+		printf("error: SIZE_MAX / buf_sz_one < NUM_BUFS_ADM, buf_sz_one = %zu.\n", buf_sz_one);
```
kylophone: `fprintf(stderr, ...)` ?

kylophone: This goes for all `printf()'`s. We may want to use this library with piped video data in the future, and we don't log messages mixed in.

feature/src/adm.c
```diff
@@ -48,6 +53,228 @@ static char *init_dwt_band(adm_dwt_band_t *band, char *data_top, size_t buf_sz_o
...
+	}
+
+	if (!(data_buf = aligned_malloc(buf_sz_one * NUM_BUFS_ADM, MAX_ALIGN)))
+	{
+		printf("error: aligned_malloc failed for data_buf.\n");
+		fflush(stdout);
+		goto fail;
+	}
+
+	data_top = (char *)data_buf;
```
kylophone: Don't cast this to a `char*`

PR #295, opened 2019-01-23
Print out individual vmaf bootstrap scores in text file from vmafosse…

christosbampis:
…xec.

### DIFF
--- a/wrapper/src/vmaf.cpp
+++ b/wrapper/src/vmaf.cpp
@@ -900,8 +900,14 @@ void BootstrapVmafQualityRunner::_postproc_transform_clip(
     scoreStdDev *= slope;
 }
 
-static const char BOOTSTRAP_MODEL_NAME_TRAILING_ZEROS[] = "%04d";
-static const int BOOTSTRAP_MODEL_NAME_BUF_SIZE = 100;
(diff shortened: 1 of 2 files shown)


wrapper/src/vmaf.cpp
```diff
@@ -922,13 +925,13 @@ void BootstrapVmafQualityRunner::_set_prediction_result(
     size_t num_models = predictionStructs.at(0).vmafMultiModelPrediction.size();
     std::vector<double> perModelScore;
     // character array to put the name of the vmaf bootstrap model, e.g. vmaf_0001 is the first one
-    char char_buffer[50];
+    char char_buffer[BOOTSTRAP_MODEL_NAME_BUF_SIZE];
```
li-zhi: Suggestion: use c++ style std::string instead of c style char. For example:
https://github.com/Netflix/vmaf/blob/master/wrapper/src/vmaf.cpp#L147

christosbampis: just updated the PR with a commit, what do you think?

li-zhi: Looks good. Go ahead with the merge!

christosbampis: done, thanks!

PR #298, opened 2019-01-26
Fix incorrect VMAF score on VS2015 and refactor MSVC project files

HolyWu:
1. Fix incorrect VMAF score on VS2015 due to the instruction set of convolution_avx.c is not set to /arch:AVX.
2. Refactor MSVC project files. Remove redundant settings and cleanup.


li-zhi:
@HolyWu I'll wait to merge this until we find the root cause of the discrepancy. We are looking into why libvmaf and vmafossexec are generating different results.

PR #325, opened 2019-04-15
Avoid dis-filtering ref when not needed (!ANSNR_OPT_NORMALIZE, which is

legrosbuffle:
the default).

We measured that we're spending most of our time doing the 2D convolution.
This makes compute_ansnr() ~1.8x faster.

In addition, reuse the filtd temporary buffer. This makes the function
use 2/3 of the original memory.


li-zhi:
@legrosbuffle Thanks for the contribution. Note that ANSNR is one of the elementary metrics, but is not used in the latest VMAF models (e.g. 0.6.1). It was used in earlier versions.

PR #326, opened 2019-04-17
A few compilation fixes (warnings and errors).

legrosbuffle:


li-zhi:
Sorry it took so long. Completely slipped under my radar.

PR #332, opened 2019-06-03
Support for python3

zsimic:
Adapted code for python3, exercise tests in both python2 and python3


python/src/vmaf/__init__.py
```diff
@@ -1,13 +1,44 @@
 import os
+import subprocess
+
+try:
+    from matplotlib import pyplot as plt
```
zsimic: Importing `matplotlib.pyplot` only once, to avoid having to repeat this same pattern in multiple places in the code.

`matplot` has its quirks... they're better dealt with once like this

python/src/vmaf/__init__.py
```diff
@@ -1,13 +1,44 @@
...
 
 # Path to folder containing this file
 VMAF_LIB_FOLDER = os.path.dirname(os.path.abspath(__file__))
 
 # Assuming vmaf source checkout, path to top checked out folder
 VMAF_PROJECT = os.path.abspath(os.path.join(VMAF_LIB_FOLDER, '../../..',))
 
+def to_list(value):
+    """
+    TODO python3: map() and filter() return a generator in python3, however vmaf assumes list
```
zsimic: I'm not sure why `map()` was used so often, perhaps to just keep code more compact.
There is a drastic change in py2/py3 for `map()` (as well as stuff like `dict.keys()` etc)

python3 uses generators in many places now (like `map()`, `range()`, `dict.keys()` etc)

python/src/vmaf/__init__.py
```diff
@@ -1,13 +1,44 @@
...
+
+    Returns:
+        (list): `value` converted to a list when applicable
+    """
+    if value is None or isinstance(value, list):
+        return value
+    return list(value)
+
+def run_process(cmd, **kwargs):
```
zsimic: `run_process` moved here to avoid import loops (made evident by stuff like `pytest --doctest-modules`).

python/src/vmaf/config.py
```diff
@@ -20,7 +20,7 @@ def _path_from_external(name):
         :return str: Configured path, if any
         """
         try:
-            import externals
+            from . import externals
```
zsimic: Forget about stuff like this -> with python3 all imports must be absolute, and one has to explicitly use `.` to do a relative import.

python/src/vmaf/core/executor.py
```diff
@@ -377,7 +378,7 @@ def _post_process_result(cls, result):
     def _get_log_file_path(self, asset):
         return "{workdir}/{executor_id}_{str}".format(
             workdir=asset.workdir, executor_id=self.executor_id,
-            str=hashlib.sha1(str(asset)).hexdigest())
+            str=hashlib.sha1(str(asset).encode("utf-8")).hexdigest())
```
zsimic: `hashlib` requires a `bytes` object in python3
In python2, there was an amalgamation/sometimes-confusion of str/unicode/bytes.

python/src/vmaf/core/feature_assembler.py
```diff
@@ -87,18 +88,18 @@ def run(self):
...
             assert fextractor_type in self.type2results_dict
             for atom_feature in self._get_atom_features(fextractor_type):
                 scores_key = self._get_scores_key(fextractor_type, atom_feature)
                 for result_index, result in enumerate(self.type2results_dict[fextractor_type]):
                     result_dicts[result_index][scores_key] = result[scores_key]
 
-        self.results = map(
-            lambda (asset, result_dict): BasicResult(asset, result_dict),
+        self.results = to_list(map(
+            lambda tasset: BasicResult(tasset[0], tasset[1]),
```
zsimic: `lambda (a, b): ...` is a syntax error in python3.
See for example http://cheng.logdown.com/posts/2016/11/30/python3-lambda-parameter

python/src/vmaf/core/nn_train_test_model.py
```diff
@@ -5,7 +5,12 @@
 import numpy as np
 import scipy.stats
 from sklearn.metrics import f1_score
-import tensorflow as tf
+
+try:
+    import tensorflow as tf
```
zsimic: `tensorflow` is not declared as a requirement by vmaf, so this is either experimental code, or never used in tests
(however fails to even `import`, when doctests tries to discover stuff, it fails to even look at this file without this `try:` protection)

python/src/vmaf/core/noref_feature_extractor.py
```diff
@@ -384,8 +383,8 @@ def mscn_extract_niqe(cls, img, patch_size, mode):
 
         for j in range(0, h-block_h+1, shift_h):
             for i in range(0, w-block_w+1, shift_w):
-                m_patch1 = m_image1[j  :(j+block_h),   i  :(i+block_w)  ]
-                m_patch2 = m_image2[j/2:(j+block_h)/2, i/2:(i+block_w)/2]
+                m_patch1 = m_image1[j:j + block_h, i:i + block_w]
+                m_patch2 = m_image2[j // 2:(j + block_h) // 2, i // 2:(i + block_w) // 2]
```
zsimic: `/` yields floats in python3
One should use explicitly integer division `//` where an integer is desired (works in both py2/py3)

python/src/vmaf/tools/reader.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, filepath, width, height, yuv_type):
 
         self._asserts()
 
+        # TODO python3: this doesn't work well with python3, need to refactor this class
         self.file = open(self.filepath, 'rb')
```
zsimic: This wasn't great in python2 either, and becomes quite hard to follow with the changes around how files are handled in python3

If you're familiar with `YuvReader`, please refactor this to not use things like this

python/src/vmaf/tools/reader.py
```diff
@@ -56,10 +57,8 @@ def __iter__(self):
         return self
 
     def next(self):
-        try:
-            return self.next_y_u_v()
-        except EOFError:
-            raise StopIteration
+        """next() is for python2 only, in python3 all you need to define is __next__(self)"""
+        return self.__next__()
```
zsimic: See http://python-future.org/compatible_idioms.html#custom-iterators

python/test/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+def pytest_collection_modifyitems(items):
+    """
+    Dynamically mark tests based on their file name:
+    - *_test.py: main test (always exercise)
+    - *_extratest.py: exercised only when testing with ffmpeg
+    - *_libtest.py: exercised only to test testlib
+    """
+    for item in items:
+        if "_extratest" in item.nodeid:
+            item.add_marker(pytest.mark.extra)
```
zsimic: This here applies to `pytest` only, it simply systematically marks all tests based on their filename

One can still use `unittest`, this will have no effect on that

python/test/feature_assembler_test.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
```
zsimic: This statement is not required anymore (used to be with py2.6 and early 2.7s)
I added it due to `unittest` failing to see `./test/` as a module, but it didn't help...

Doesn't hurt, but modern 2.7 versions don't usually even need this...

python/test/lib/libvmaf_libtest.py
```diff
@@ -8,13 +8,15 @@
 __copyright__ = "Copyright 2016-2018, Netflix, Inc."
 __license__ = "Apache, Version 2.0"
 
-class TestLibRunner(VmafossExecQualityRunner):
+
+class LibRunner(VmafossExecQualityRunner):
```
zsimic: Renamed to `LibRunner` because otherwise `TestLibRunner` issues a warning that this is not a valid test case (which it isn't supposed to be... it's just that class prefix `Test` seems to imply that it is by convention...)

python/test/reader_test.py
```diff
@@ -28,7 +28,7 @@ def test_with(self):
             height=324,
             yuv_type='yuv420p'
         ) as yuv_reader:
-            self.assertEquals(yuv_reader.file.__class__, file)
+            assert hasattr(yuv_reader.file, "read")
```
zsimic: There is no more type `file` in python3

python/test/result_test.py
```diff
@@ -113,7 +113,7 @@ def test_to_score_str(self):
         self.result.set_score_aggregate_method(ListStats.total_variation)
         self.assertAlmostEquals(self.result.get_result('VMAF_legacy_score'), 6.5901873052628375, places=4)
         self.result.set_score_aggregate_method(partial(ListStats.moving_average, n=2))
-        self.assertItemsEqual(self.result.get_result('VMAF_legacy_score'),
+        self.assertEqual(list(self.result.get_result('VMAF_legacy_score')),
```
zsimic: A few of those `assertSomething` methods were removed in python3's unittest...

The consensus is roughly that there are too many of them, they're a lot java-like, and don't really provide much value...

With a modern test runner (like `pytest`), you don't need this at all.
A simple `assert a == b` will usually be sufficient (you will get a diff output of the difference too, no need for elaborate functions)

python/tox.ini
```diff
@@ -0,0 +1,56 @@
...
+
+[testenv]
+passenv = APPVEYOR* CI SSL_CERT_FILE TEST_MARKER TRAVIS*
+setenv = COVERAGE_FILE={toxworkdir}/.coverage.{envname}
+usedevelop = True
+deps = -rrequirements.txt
+       -rtest/requirements.txt
+# TODO: way too many warnings in tests, remove `-p no:warnings` and get rid of them all
+commands = pytest {posargs:-vv -p no:warnings -m {env:TEST_MARKER:main} --doctest-modules --cov-report term-missing --cov=src/}
+#          python -m unittest discover -s test/ -p '*_test.py'
```
zsimic: One can simply comment the `pytest` line here, and uncomment the `unittest` one to switch to unittest to exercise the tests

python/tox.ini
```diff
@@ -0,0 +1,56 @@
...
+
+
+# Configuration
+[pytest]
+cache_dir = .tox/.cache
+python_files = *test.py
+markers =
+    main: Main tests (exludes 'extra' and 'lib')
+    extra: Exercise test/extra/ (requires ffmpeg installed)
+    lib: Exercise test/lib/ (requires testlib installed).
```
zsimic: These just define all the markers used, `pytest --markers` would show those 3 (with their description), it's like a relatively common practice to declare them like this.

pytest looks at the following file for configuration (in that order): `pytest.ini`, `tox.ini`, `setup.cfg`

python/src/vmaf/tools/reader.py
li-zhi: I'll try to refactor YuvReader. Which tests does it break in python3?

zsimic: `test/extra/noref_feature_extractor_extratest.py`, they're marked with:
```
@unittest.skipIf(sys.version_info > (3,), reason="TODO python3: unclear how YuvReader is supposed to work")
```

python/README.rst
```diff
@@ -0,0 +1,15 @@
...
+
+.. image:: https://travis-ci.org/Netflix/vmaf.svg?branch=master
+    :target: https://travis-ci.org/Netflix/vmaf
+    :alt: Travis Build Status
+
+.. image:: https://ci.appveyor.com/api/projects/status/68i57b8ssasttngg?svg=true
+    :target: https://ci.appveyor.com/project/li-zhi/vmaf
+    :alt: AppVeyor Build Status
+
+VMAF is a perceptual video quality assessment algorithm developed by Netflix.
```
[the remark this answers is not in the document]
li-zhi: Sounds good.

python/script/run_toddnoiseclassifier.py
```diff
@@ -13,66 +13,71 @@
...
+    DisYUVRawVideoExtractor.close_h5py_file(raw_video_h5py_file)
+    ToddNoiseClassifierTrainTestModel.close_h5py_file(patch_h5py_file)
+    os.remove(raw_video_h5py_filepath)
+    os.remove(patch_h5py_filepath)
+
+    print('Done.')
+
+
+if __name__ == "__main__":
+    main()
```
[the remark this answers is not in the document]
li-zhi: Thanks. Good catch.

Curious: how did you find this file? Do you have some analysis tool or you just did spot check?

python/src/vmaf/__init__.py
li-zhi: Personally i prefer map to list comprehension, since it can be more easily converted into parallel map. What's your take on it?

python/script/run_toddnoiseclassifier.py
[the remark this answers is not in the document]
zsimic: `pytest --doctest-modules` found it actually -> it was hitting this because it tries to simply `import` the module (in order to scour it for doctests), and this one was trying to run immediately at import time, and failing.

python/src/vmaf/__init__.py
zsimic: Generators can be cheaper to run basically.
What python2 was doing before is the equivalent of this, ie `map()` in python2 is the same as `to_list(map())` here in this case.

Since you're accessing the outcome of `map()` by array-like index, then I guess there is no other way than turning it into something that can be accessed in such a way.

Perf will depend on the size. If `map()` deals with a million items, obviously then expanding that into a list will be costly (must expand full million items...).

With python3, if `map()` yields a million items, you can process them without ever having the whole million in RAM basically. So depends on usage I think. We can't change the way py3 works though, so must deal with the generator :)

zsimic: Google around a bit, and this may be a thing that could be done. Consider this:
```
>>> data = [1,2,3,None]
>>> a = np.array(data)
>>> a
array([1, 2, 3, None], dtype=object)
>>> a[a != np.array(None)]
array([1, 2, 3], dtype=object)
```

Now if we look at `stats.py` for example this function:
```
    @staticmethod
    def nonemean(my_list):
        return np.mean(to_list(filter(lambda x: x is not None, my_list)))
```

Looks like that we could replace this using `np` only like this:

```
    @staticmethod
    def nonemean(my_list):
        a = np.array(my_list)
        return np.mean(a[a != np.array(None)])
```

I'm not super familiar with numpy, and it does look like it hates `None` (whereas pandas seems fine with `None`...).

I found the trick here: https://stackoverflow.com/questions/25254929/efficient-way-of-removing-nones-from-numpy-array

Not sure which is better performance-wise: the np `a[a != ..]`, or the `filter()` approach... one would have to benchmark this to find out.

PR #334, opened 2019-06-04
errors: update name of file for failed dis_path fopen

zsiec:
I got stuck for a second because of this incorrect label in the error message. Hopefully this will help others!


li-zhi:
good catch! Thanks.

PR #337, opened 2019-06-24
Print progress in vmafossexec

### LINKED ISSUE #305: vmafossexec no progress reported
Hello, 
As 'vmafossexec' does its processing, it does not report how far along the process is. It would be nice, if possible, to include the indicator of progress. 

Thanks,
Max

HolyWu:
Closes #305.

### DIFF
--- a/feature/src/common/frame.h
+++ b/feature/src/common/frame.h
@@ -27,6 +27,7 @@ struct data
     size_t offset;
     FILE *ref_rfile;
     FILE *dis_rfile;
+    int num_frames;
 };
 
 struct noref_data
--- a/wrapper/src/main.cpp
+++ b/wrapper/src/main.cpp
@@ -118,6 +118,8 @@ int run_wrapper(char *fmt, int width, int height, char *ref_path, char *dis_path
     s->height = height;
     s->ref_rfile = NULL;
     s->dis_rfile = NULL;
+    long file_size = 0;
+    int frame_size = 0;
 
     if (!strcmp(fmt, "yuv420p") || !strcmp(fmt, "yuv420p10le"))
     {
@@ -157,6 +159,37 @@ int run_wrapper(char *fmt, int width, int height, char *ref_path, char *dis_path
         fprintf(stderr, "fopen dis_path %s failed.\n", dis_path);
         ret = 1;
         goto fail_or_end;
+    }
+
+    if (strcmp(ref_path, "-"))
+    {
+        if (fseek(s->ref_rfile, 0, SEEK_END))
+        {
+            fprintf(stderr, "fseek failed.\n");
+            ret = 1;
+            goto fail_or_end;
+        }
+
+        if ((file_size = ftell(s->ref_rfile)) == -1)
+        {
+            fprintf(stderr, "ftell failed.\n");
+            ret = 1;
+            goto fail_or_end;
+        }
+
+        rewind(s->ref_rfile);
+
+        frame_size = width * height + s->offset;
+        if (!strcmp(fmt, "yuv420p10le") || !strcmp(fmt, "yuv422p10le") || !strcmp(fmt, "yuv444p10le"))
+        {
+            frame_size *= 2;
+        }
+
+        s->num_frames = file_size / frame_size;
+    }
+    else
+    {
+        s->num_frames = -1;
     }
 
     /* Run VMAF */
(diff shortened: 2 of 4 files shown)


HolyWu:
No idea why the Travis CI build succeeded in Python 3.6 but failed in Python 2.7. I'm not familiar with the test scripts so I can't figure it out.

slhck:
The tests have been updated recently with Python 3 compatibility, but they passed in `master` for 2.7 and 3.x: https://travis-ci.org/Netflix/vmaf/builds/544480780

Could it be that the output changing is causing the test to fail? (Due to `printf` statements…)

Either way here are my two cents:

- It'd be better to have make this feature optional and accessible behind a command line flag, as you don't want progress to be printed.
- Don't print newlines with `\r`, but use the respective system-specific newline delimiters.
- Consider printing to STDERR instead, if you don't want to mix actual parsable output with human-readable one. See other `fprintf` calls in the code.

HolyWu:
OK. Added a command line option `--progress` to make it optional and also passed the test. The use of `\r` is intentional. Its purpose is not to produce a new line, but to keep the progress indicator appearing on the same line. We don't want to flood the console with a lot of lines.

slhck:
Sorry, I didn't notice the exact purpose. Good that the tests pass now!

feature/src/common/frame.h
```diff
@@ -27,6 +27,8 @@ struct data
     size_t offset;
     FILE *ref_rfile;
     FILE *dis_rfile;
+    int print_progress;
```
li-zhi: Not sure if it is best to put `print_progress` in `data`. It doesn't go well with the rest of the fields.

Why do we need to add it as optional in the first place? What was the reason it failed py2.7?

slhck: Yes, that was what caused the failures.

IMO it does not make a lot of sense for a command-line program to automatically print progress, as it gets “muddled” with other, potentially machine-parsable output (unless progress is sent to STDERR and real info is sent to STDOUT). But this is a preference that some tools follow, some don't.

HolyWu: @li-zhi OK. I removed the `--progress` option again. The progress was printed to stdout instead of stderr before, and that seemed to cripple py2.7's parsing code. The tests passed after sending the progress to stderr according to slhck's suggestion.

wrapper/src/main.cpp
```diff
@@ -159,6 +161,30 @@ int run_wrapper(char *fmt, int width, int height, char *ref_path, char *dis_path
...
+
+    rewind(s->ref_rfile);
+
+    frame_size = width * height + s->offset;
+    if (!strcmp(fmt, "yuv420p10le") || !strcmp(fmt, "yuv422p10le") || !strcmp(fmt, "yuv444p10le"))
+    {
+        frame_size *= 2;
+    }
+
+    s->num_frames = file_size / frame_size;
```
li-zhi: I see the way you determine the num_frames is to have a first pass of the file size, and then use the file size divided by the frame size. This works when the input is YUV files, but probably won't work if it operates in the stream mode (for example, having libvmaf as a filter for FFmpeg), where the input is fed from an instantaneously decoded stream  of compressed video. 

My recommendation is not to to print out the num_frames as the denominator.

feature/src/common/frame.h
```diff
@@ -27,6 +27,7 @@ struct data
     size_t offset;
     FILE *ref_rfile;
     FILE *dis_rfile;
+    int num_frames;
```
li-zhi: We probably don't want this field either. See my comment below.

wrapper/src/main.cpp
HolyWu: From what I see, the `run_wrapper` function in `wrapper/src/main.cpp ` is exclusive invoked by vmafossexec. FFmpeg invokes `compute_vmaf` in `wrapper/src/libvmaf.cpp`, which doesn't go through this `run_wrapper` function. Nevertheless, I could skip the file size determination and set `num_frames` to -1 when "-" (pipe) is used as input to indicate that the total frames is unknown. Is that acceptable to you?

li-zhi: Got it. That works. Thanks!

HolyWu:
Rebased. Please check.

li-zhi:
Thanks!

PR #344, opened 2019-07-19
Fix ocsynchronizer.h to work with MinGW

Zeranoe:
MinGW does not declare clock_gettime(), so use GetSystemTimeAsFileTime() for both VS and MinGW.


kylophone:
Closing this for now. Added a few labels in case we need to reference or re-open this issue in the future.

PR #356, opened 2019-08-27
Libvmaf changes and multiple model support.

christosbampis:
This PR has a number of important changes:

1. Libvmaf interface change
2. Support for multiple prediction models, each additional model has the flexibility to use different enable_transform, disable_clip and ci than other models
3. Decouple feature extraction and model prediction
4. As part of the API change, we should also change read_frame and extend it to read all three color channels. To do so, we introduce a VmafPicture struct as a helper.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## (9/9/2019) [1.4.0]
+
+**New features:**
+- Support for multiple models.
+- Libvmaf interface changes.
+
 ## (9/8/2019) [1.3.15]
 
 **Fixed bugs:**
--- a/feature/Makefile
+++ b/feature/Makefile
@@ -6,6 +6,8 @@ SRCDIR  = $(TOP)/src
 OBJDIR  = $(TOP)/obj
 TOOLDIR = $(TOP)/tool
 LIBVMAF = $(TOP)/../wrapper/libvmaf.a
+
+# TODO: improve on the circular dependency introduced with feature/src/common, wrapper and libvmaf.h
 LIBVMAF_HEADER = $(TOP)/../wrapper/src
 
 CFLAGS_COMMON = -g -O3 -fPIC -w -Wextra -pedantic
--- a/feature/Makefile
+++ b/feature/Makefile
@@ -6,6 +6,7 @@ SRCDIR  = $(TOP)/src
 OBJDIR  = $(TOP)/obj
 TOOLDIR = $(TOP)/tool
 LIBVMAF = $(TOP)/../wrapper/libvmaf.a
+LIBVMAF_HEADER = $(TOP)/../wrapper/src
 
 CFLAGS_COMMON = -g -O3 -fPIC -w -Wextra -pedantic
 # CFLAGS_COMMON = -g -O0 -fPIC -Wall -Wextra -pedantic
@@ -19,19 +20,19 @@ LDFLAGS  := $(LDFLAGS)
 $(AVX_OBJS): EXTRA_CFLAGS := -mavx
 
 $(OBJDIR)/vmaf_main.o: $(SRCDIR)/vmaf_main.c
-	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) $<
+	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) -I $(LIBVMAF_HEADER) $<
 
 $(OBJDIR)/psnr_main.o: $(SRCDIR)/psnr_main.c
-	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) $<
+	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) -I $(LIBVMAF_HEADER) $<
 
 $(OBJDIR)/moment_main.o: $(SRCDIR)/moment_main.c
-	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) $<
+	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) -I $(LIBVMAF_HEADER) $<
 
 $(OBJDIR)/ssim_main.o: $(SRCDIR)/ssim_main.c
-	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) $<
+	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) -I $(LIBVMAF_HEADER) $<
 
 $(OBJDIR)/ms_ssim_main.o: $(SRCDIR)/ms_ssim_main.c
-	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) $<
+	$(CC) $(EXTRA_CFLAGS) -c -o $@ $(CFLAGS) $(CPPFLAGS) -I $(LIBVMAF_HEADER) $<
 
 vmaf: $(OBJDIR)/vmaf_main.o $(LIBVMAF)
 	$(CXX) -s -o $@ $(LDFLAGS) $^ $(LIBS) -pthread
--- a/feature/src/adm.c
+++ b/feature/src/adm.c
@@ -274,7 +274,7 @@ fail:
 	return ret;
 }
 
-int adm(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, const char *fmt)
+int adm(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h)
 {
     double score = 0;
     double score_num = 0;
--- a/feature/src/common/frame.c
+++ b/feature/src/common/frame.c
@@ -82,7 +82,7 @@ const char* get_fmt_str_from_fmt_enum(enum VmafPixelFormat fmt_enum)
 int read_vmaf_picture(VmafPicture *ref_vmaf_pict, VmafPicture *dis_vmaf_pict, float *temp_data, void *s)
 {
     struct data *user_data = (struct data *)s;
-    bool use_color = user_data->use_color;
+    bool use_chroma = user_data->use_chroma;
     enum VmafPixelFormat format = user_data->format;
     unsigned int w = user_data->width;
     unsigned int h = user_data->height;
@@ -96,20 +96,20 @@ int read_vmaf_picture(VmafPicture *ref_vmaf_pict, VmafPicture *dis_vmaf_pict, fl
     dis_vmaf_pict->w[0] = w;
     dis_vmaf_pict->h[0] = h;
 
-    int color_resolution_ret = get_color_resolution(ref_vmaf_pict->pix_fmt, ref_vmaf_pict->w[0],
+    int chroma_resolution_ret = get_chroma_resolution(ref_vmaf_pict->pix_fmt, ref_vmaf_pict->w[0],
         ref_vmaf_pict->h[0], &(ref_vmaf_pict->w[1]), &(ref_vmaf_pict->h[1]),
         &(ref_vmaf_pict->w[2]), &(ref_vmaf_pict->h[2]));
 
-    if (color_resolution_ret) {
-        fprintf(stderr, "Calculating resolutions for color channels for ref failed.\n");
+    if (chroma_resolution_ret) {
+        fprintf(stderr, "Calculating resolutions for chroma channels for ref failed.\n");
         return 1;
     }
-    color_resolution_ret = get_color_resolution(dis_vmaf_pict->pix_fmt, dis_vmaf_pict->w[0],
+    chroma_resolution_ret = get_chroma_resolution(dis_vmaf_pict->pix_fmt, dis_vmaf_pict->w[0],
         dis_vmaf_pict->h[0], &(dis_vmaf_pict->w[1]), &(dis_vmaf_pict->h[1]),
         &(dis_vmaf_pict->w[2]), &(dis_vmaf_pict->h[2]));
 
-    if (color_resolution_ret) {
-        fprintf(stderr, "Calculating resolutions for color channels for dis failed.\n");
+    if (chroma_resolution_ret) {
+        fprintf(stderr, "Calculating resolutions for chroma channels for dis failed.\n");
         return 1;
     }
 
@@ -172,7 +172,7 @@ int read_vmaf_picture(VmafPicture *ref_vmaf_pict, VmafPicture *dis_vmaf_pict, fl
         return ret;
     }
 
-    if (!use_color)
+    if (!use_chroma)
     {
         // ref skip u and v
         if (ref_fmt == VMAF_PIX_FMT_YUV420P || ref_fmt == VMAF_PIX_FMT_YUV422P || ref_fmt == VMAF_PIX_FMT_YUV444P)
@@ -510,7 +510,7 @@ int get_frame_offset(enum VmafPixelFormat fmt, int w, int h, size_t *offset)
     return 0;
 }
 
-int get_color_resolution(enum VmafPixelFormat fmt, unsigned int w, unsigned int h, unsigned int *w_u, unsigned int *h_u, unsigned int *w_v, unsigned int *h_v) {
+int get_chroma_resolution(enum VmafPixelFormat fmt, unsigned int w, unsigned int h, unsigned int *w_u, unsigned int *h_u, unsigned int *w_v, unsigned int *h_v) {
 
     // check that both width and height are positive
     if ((w <= 0) || (h <= 0))
--- a/feature/src/common/frame.c
+++ b/feature/src/common/frame.c
@@ -322,23 +322,23 @@ fail_or_end:
 int read_frame(float *ref_data, float *dis_data, float *temp_data, int stride_byte, void *s)
 {
     struct data *user_data = (struct data *)s;
-    enum VmafPixelFormat fmt = user_data->format;
+    enum VmafPixelFormat fmt_enum = user_data->format;
     int w = user_data->width;
     int h = user_data->height;
     int ret;
 
     // read ref y
-    if (fmt == VMAF_PIX_FMT_YUV420P || fmt == VMAF_PIX_FMT_YUV422P || fmt == VMAF_PIX_FMT_YUV444P)
+    if (fmt_enum == VMAF_PIX_FMT_YUV420P || fmt_enum == VMAF_PIX_FMT_YUV422P || fmt_enum == VMAF_PIX_FMT_YUV444P)
     {
         ret = read_image_b(user_data->ref_rfile, ref_data, 0, w, h, stride_byte);
     }
-    else if (fmt == VMAF_PIX_FMT_YUV420P10LE || fmt == VMAF_PIX_FMT_YUV422P10LE || fmt == VMAF_PIX_FMT_YUV444P10LE)
+    else if (fmt_enum == VMAF_PIX_FMT_YUV420P10LE || fmt_enum == VMAF_PIX_FMT_YUV422P10LE || fmt_enum == VMAF_PIX_FMT_YUV444P10LE)
     {
         ret = read_image_w(user_data->ref_rfile, ref_data, 0, w, h, stride_byte);
     }
     else
     {
-        fprintf(stderr, "unknown format %s.\n", get_fmt_str_from_fmt_enum(fmt));
+        fprintf(stderr, "unknown format %s.\n", get_fmt_str_from_fmt_enum(fmt_enum));
         return 1;
     }
     if (ret)
@@ -351,17 +351,17 @@ int read_frame(float *ref_data, float *dis_data, float *temp_data, int stride_by
     }
 
     // read dis y
-    if (fmt == VMAF_PIX_FMT_YUV420P || fmt == VMAF_PIX_FMT_YUV422P || fmt == VMAF_PIX_FMT_YUV444P)
+    if (fmt_enum == VMAF_PIX_FMT_YUV420P || fmt_enum == VMAF_PIX_FMT_YUV422P || fmt_enum == VMAF_PIX_FMT_YUV444P)
     {
         ret = read_image_b(user_data->dis_rfile, dis_data, 0, w, h, stride_byte);
     }
-    else if (fmt == VMAF_PIX_FMT_YUV420P10LE || fmt == VMAF_PIX_FMT_YUV422P10LE || fmt == VMAF_PIX_FMT_YUV444P10LE)
+    else if (fmt_enum == VMAF_PIX_FMT_YUV420P10LE || fmt_enum == VMAF_PIX_FMT_YUV422P10LE || fmt_enum == VMAF_PIX_FMT_YUV444P10LE)
     {
         ret = read_image_w(user_data->dis_rfile, dis_data, 0, w, h, stride_byte);
     }
     else
     {
-        fprintf(stderr, "unknown format %s.\n", get_fmt_str_from_fmt_enum(fmt));
+        fprintf(stderr, "unknown format %s.\n", get_fmt_str_from_fmt_enum(fmt_enum));
         return 1;
     }
     if (ret)
@@ -374,7 +374,7 @@ int read_frame(float *ref_data, float *dis_data, float *temp_data, int stride_by
     }
 
     // ref skip u and v
-    if (fmt == VMAF_PIX_FMT_YUV420P || fmt == VMAF_PIX_FMT_YUV422P || fmt == VMAF_PIX_FMT_YUV444P)
+    if (fmt_enum == VMAF_PIX_FMT_YUV420P || fmt_enum == VMAF_PIX_FMT_YUV422P || fmt_enum == VMAF_PIX_FMT_YUV444P)
     {
         if (fread(temp_data, 1, user_data->offset, user_data->ref_rfile) != (size_t)user_data->offset)
         {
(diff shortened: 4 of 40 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Change Log
 
+## (8/27/2019) [1.3.15]
```
slhck: Consider making a minor version bump if the API changes (i.e. bump to 1.4.0).

christosbampis:
Not sure about other users besides the ones you wrote down, how do you suggest doing an RFC?

feature/src/common/frame.c
```diff
@@ -126,9 +126,6 @@ int read_frame(float *ref_data, float *dis_data, float *temp_data, int stride_by
         goto fail_or_end;
     }
 
-    fprintf(stderr, "Frame: %d/%d\r", completed_frames++, user_data->num_frames);
```
HolyWu: Is this intentional?

wrapper/libvmaf.pc
```diff
@@ -5,7 +5,7 @@ includedir=/usr/local/include
 
 Name: libvmaf
 Description: Netflix's VMAF library
-Version: 1.3.14
+Version: 1.3.15
```
HolyWu: 1.4.0

HolyWu: `CHANGELOG.md` and `VERSION` were addressed. But you missed the version in `wrapper/libvmaf.pc` and `wrapper/src/vmaf.cpp`.

HolyWu: By the way, can't you place the version string in `wrapper/src/vmaf.cpp` at the top of the file, or place it in `wrapper/src/vmaf.h`. Placing a version string in the middle of the file looks weird to me, and could easily get missed when bump version.

wrapper/src/libvmaf.h
```diff
@@ -99,23 +133,24 @@ class Result
     double get_score(const std::string &key);
     std::vector<std::string> get_keys();
     void setScoreAggregateMethod(ScoreAggregateMethod scoreAggregateMethod);
+    int get_num_frms();
+    void set_num_frms(int num_frms);
```
kylophone: Not added by your changes, but I've noticed that these cpp classes are internal only. Please remove them from the public header.

wrapper/src/libvmaf.h
```diff
@@ -99,23 +133,24 @@ class Result
...
 private:
     std::map<std::string, StatVector> d;
     ScoreAggregateMethod score_aggregate_method;
+    int num_frms;
```
kylophone: unsigned

wrapper/src/libvmaf.h
```diff
@@ -32,9 +32,43 @@
...
+    int do_ms_ssim;
+    int enable_conf_interval;
+    int use_color;
+    char *format;
+    char *model_path;
+    char *additional_model_paths;
+    char *log_path;
+    char *log_fmt;
+    char *pool_method;
+} VmafContext;
```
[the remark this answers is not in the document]
christosbampis: Fixed.

wrapper/src/libvmaf.h
```diff
@@ -32,9 +32,43 @@
...
+    int n_thread;
+    int n_subsample;
+    int disable_clip;
+    int disable_avx;
+    int enable_transform;
+    int phone_model;
+    int do_psnr;
+    int do_ssim;
+    int do_ms_ssim;
+    int enable_conf_interval;
```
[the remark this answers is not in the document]
christosbampis: used an enum for a number of options such as do_psnr etc.
enable_conf_interval is now part of the VmafModel struct (a bool there)

wrapper/src/libvmaf.h
```diff
@@ -32,9 +32,43 @@
...
+    int use_color;
+    char *format;
```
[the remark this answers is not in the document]
christosbampis: fixed in the API, need to propagate this into other places in the repo

wrapper/src/libvmaf.h
```diff
@@ -32,9 +32,43 @@
...
+    char *model_path;
+    char *additional_model_paths;
```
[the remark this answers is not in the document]
christosbampis: Included the VmafModel type, need to improve on the helper for turning a model path (and associated parameters) into VmafModel

wrapper/src/libvmaf.h
```diff
@@ -32,9 +32,43 @@
...
+    char *log_path;
+    char *log_fmt;
+    char *pool_method;
```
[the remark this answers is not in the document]
christosbampis: Fixed

wrapper/src/libvmaf.h
```diff
@@ -32,9 +32,43 @@
...
+} VmafContext;
+
+struct ModelPredictionContext
```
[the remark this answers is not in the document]
christosbampis: Moved many things out of this file and this context is now removed (VmafModel satisfies the model part)

wrapper/src/libvmaf.h
christosbampis: Moved a number of stuff into vmaf.h, can improve on the cleanup

wrapper/src/libvmaf.h
christosbampis: Fixed. Need to fix other places where w and h are int, perhaps will rely on VmafPicture struct definition with unsigned int there and re-use that.

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,87 @@
...
+#include <stdbool.h>
 
-#ifdef __cplusplus
-}
-#endif
+enum VmafLogFmt {
+    VMAF_LOG_FMT_XML  = (1 << 0),
+    VMAF_LOG_FMT_JSON = (1 << 1),
+    VMAF_LOG_FMT_CSV  = (1 << 2),
+};
```
kylophone: Don't bitshift these enum values unless you intend for them to be or'd and used as a bitfield.

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,87 @@
...
-#include <vector>
-#include <cstring>
-#include <map>
-#include <memory>
-#include <string>
+enum VmafPoolingMethod {
+    VMAF_POOL_MIN                                      = (1 << 0),
+    VMAF_POOL_MEAN                                     = (1 << 1),
+    VMAF_POOL_HARMONIC_MEAN                            = (1 << 2),
+};
```
kylophone: Ditto about the bitfield.

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,87 @@
...
-    double var();
-    double std();
-    void append(double e);
-    double at(size_t idx);
-    size_t size();
-private:
-    std::vector<double> l;
-    void _assert_size();
-};
+    float *data[3];
```
kylophone: `VmafPicture.data` should not be a `float*`?

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,87 @@
...
+    unsigned int w[3], h[3];
```
kylophone: No need to store the width and height of all planes, it should be derived from pix_fmt. Just use `unsigned w, h`.

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,87 @@
...
+    unsigned int stride_byte[3];
```
kylophone: `ptrdiff_t stride[2]`. The stride for cb and cr is always the same.

wrapper/src/libvmaf.h
kylophone: Use `uint8_t*` or `uint16_t*`, or possibly a custom defined type like `pixel*`.

feature/src/common/picture.h
```diff
@@ -0,0 +1,31 @@
...
+#define PICTURE_H_
+
+typedef struct VmafPicture VmafPicture;
+
+struct VmafPicture
+{
+    float *data_y;
+    float *data_u;
+    float *data_v;
+};
```
kylophone: use `data[3]`. Also, `w`, `h`, `pix_fmt`, and `ptrdiff_t stride[2]` needs to be in here also. This container should be the place to hold other attributes like colorspace, color matrix, etc.

feature/src/psnr_tools.c
```diff
@@ -23,14 +23,14 @@
 /*
  * For a given format, returns the constants to use in psnr based calculations
  */
-int psnr_constants(const char *fmt, double *peak, double *psnr_max) {
-    if (!strcmp(fmt, "yuv420p") || !strcmp(fmt, "yuv422p") || !strcmp(fmt, "yuv444p"))
+int psnr_constants(enum VmafPixelFormat fmt, double *peak, double *psnr_max) {
+    if (fmt == VMAF_PIX_FMT_YUV420P || fmt == VMAF_PIX_FMT_YUV422P || fmt == VMAF_PIX_FMT_YUV444P)
```
kylophone: Could be a switch.

feature/src/vmaf_main.c
```diff
@@ -23,12 +23,12 @@
...
-int all(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, const char *fmt);
+int adm(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, enum VmafPixelFormat fmt);
+int ansnr(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, enum VmafPixelFormat fmt);
+int vif(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, enum VmafPixelFormat fmt);
+int vifdiff(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, enum VmafPixelFormat fmt);
+int motion(int (*read_noref_frame)(float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, enum VmafPixelFormat fmt);
+int all(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, enum VmafPixelFormat fmt);
```
kylophone: These should all just take a pair of `VmafPicture`s.

feature/src/common/frame.h
```diff
@@ -19,20 +19,25 @@
...
 {
-    char* format; /* yuv420p, yuv422p, yuv444p, yuv420p10le, yuv422p10le, yuv444p10le */
+    enum VmafPixelFormat format; /* yuv420p, yuv422p, yuv444p, yuv420p10le, yuv422p10le, yuv444p10le */
     int width;
     int height;
     size_t offset;
     FILE *ref_rfile;
     FILE *dis_rfile;
     int num_frames;
+    int use_color;
```
kylophone: why `use_color`?

kylophone: I'm confused by this struct. It's like a copy of the `VmafPicture` struct.

feature/src/common/picture.h
christosbampis: This is just a dangling file that I deleted in a local commit. This file will go away.

wrapper/src/libvmaf.h
christosbampis: The reason I did this is to do the derivation of width and height of all planes in one place (e.g. in read_frame or whoever populates/allocates a VmafPicture). Otherwise you have to go and do the calculation in every feature extractor that is applied on color.

wrapper/src/libvmaf.h
christosbampis: Fixed

wrapper/src/libvmaf.h
christosbampis: Fixed

feature/src/common/frame.h
christosbampis: use_color was used to determine whether the read_frame/read_vmaf_picture should populate or skip the U and V planes. If we decide we always wanna do that, we can remove this field.

christosbampis: This struct is the definition of user_data under the context of read_frame. It does not only have things like width and height as in VmafPicture, it also has the FILE* to ref and dis files.

feature/src/common/picture.h
christosbampis: The definition of VmafPicture (what this file used to refer to) is now placed under libvmaf.h

feature/src/psnr_tools.c
christosbampis: Fixed

wrapper/src/libvmaf.h
```diff
@@ -32,9 +32,43 @@
...
 #endif
 
-int compute_vmaf(double* vmaf_score, char* fmt, int width, int height, int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride_byte, void *user_data),
-				 void *user_data, char *model_path, char *log_path, char *log_fmt, int disable_clip, int disable_avx, int enable_transform, int phone_model, int do_psnr,
-				 int do_ssim, int do_ms_ssim, char *pool_method, int n_thread, int n_subsample, int enable_conf_interval);
+typedef struct {
+    int width;
+    int height;
+    int n_thread;
+    int n_subsample;
```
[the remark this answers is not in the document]
christosbampis: Fixed

wrapper/src/libvmaf.h
christosbampis: Cleanup of the rest of the repo todo.

wrapper/src/libvmaf.h
christosbampis: The stride is defined in terms of the width, if we keep the chroma widths distinct, we can keep the strides also distinct to make that easier to read.

feature/Makefile
```diff
@@ -6,6 +6,7 @@ SRCDIR  = $(TOP)/src
 OBJDIR  = $(TOP)/obj
 TOOLDIR = $(TOP)/tool
 LIBVMAF = $(TOP)/../wrapper/libvmaf.a
+LIBVMAF_HEADER = $(TOP)/../wrapper/src
```
li-zhi: This seems to introduce circular dependency between /wrapper and /feature. Previously it was only /wrapper depends on /feature. Can we avoid this?

feature/src/adm.c
```diff
@@ -274,7 +274,7 @@ int compute_adm(const float *ref, const float *dis, int w, int h, int ref_stride
 	return ret;
 }
 
-int adm(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h, const char *fmt)
+int adm(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride, void *user_data), void *user_data, int w, int h)
```
li-zhi: adm doesn't need `enum VmafPixelFormat fmt_enum`?

christosbampis: yeah, the pixel format was not used inside there

feature/Makefile
christosbampis: True, I have had this in mind. The reason this happens is because read_frame (in feature/common) and feature extractors, need some stuff from libvmaf.h. Perhaps there is a better way to do this.

wrapper/src/libvmaf.h
```diff
@@ -19,6 +19,8 @@
 #ifndef LIBVMAF_H_
 #define LIBVMAF_H_
 
+#define MAX_NUM_VMAF_MODELS 5 // there can be MAX_NUM_VMAF_MODELS - 1 additional models at most
```
kylophone: Why is there a maximum number of models?

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,94 @@
...
+
+    enum VmafPixelFormat pix_fmt;
+    enum VmafLogFmt log_fmt;
+    enum VmafPoolingMethod pool_method;
 
+} VmafSettings;
+
+int compute_vmaf(double* vmaf_score,
+                 int (*read_frame)(float *ref_data, float *main_data, float *temp_data, int stride_byte, void *user_data),
+                 int (*read_vmaf_picture)(VmafPicture *ref_vmaf_pict, VmafPicture *dis_vmaf_pict, float *temp_data, void *user_data),
```
kylophone: Can you write doxygen style API docs for the library functions/types in the header file? I have questions about how to use `compute_vmaf()` for example: 

- Why are there two `read_*` function pointers?
- Why is there only one `vmaf_score` returned, when multiple models are used? What about the confidence interval, etc? Is that available in data somewhere?

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,94 @@
...
+				 void *user_data, VmafSettings *vmafSettings);
+
+unsigned int get_additional_models(char *additional_model_paths, VmafModel *vmaf_model);
```
kylophone: Is this supposed to work with multiple models? or just a single model? What is the format of the `additional_vmaf_model_paths` string supposed to be? `VmafModel` just seems to be a thin wrapper for a filesystem path, not an actual data structure?

Like `compute_vmaf()`, can you add API usage info in the header file?

feature/src/adm.c
kylophone: Feature extractors should just take a pair of `VmafPicture`s? `VmafPicture` is a data structure that holds all of this info already. If we need to add something like colorspace in the future, it just gets added to the `VmafPicture` struct and you don't need to go and add it as an additional function parameter for all N feature extractors.

christosbampis: Yeah, we can propagate the changes from the API and the usage of VmafPicture internally. Perhaps we can make that change later and focus more on the API for now.

wrapper/src/libvmaf.h
christosbampis: Sure, I will be adding some documentation. Regarding the two questions:
1. there are two function pointers just for us to look at the before vs. after state. Will be removing this from the signature soon.
2. So far, the convention has been to return only a single value (vmaf_score) to the caller and write to a log file everything else. We could extend that to return something like a pointer to a result struct. Thinking of ways to do that.

wrapper/src/libvmaf.h
christosbampis: Added documentation in libvmaf.h to cover these questions

wrapper/src/libvmaf.h
christosbampis: This has been done to simplify the generation of the additional models. Will bump it up to 20.

wrapper/src/libvmaf.h
christosbampis: Created a struct VmafOutput which can be extended in the future as we see fit. For now, compute_vmaf is changed to return a pointer to that struct.

feature/Makefile
li-zhi: In this case, move the dependent stuff from libvmaf.h to feature/common to avoid the circular depedency?

feature/src/common/frame.c
```diff
@@ -219,3 +510,64 @@ int get_frame_offset(const char *fmt, int w, int h, size_t *offset)
     return 0;
 }
 
+int get_color_resolution(enum VmafPixelFormat fmt, unsigned int w, unsigned int h, unsigned int *w_u, unsigned int *h_u, unsigned int *w_v, unsigned int *h_v) {
```
li-zhi: you've been calling them "color" (among other functions). I think it is a bit sloppy. Can we use "chroma" instead?

feature/src/common/frame.c
```diff
@@ -219,3 +510,64 @@ int get_frame_offset(const char *fmt, int w, int h, size_t *offset)
...
+
+    // check that both width and height are positive
+    if ((w <= 0) || (h <= 0))
+    {
+        fprintf(stderr, "Width or height is not positive, width = %d, height = %d.\n", w, h);
+        return 1;
+    }
+
+    if (fmt == VMAF_PIX_FMT_YUV420P || fmt == VMAF_PIX_FMT_YUV420P10LE)
```
li-zhi: Refactor to use switch? just keep the format consistent with other refactors.

wrapper/src/combo.c
```diff
@@ -72,8 +73,9 @@ void* combo_threadfunc(void* vmaf_thread_data)
     int h = thread_data->h;
     char* errmsg = thread_data->errmsg;
     void* user_data = thread_data->user_data;
-    const char* fmt = thread_data->fmt;
-    int n_subsample = thread_data->n_subsample;
+    enum VmafPixelFormat fmt = thread_data->fmt;
+    unsigned int n_subsample = thread_data->n_subsample;
+    bool use_color = thread_data->use_color;
```
li-zhi: Would `use_color` also be added to all.c, and other functions like ssim.c, psnr.c? I am debating if adding it here now is pre-mature optimization. Clearly, with this you can do one pass and compute for all Y, U and V, and this is more efficient. But consider the alternative: do 3 passes of the video file, each one use a different read_frame callback function to pass Y, U and V. This is less efficient for now, but will result in less boiler-plate code. If we do this, we can leave speed optimization to the future.

wrapper/src/libvmaf.cpp
```diff
@@ -287,4 +249,124 @@ extern "C" {
             return -4;
         }
     }
+
+    void replace_string_in_place(std::string& subject, const std::string& search,
```
li-zhi: indentation issue.

wrapper/src/combo.c
kylophone: Why not just ask for each feature explicitly? `psnr_y`, `psnr_cb`, `psnr_cr` ...

Using a global `use_color` bool is a little confusing. Which features have color? Do I really have to extract all of them if I set this?

wrapper/src/libvmaf.h
```diff
@@ -32,92 +34,140 @@
...
+  * The additional models are passed in using the additional_model_paths string.
+  * This string contains model-specific parameters for each additional model (path, disable_clip, etc.)
+  * This string should be formatted as an escaped json-like string with the following format:
+  * {\"<name>\":{\"model_path\"\:\"<path>\"}}, e.g.:
+  * {\"another_model\":{\"model_path\"\:\"some/path/to/some/model/another_model.pkl\"}}
+  * The fields <name> and <path> are required and optional parameters can be used by adding an escaped comma, e.g.:
+  * {\"<name>\":{\"model_path\"\:\"<path>\"\,\"disable_clip\"\:\"1\"}} --> disable clipping of predictions
+  * {\"<name>\":{\"model_path\"\:\"<path>\"\,\"enable_transform\"\:\"1\"}} --> use transformation for the predictions (e.g. phone model transformation)
+  * {\"<name>\":{\"model_path\"\:\"<path>\"\,\"enable_conf_interval\"\:\"1\"}} --> calculate confidence interval (CI) for the predictions (the model should support CIs)
+  * {\"<name>\":{\"model_path\"\:\"<path>\"\,\"disable_clip\"\:\"0\"\,\"enable_transform\"\:\"1\"}} --> enable clipping and transform scores
```
kylophone: If someone is using this as a C library they need to allocate a` char*` with enough space to hold a properly escaped and formatted JSON string?

Could this design be better?

wrapper/src/libvmaf.h
kylophone: > Created a struct VmafOutput which can be extended in the future as we see fit. For now, compute_vmaf is changed to return a pointer to that struct.

What about the scores from the additional models?

feature/src/common/frame.c
christosbampis: Fixed all "color" to "chroma"

wrapper/src/libvmaf.h
christosbampis: In that case one can create a VmafModel similar to how run_wrapper inits the first model.

christosbampis: vmafSettings->vmaf_model[0].name = "vmaf";
    vmafSettings->vmaf_model[0].path = model_path;
    vmafSettings->vmaf_model[0].vmaf_model_setting = vmaf_model_setting;

wrapper/src/combo.c
christosbampis: 2.

The global nature of use_color is related to the fact that we have one read_frame. We can rename this to "read_color". In fact, "read_color" is not tied to a specific feature, it just means that you have read the color planes and have them available for extracting color features. I am now adding further logic to decouple the two: extract psnr color when the corresponding feature field is also turned on, i.e., read_color && DO_PSNR_U.

wrapper/src/libvmaf.cpp
christosbampis: Fixed

feature/src/common/frame.c
christosbampis: Fixed

wrapper/src/combo.c
christosbampis: Update: let's actually get rid of use_color, read_vmaf_picture already can have that data there. It is up to the caller if a feature with chroma has to be calculated. For PSNR, it is a field: VMAF_FEATURE_MODE_SETTING_DO_CHROMA_PSNR. Similar fields can be added in the future if we want chroma in other features. This will also simplify read_vmaf_picture.

feature/Makefile
christosbampis: Adding a todo about that inside.

wrapper/src/combo.c
christosbampis: Fixed this.

wrapper/src/libvmaf.h
christosbampis: The scores from additional models, along with other data that we want to return can be part of that struct.

feature/src/vmaf_main.c
christosbampis: We can propagate the changes to the feature/src apps, but for the purposes of the API change we can focus on that. For now, we can keep having read_frame for them, since combo (and the wrapper code) is not affected by that.

wrapper/src/libvmaf.h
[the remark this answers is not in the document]
christosbampis: Currently, VmafModel is just a thin wrapper for a filesystem model. Internally there is another class (LibsvmNusvrTrainTestModel) which has further helpers for reading in the actual file and assigning a set of model-relevant fields.

feature/src/common/frame.h
christosbampis: As of the latest commits, use_color has now been removed, and we explicitly ask for a chroma feature, like PSNR using the VmafFeatureModeSetting enum.

christosbampis:
> I get that either 5 or 20 would probably be OK for most cases. The point of my question was more about API design: why does this have a hardcoded upper limit?

This was done to simplify the memory allocation for VmafModels. The API is not directly affected by that since the settings have an array of VmafModels.

kylophone:
closing this for now

PR #362, opened 2019-09-14
Removed unsuppressable printf statements.

colbyn:
Small change, just removed a few printf statements. It’s completely worthless when there are multiple instances running in parallel. Also when used VIA ffmpeg it can’t be suppressed via `-loglevel` parameter.


li-zhi:
Not entirely sure if it is worth removing them. Could they be redirected to /dev/null?

colbyn:
No, not that I’m aware of (when invoked from the C API).

Also it gets in the way with other reporting stuff like rendering a progress bar to stdout.

kylophone:
Fine to remove these for now, I think. I've created an issue to track a more proper fix: #365.

colbyn:
Thanks!

PR #366, opened 2019-09-19
Templatize libvmaf.pc

### LINKED ISSUE #221: libvmaf.pc should be templatized
`libdir` and `prefix` should match the variables passed by `make`.

rswarbrick-argon:
This fixes #221.


kylophone:
lgtm

This can be merged as-is, but I wonder,  can `Version:` be templatized somehow too?

Thanks!

PR #377, opened 2019-10-09
Add dynamic library for libvmaf.

christosbampis:


kylophone:
I have another patch coming later today which will also provide a dylib. What does this test check for, linkability?

christosbampis:
Yeah, it checks for linkability for the dynamic library. We can cherry pick that test file.

src/libvmaf/Makefile
```diff
@@ -1,8 +1,10 @@
 LIBVMAF = libvmaf.a
+LIBVMAF_DYN = libvmafdyn.so
```
li-zhi: could we just call it `libvmaf.so`?

christosbampis: Yeah, we can rename if we want to. Just gave this name for simplicity to avoid confusion with libvmaf.a when doing -lvmaf.

kylophone: The suffix for these is platform dependent. `.dylib` on MacOS or `.dll` on Windows for example.

See line 3b here: http://web.archive.org/web/20110720003722/http://www.fortran-2000.com/ArnaudRecipes/sharedlib.html

It is better to use a build system like `CMake` or `meson` to handle this. I will be sending a PR soon to address this.

christosbampis: sure thing, that will help make building cleaner!

PR #378, opened 2019-10-09
Cmake: Add a CMakeLists

### LINKED ISSUE #311: Provide CMakeLists
For building & compiling with different compilers a CMakeLists.txt would be very useful.

1480c1:
Closes #311 

I've only added the basic parts for it and only did what `make all` and `make install` does. Nothing else special added since this is just a basic one.

It's possible to add the other executables and stuff, but it seemed very unwieldy to do with the current file structure.

Went with a single file list since it seems that there were a lot
of going in between folders in the makefiles, which was pretty
ugly imo since it was hard to actually tell what line meant what.

Later on, I might try to improve it

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,176 @@
+cmake_minimum_required(VERSION 3.5)
+project(vmaf VERSION 1.3.15 LANGUAGES C CXX)
+
+include(GNUInstallDirs)
+include(CheckCCompilerFlag)
+include(CheckCXXCompilerFlag)
+find_package(Threads)
+
+set(CMAKE_C_STANDARD 99)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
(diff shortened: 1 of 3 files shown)


1480c1:
I wish it was a cake list

kylophone:
I would wait a minute on this. I have another patch on the way which reorganizes some of the lib code, and uses meson as a build system.

1480c1:
If all of the features can be implemented in the meson build system, then that would be great as that could mean that cmake doesn't have to be added since most of the things cmake can do, meson can also do, or the cmake part can be based on the meson.build file.

Although personally I don't have much knowledge about meson since I've never really had much of a good experience with python on windows in general

kylophone:
> I would wait a minute on this. I have another patch on the way which reorganizes some of the lib code, and uses meson as a build system.

#379

kylophone:
> > I would wait a minute on this. I have another patch on the way which reorganizes some of the lib code, and uses meson as a build system.
> 
> #379

Replaced by #384.

kylophone:
#384 merged.

Issue #387, opened 2019-11-06
Errors running make installing VMAF Python Library on macOS


jayholman:
It looks like there might be some active work on this regarding `meson`, but I wanted to note a couple errors I ran into when running `make`, it failed because it was missing both the `meson` package and `ninja` dependency. I was working off the these instructions: https://github.com/Netflix/vmaf/blob/master/resource/doc/VMAF_Python_library.md#macos and I'm currently on mac OS `10.15.1`.

I solved the issue with`ninja` by simply doing `brew install ninja`. 

`meson` is not quite as simple as `Meson requires Python 3.5.2 or greater` and you will encounter an error message attempting `pip install meson`: 
<img width="685" alt="image" src="https://user-images.githubusercontent.com/15311387/68334127-f5b46480-00a7-11ea-843f-8558cb7f9e58.png">
To avoid that error, I used `pip3 install meson`, which made it possible to complete the `make` installation.

I'm happy to open a PR with those changes added to the macOS instructions if that would be helpful, but it also seems like there is work around that `make` file going on in the most recent closed PRs.

christosbampis:
Sure, feel free to make a PR with those changes in the documentation for mac os.

kylophone:
Closing, should be addressed by #388.

PR #388, opened 2019-11-07
Update macOS instructions to include ninja and meson

jayholman:
This adds the `ninja` dependency for in the brew step and the `meson` package for Python in the macOS instructions. They make it is possible to run `make` in the VMAF repository. The `meson` package requires Python 3.5.2 or greater, which is why it is using `pip3`. This is meant to solve #387.


christosbampis:
Given that pip3 might not be already installed, wanna add that in?

kylophone:
Ah, great. Merged. Thanks!

PR #389, opened 2019-11-08
fix docker image after using meson for building

edmond-zhu:
docker build broke after meson build added
this commit is to fix docker build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,36 @@
-FROM ubuntu
-RUN apt-get update -qq
+FROM ubuntu:18.04
 
-RUN apt-get install -y build-essential git
+# setup timezone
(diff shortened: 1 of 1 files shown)


slhck:
AFAICT, the only real change here is the `ninja-build` dependency. The rest is just formatting/style changes, correct?

kylophone:
Thank you. Merged.

edmond-zhu:
> AFAICT, the only real change here is the ninja-build dependency. The rest is just formatting/style changes, correct?

I'm sorry for not reply in time, I was not with my laptop during last weekend.
The issue was python 3 is required by meson, but the older Dockerfile only installed python 2, meson reported error when installing.
This fix upgraded python and adjust a little for the format and style.

PR #391, opened 2019-11-11
Bugfix/fix remaining tests for py3

li-zhi:


christosbampis:
Thanks for these changes, will be making some more on top of this branch to finish the todos and then we can merge to master.

PR #394, opened 2019-11-13
meson: fix pkg-config file name

quink-black:


quink-black:
1. The rename of libvmaf.pc breaks ffmpeg building
[ffmpeg-dev](http://ffmpeg.org/pipermail/ffmpeg-devel/2019-November/252848.html)

2. [meson build: thread dependency](https://mesonbuild.com/howtox.html#enable-threads)

kylophone:
Merged. Thanks.

Issue #385, opened 2019-11-01
Implement VmafPicture


kylophone:
Let's use this issue to track the integration of `VmafPicture`.  `VmafPicture` is a type used in the new API, which is designed as a container for pixel data, as well as all associated metadata. Roughly, it should look something like this. If you have any feedback on the design please leave it here.

```c
#ifndef __VMAF_PICTURE_H__
#define __VMAF_PICTURE_H__

#include <stddef.h>

enum VmafPixelFormat {
    VMAF_PIX_FMT_UNKNOWN,
    VMAF_PIX_FMT_YUV420P,
    VMAF_PIX_FMT_YUV422P,
    VMAF_PIX_FMT_YUV444P,
};

typedef struct {
    enum VmafPixelFormat pix_fmt;
    unsigned w, h;
    ptrdiff_t stride[2];
    pixel *data[3];
    int bpc;
} VmafPicture;

int vmaf_picture_alloc(VmafPicture *pic, enum VmafPixelFormat pix_fmt,
                       unsigned bpc, unsigned w, unsigned h);

void vmaf_picture_unref(VmafPicture *pic);

#endif /* __VMAF_PICTURE_H__ */

```

kylophone:
Question for @christosbampis, do we want to add other things like `.colorspace`, `.colormatrix`, etc. to `VmafPicture` ? Could these be useful down the line?

christosbampis:
Sure, they can be useful later, e.g., for HDR where we might want to capture relevant information.

kylophone:
#397 is merged, so closing this issue.

PR #397, opened 2019-11-18
Implement VmafPicture

kylophone:
As initially discussed here: #385 

The first commit in this PR adds a unit testing framework for use within `libvmaf`. The third commit just moves the aligned_malloc implementation from the features directory to the src root.

The remainder of this PR implements an API for creating, referencing, and unreferencing`VmafPicture`.  Also a couple of unit tests.

- `VmafPicture` is init'd with `vmaf_picture_alloc()`, this includes all memory allocations.
- Pixel data (each line) is always aligned to a 32-byte memory address.
- `VmafPicture` can hold 8-bit pixel data as well as high bit depth pixel data. 
- `VmafPicture` can be ref'd and unref'd. When a `VmafPicture` is referenced, pixel data buffers are shared. This is useful to conserve memory as well as to increase cache efficiency.
- `VmafPicture` picture data is automatically free'd when `ref_cnt` goes to 0.
-  `VmafPicture` reference counting is thread safe.

Will likely merge this tomorrow. At this point, `picture.h` is not a published header, nor is `picture.c` compiled into the library.


kylophone:
@li-zhi, @christosbampis: I will probably merge this later today. Let me know if there are any comments or questions.

PR #398, opened 2019-11-21
[WIP] libvmaf: add VmafFeatureCollector

kylophone:
This adds a new internal struct `VmafFeatureCollector`, as well as an API for interacting with it. The  purpose of this is to collect extracted features on a per frame basis. See `feature_collector.h` and `test_feature_collector_init_append_and_destroy` for usage.

`VmafFeatureCollector` has the following attributes:
- Hold scores for an extracted feature with any `feature_name`
- Thread safe
- OOO insertion, for feature extraction parallelism on a frame level
- Avoid this mess: https://github.com/Netflix/vmaf/blob/master/libvmaf/src/combo.h#L40-L63
- Add a new feature extractor to `libvmaf` with no internal changes.

I've marked this as **WIP** for now, since I don't exactly know when the best time to integrate it is. Also there still needs to be an API to get scores out from the `VmafFeatureCollector`. Depending on how this gets used, it may be necessary to adjust the internal data structure.

### DIFF
--- a/libvmaf/src/feature/feature_collector.c
+++ b/libvmaf/src/feature/feature_collector.c
@@ -1,0 +1,202 @@
+#include <errno.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "feature_collector.h"
+
+typedef struct {
+    char *name;
+    struct {
+        bool written;
+        double value;
+    } *score;
+    unsigned capacity;
+} FeatureVector;
+
+typedef struct VmafFeatureCollector {
+    FeatureVector **feature_vector;
+    unsigned cnt, capacity;
+    pthread_mutex_t lock;
+} VmafFeatureCollector;
+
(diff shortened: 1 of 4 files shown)


li-zhi:
Sounds functionally similar to the FeatureAssembler class in Python. Would be nice to match the class name.
https://github.com/Netflix/vmaf/blob/master/python/src/vmaf/core/feature_assembler.py#L10

kylophone:
> Sounds functionally similar to the FeatureAssembler class in Python. Would be nice to match the class name.
> https://github.com/Netflix/vmaf/blob/master/python/src/vmaf/core/feature_assembler.py#L10

OK.

kylophone:
>  there still needs to be an API to get scores out from the `VmafFeatureCollector`. 

Done.

> Sounds functionally similar to the FeatureAssembler class in Python.

I'm not going to rename this at this point since this is only a score collector, not a class with runner like the class in Python.  This is not planned to be part of any public API, so this could be easily be renamed later if there is a better name for it.

Issue #383, opened 2019-10-31
[RFC] VMAF models, new data format


kylophone:
VMAF models are currently published as a pickled Python dictionary. For `libvmaf`, this adds a pretty large dependency (`ptools`) which itself also requires `pthreads` for some reason. My proposal for libvmaf v2.0 API (see #381) is to pick a more lightweight data format for published VMAF models. JSON probably makes the most sense, but it could always be something else. Please leave your comments here if you have an opinion on this.

slhck:
Definitely +1 for JSON.

kylophone:
https://github.com/Netflix/vmaf/issues/324#issuecomment-548508736

kylophone:
> See also: [#324 (comment)](https://github.com/Netflix/vmaf/issues/324#issuecomment-548508736)

@slhck, do you know of any other packagers that currently provide libvmaf?

slhck:
Not that I'm aware of. Linuxbrew and Homebrew as far as I know.

kylophone:
Just to demonstrate how large of a dependency `ptools` is: 

```
github.com/AlDanial/cloc v 1.76  T=1.06 s (202.8 files/s, 64293.8 lines/s)
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
C++                            110           5175           3768          23013
C/C++ Header                   101           5825           8545          21701
C Shell                          2             25             31             68
XML                              1              0              0             10
make                             1              0              0              1
-------------------------------------------------------------------------------
SUM:                           215          11025          12344          44793
-------------------------------------------------------------------------------
```

Also, it complies with 90 warnings on my machine.

kylophone:
@christosbampis, can you please convert our currently published models to a JSON format?

- [ ] Provide a JSON schema for the new format (https://json-schema.org/)
- [ ]  Convert the currently published models to this format
- [ ] Create (or reuse) a Python dict to read the data into, provide a Python function to do this reading
- [ ] Create a C structure (`VmafModel`) to hold this data

christosbampis:
sure, on it!

christosbampis:
Here is a draft for the json schema:

schema = {
	"type": "object",
	"properties": {

		"svm_type": {"type": "string"},
		"kernel_type": {"type": "string"},
		"classes_num": {"type": "number"},
		"support_vector_num": {"type": "number"},

		"gamma": {"type": "number"},
		"rho": {"type":  "number"},
[...]

PR #408, opened 2019-12-06
pkg-config: Add libvmaf as a subdir to actually point to the correct includedir

1480c1:
FFmpeg configuration on current master branches fails since pkg-config isn't giving `${prefix}/include/libvmaf` to find libvmaf.h

Changes Cflags line in libvmaf.pc file from
`Cflags: -I${includedir}`
to
`Cflags: -I${includedir}/libvmaf`

### DIFF
--- a/libvmaf/src/meson.build
+++ b/libvmaf/src/meson.build
@@ -134,5 +134,6 @@ pkg_mod.generate(libraries: libvmaf,
     version: meson.project_version(),
     name: 'libvmaf',
     filebase: 'libvmaf',
-    description: 'VMAF, Video Multimethod Assessment Fusion'
+    description: 'VMAF, Video Multimethod Assessment Fusion',
+    subdirs: 'libvmaf'
 )


1480c1:
I am currently working on a travis-ci job for testing with ffmpeg builds, but it's currently blocked by this.

I'm also working on appveyor, but I'm not sure how I should handle the lack of `pthread.h` on windows, I usually handle it personally by using winpthread with msys2, but I don't think msvc will pick it up automatically.

1480c1:
Ping @kylophone

kylophone:
> Ping @kylophone
Sorry, had to take a second to give this a closer look. Merged, thanks.

kylophone:
> I'm also working on appveyor, but I'm not sure how I should handle the lack of `pthread.h` on windows, I usually handle it personally by using winpthread with msys2, but I don't think msvc will pick it up automatically.

Cool. Looking forward to that. If we kill `ptools` (#383) , then we can just provide a simple `pthread` to `Win32` wrapper and then the problem will go away completely.

PR #409, opened 2019-12-06
Fix n_estimators in test for extra trees test to allow sklearn>=0.22.

christosbampis:


christosbampis:
The default number of estimators in Extra Trees classifier of sklearn changed to 100 from 10 when sklearn>=0.22. By changing the test to specify the number of estimators, requirements can now be changed back to have any sklearn version >=0.18.

PR #411, opened 2019-12-09
Travis: Test multiple compilers and FFmpeg integration

1480c1:
Add GCC-7,8,9 and Clang-10 compile tests along with keeping the python tests and add a ffmpeg compile test.

@kylophone, is vmaf targeting a minimum compiler version? I don't particularly see any mention of anything like that.

Questions:
- Should I add a macOS build job?
- Should I test if running vmaf through ffmpeg works?
- Should I remove the python2 tox job? https://pythonclock.org/
- Would you prefer these to be in Travis-CI or in Github Actions?

Anything else you think I should add?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -89,5 +89,17 @@ matrix:
       script:
         - git clone https://github.com/FFmpeg/FFmpeg --branch master --depth=1 ffmpeg && cd ffmpeg
         - "sudo chown -R travis: $HOME/.ccache"
-        - ./configure --enable-libvmaf --enable-version3 || less ffbuild/config.log
+        - CFLAGS="-static" ./configure --enable-libvmaf --enable-version3 || less ffbuild/config.log
         - make --quiet -j $(nproc) && sudo make install
+        - curl https://gist.githubusercontent.com/1480c1/0c4575da638ef6e8203feffd0597de16/raw/akiyo_cif.tar.xz.base64 | base64 -d | tar xJ
+        - export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib/x86_64-linux-gnu
+        - vmaf_score=$(ffmpeg -hide_banner -nostats -i encoded.mkv -i orig.mkv -filter_complex libvmaf -f null - 2>&1 | grep 'VMAF score' | tr ' ' '\n' | tail -n1)
+        - echo "$vmaf_score"
+        - |
+          if [[ $vmaf_score != "93.665821" ]]; then
+            echo "vmaf Score doesn't match 93.665821"
+            return 1
+          else
+            echo "vmaf score matches"
+            return 0
+          fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,31 +1,93 @@
-language: python
+os: linux
 dist: bionic
-python:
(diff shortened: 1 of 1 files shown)


kylophone:
Looks good. Thanks! I can merge this after we sort through all your questions.

> Add GCC-7,8,9 and Clang-10 compile tests along with keeping the python tests and add a ffmpeg compile test.

I'm not sure we need all of those versions of GCC, but it is nice to have. Perhaps we could eliminate the older versions if overall test time ever becomes an issue.

> 
> @kylophone, is vmaf targeting a minimum compiler version? I don't particularly see any mention of anything like that.

Formally, not really.

> 
> Questions:
> 
> * Should I add a macOS build job?

That'd certainly be nice to have.

> * Should I test if running vmaf through ffmpeg works?

Yes. I'd like to know if we're breaking it, because we've unintentionally done so in the past.

> * Should I remove the python2 tox job? https://pythonclock.org/

@li-zhi? @christosbampis?

> * Would you prefer these to be in Travis-CI or in Github Actions?

Since it's already set up, Travis is good for now. If there's a good reason to move, let me know.

1480c1:
> I'm not sure we need all of those versions of GCC, but it is nice to have. Perhaps we could eliminate the older versions if overall test time ever becomes an issue.

 I personally am of the opinion that you should only really test the version that's on a major distro (GCC 7 for the case of Ubuntu 18.04) and the latest (GCC 10), so I can probably eliminate GCC 9 build since it doesn't seem necessary to test the versions in between



> > Questions:
> > 
> > * Should I add a macOS build job?
> 
> That'd certainly be nice to have.

Okay, will replace the GCC-9 build with a macOS clang one



> Yes. I'd like to know if we're breaking it, because we've unintentionally done so in the past.

Sure, I'll try to get a reproducible way to test vmaf in travis

> If there's a good reason to move, let me know.

afaik, none other than GitHub integration and third-party Github Actions, but imo, the caching sucks to use compared to others like travis and appveyor

1480c1:
Don't merge yet, found some problems

1480c1:
Okay, done, added a ffmpeg vmaf filter test

christosbampis:
yeah, we can remove the tox for py2 (py3 support should be mostly complete).

1480c1:
> yeah, we can remove the tox for py2 (py3 support should be mostly complete).

Done

.travis.yml
```diff
@@ -1,31 +1,98 @@
...
+      cache: &python-cache
+        pip: true
+        directories:
+          - $HOME/.ccache
+      before_script: *base-gcc-script
+      script: &python-tox
+        - pip3 install tox-travis
+        - make -C third_party/libsvm -j lib
+        - tox -c python/ -- -v -p no:warnings -m 'main or lib' --doctest-modules
+    - name: libvmaf FFmpeg
```
li-zhi: This looks good for now. I may later move this into a python test, to be aligned with the rest of the unittests.

1480c1: Sure, that would be good if possible, I'm not sure how "stable" my "test" is with just comparing the vmaf output to a previous generated number

kylophone:
Merged. Thanks.

PR #422, opened 2019-12-17
Update vmafossexec build instruction

ltrudeau-twoorioles:
vmafossexec is now built using meson instead of make


kylophone:
Thanks, @ltrudeau-twoorioles. Merged.

Issue #413, opened 2019-12-11
Ffmpeg+libvmaf examples


christosbampis:
Placeholder to add some documentation on ffmpeg+libvmaf examples.

slhck:
Do you need some input/help on this? There are a few lines here: https://github.com/Netflix/vmaf/blob/master/resource/doc/libvmaf.md#use-libvmafa-with-ffmpeg – but there could of course be more examples, e.g. on how to parse the stats or use different options/models.

li-zhi:
We have some recent experience with this and will share a page soon. The
key things to use FFmpeg+libvmaf is to make sure 1) PTS is correctly
aligned and 2) the correct upsampling algorithm is used (need to explicitly
specify bicubic upsampling, since by default bilinear is used).

christosbampis:
Sure, all help and input is useful! We will make a PR with the documentation and then feel free to suggest improvements and give feedback.

1480c1:
Perhaps for windows ffmpeg+libvmaf, someone could mention that the `model_path` needs to be in a "mixed mode" like `libvmaf=model_path=C\\:/vmaf/share/model/vmaf_v0.6.1.pkl` with the `:` escaped and the `\` escaped to escape the `:` for ffmpeg to accept it as an argument to model_path instead of splitting it

slhck:
@1480c1 Good idea. See the open PR #423 where I already left a few comments.

slhck:
@1480c1 Do you have any idea what could cause the issue here? https://github.com/slhck/ffmpeg-quality-metrics/issues/10

We've tried all kinds of combinations of specifying the VMAF model path under Windows, when called from Python, but it's not working. What sort of escaping needs to be used when generating the command line with, say, `C:/test.pkl` from Python?

1480c1:
hmm

there's a few issues I could see, and some ways I can think of "fixing" them.

assuming the average windows user doesn't know about needing to escape parts of a path in windows and they may try doing `Copy as path` or tabbing a whole path, you may first want to resolve the pkl file to an absolute path, that way you are only going to be doing any string manipulation on a sort of known format.

Next, then open using the `$abspath` to check, also check `${abspath}.model`

next, after you confirmed that python's standard functions can fopen them correctly, then convert the windows path to a "mixed" path using `/` as the path sep. I personally don't do any python scripting because of how convoluted the distributions are on windows and prefer bash, so I am not aware of any standard function in python, but in terms of bash, it might be as simple as 
[...]

PR #423, opened 2019-12-21
Add some more ffmpeg+libvmaf examples.

christosbampis:


resource/doc/libvmaf.md
```diff
@@ -54,9 +54,58 @@ Using FFmpeg with libvmaf is very powerful, as you can create complex filters to
...
+See the [FFmpeg's guide to libvmaf](https://ffmpeg.org/ffmpeg-filters.html#libvmaf), the [FFmpeg Filtering Guide](https://trac.ffmpeg.org/wiki/FilteringGuide) for more examples of complex filters, and the [Scaling Guide](https://trac.ffmpeg.org/wiki/Scaling) for information about scaling and using different scaling algorithms. Please refer to the next section for some additional examples on more complex cases.
+
+### Use `libvmaf.a` with FFmpeg - More examples
+
+The above command line works well for most cases, but additional steps might be necessary.
+
+For example, when the frame rates between reference and encoded video do not match, e.g., when the source is 60 fps and the encode is 30 fps:
+
+```
+ffmpeg -r 1 -i main.mpg -r 2 -i ref.mpg -filter_complex \
```
slhck: This seems like a weird choice for forcing different frame rates.

Would you not recommend using the fps filter to (deterministically) upsample the framerate of the encoded video?

That is:

```
ffmpeg -i main.mpg -i ref.mpg -filter_complex \
"[0:v]scale=1920x1080:flags=bicubic,fps=fps=60[main];[main][1:v]libvmaf" -f null -
```

In all pixel-based full reference models I've used so far, it was more common to upsample framerates to the reference rather than doing the opposite.

Or would you recommend not using VMAF in this way since it cannot explicitly handle framerate differences between source/degraded video?

resource/doc/libvmaf.md
```diff
@@ -54,9 +54,58 @@ Using FFmpeg with libvmaf is very powerful, as you can create complex filters to
...
+Here we explicitly state that the reference video has double the frame rate and also use the select filter to select half of the reference frames to calculate VMAF.
+
+If we want a specific model to be used (e.g. the `baz.pkl` model file under the `/foo/bar/` directory):
+
+```
+ffmpeg -i main.mpg -i ref.mpg -filter_complex \
+"[0:v]scale=1920x1080:flags=bicubic,setpts=PTS-STARTPTS[main];[1:v]setpts=PTS-STARTPTS[ref];[main][ref]libvmaf='model_path=/foo/bar/baz.pkl'" -an -f null -
+```
+
+If we want to specify a model that relies on bootstrapping (e.g. the 0.6.3 model file under the `/foo/bar/` directory), the `enable_confidence_interval` option needs to be set:
```
slhck: I think it would be good to mention the default path where models are installed (`/usr/local/share/model`) in addition to just `/foo/bar`. A novice user will not know where these models are, or whether they need to be downloaded separately etc.

resource/doc/libvmaf.md
```diff
@@ -54,9 +54,58 @@ Using FFmpeg with libvmaf is very powerful, as you can create complex filters to
...
+```
+
+If we want to enable the calculation of an additional metric, like SSIM:
+
+```
+ffmpeg -i main.mpg -i ref.mpg -filter_complex \
+"[0:v]scale=1920x1080:flags=bicubic,setpts=PTS-STARTPTS[main];[1:v]setpts=PTS-STARTPTS[ref];[main][ref]libvmaf='ssim=1:model_path=/foo/bar/baz.pkl'" -an -f null -
+```
+
+If we want use a single thread:
```
slhck: Would it make sense to explain why someone would want that? Is multithreaded computation non-deterministic, like with some video encoders?

resource/doc/libvmaf.md
```diff
@@ -54,9 +54,58 @@ Using FFmpeg with libvmaf is very powerful, as you can create complex filters to
...
 ```
 
-Here `main.mpg` is a downscaled and encoded video and `ref.mpg` is its reference source at 1080p. The command scales the first input video (`0:v`) and forwards it to VMAF (`libvmaf`) with the label `main`, where it is compared against the second input reference video, `1:v`. Bicubic upsampling is used (also see the [techblog](https://medium.com/netflix-techblog/vmaf-the-journey-continues-44b51ee9ed12) for the recommendation on upsampling methods).
+Here `main.mpg` is a downscaled and encoded video and `ref.mpg` is its reference source at 1080p. The command scales the first input video (`0:v`) and forwards it to VMAF (`libvmaf`) with the label `main`, where it is compared against the label `ref` which corresponds to the second input reference video, `1:v`. We enforce the scaling algorithm to bicubic, since the default scaling algorithm is bilinear (also see the [techblog](https://medium.com/netflix-techblog/vmaf-the-journey-continues-44b51ee9ed12) for the recommendation on upsampling methods). Note that we need to specify the `setpts=PTS-STARTPTS` option to ensure that the PTS is correctly
```
slhck: In which cases is this strictly necessary? I don't think I've come across alignment issues yet, but probably these runs were always using the same starting PTS for source/distorted video.

1480c1:
@christosbampis, could you add a windows example? Particularly on how to set the `model_path` argument

resource/doc/libvmaf.md
christosbampis: This is a good question. There are two reasons why this example mentions -r 1 -r 2. 

The first reason is that the input to the encoder is typically the SFR (standard frame rate) source and not the HFR (high frame rate) one, which is then encoded to as an SFR encode. In this case, calculating VMAF for the SFR encode would mean to compare it with the source used to encode it (SFR version of source). The underlying assumption here is that the HFR->SFR conversion is done by dropping frames and hence the select() operation. If we know exactly the temporal downsampling operation/filter being used, that operation can be used instead. Also, note that the VMAF score represents the quality of the encode with respect to its SFR source.

The second reason is that I found it actually much safer to always include -r 1 and -r 1 when the source and the encode are both SFR or HFR, because in some cases, e.g., image-based sources where we feed in the source frames, ffmpeg will assume a nominal fps of 25, which does not match the encode's fps. Other combinations of specifying the fps as intended etc. did not match the expected results.

Long story short, I think it is safer to always include -r 1 and -1 and if there is an HFR->SFR conversion for the encode to use the -r 2 for the source.

However, I do realize that this example might be very confusing to users, since it entails a number of choices and very special cases (calculate VMAF at SFR version of source and encode rather than HFR and the image-based sources case), so we might as well remove this example.

resource/doc/libvmaf.md
christosbampis: Noted, will fix that.

resource/doc/libvmaf.md
christosbampis: The purpose of this example is to just showcase how the user can control the number of threads. One use-case is to compare the execution time of a single-threaded vs. a multi-threaded execution.

resource/doc/libvmaf.md
christosbampis: To be on the safe side, I would always include this option, since if not included (and there is misalignment), the scores will change.

resource/doc/libvmaf.md
christosbampis: I can add a comment in the docs of the above use case.

resource/doc/libvmaf.md
slhck: Thanks for explaining the reasoning. To me the choice of specifing an unrealistic framerate at input stage seemed confusing, particularly in the ffmepg context.

I think it would rather make sense to include a short description of the typical workflow then: comparing videos with the same framerate. And in the case of framerate conversions during the encoding, you have several choices for calculating VMAF depending on where and how the framerate conversion was performed. If you did it all with the `select` (or, preferably, `fps`) filter within ffmpeg, then there should be no issue reproducing the HFR → SFR conversion in one command.

We should perhaps include a description of the image-based calculation which indeed is more challenging due to the default ffmpeg input framerate. Note that here, one should actually use `-framerate` instead of `-r`. For example:

> When comparing a sequence of images with an original video source, note that ffmpeg assumes a default frame rate of 25 Hz. If you have a 60 fps source and would like to compare it with a 60 fps image sequence, use the `-framerate` option to tell the ffmpeg image demuxer the assumed framerate:
>
>     ffmpeg -framerate 60 -i "img%04d.png"  -i ref.mpg -filter_complex "[main][0:v]libvmaf" -f null -

christosbampis:
@1480c1 I actually do not have a windows machine or VM to try this out, but we can make an issue for this if you have found some specific edge case. If you know how to solve this, can you please make a pull request and we can merge that in.

resource/doc/libvmaf.md
christosbampis: Thanks for the suggestions! let me try out the -framerate option and get back to you about it.

1480c1:
The specific edge case is noted in https://github.com/Netflix/vmaf/issues/413#issuecomment-568328774

`the model_path needs to be in a "mixed mode" like libvmaf=model_path=C\\:/vmaf/share/model/vmaf_v0.6.1.pkl with the : escaped and the \ escaped to escape the : for ffmpeg to accept it as an argument to model_path instead of splitting it`

The reason why I would note it in the docs is because the default in ffmpeg (`/usr/local/share/model/vmaf_v0.6.1.pkl`) doesn't exist in the eyes of windows, so passing a `model_path` to ffmpeg is necessary unless ffmpeg allows changing the default through a configure flag for distributors. However, as windows is windows, there is no sane default paths for files like models, so having a configured default would not be helpful much.

Plus, for ffmpeg at least, the `:` is a flag separator so it will try to consume it instead of counting it as a drive, next, it would be necessary to escape the `:` to prevent ffmpeg from consuming it, but it will be necessary to escape the backslash.

So in my specific case, I would need to run
```none
ffmpeg -i .\test1.y4m -i .\testsrc.y4m -lavfi 'libvmaf=model_path=D\\:/media-autobuild_suite/local64/share/model/vmaf_v0.6.1.pkl' -f null -
```

since running without a `model_path` results in can't read file errors and running with `D:/media-autobuild_suite/local64/share/model/vmaf_v0.6.1.pkl` results in `Trouble reading the file:D`

resource/doc/libvmaf.md
```diff
@@ -54,9 +54,56 @@ Using FFmpeg with libvmaf is very powerful, as you can create complex filters to
...
+aligned. We can also the `-an` option to disregard the presence of audio.
 
-See the [FFmpeg's guide to libvmaf](https://ffmpeg.org/ffmpeg-filters.html#libvmaf), the [FFmpeg Filtering Guide](https://trac.ffmpeg.org/wiki/FilteringGuide) for more examples of complex filters, and the [Scaling Guide](https://trac.ffmpeg.org/wiki/Scaling) for information about scaling and using different scaling algorithms.
+See the [FFmpeg's guide to libvmaf](https://ffmpeg.org/ffmpeg-filters.html#libvmaf), the [FFmpeg Filtering Guide](https://trac.ffmpeg.org/wiki/FilteringGuide) for more examples of complex filters, and the [Scaling Guide](https://trac.ffmpeg.org/wiki/Scaling) for information about scaling and using different scaling algorithms. Please refer to the next section for some additional examples on more complex cases.
+
+### Use `libvmaf.a` with FFmpeg - More examples
```
kylophone: I believe we link the dynamic `libvmaf` in `vf_libvmaf`.

flip111:
I would like (as a user) to have one example where i can specify X seconds starting at offet Y of both videos.

christosbampis:
status update: given the new API changes, it is likely that some things might also change on the ffmpeg command line, so holding off this PR for later to improve and address all comments.

resource/doc/libvmaf.md
christosbampis: If I remember this correctly, ffmpeg master uses an older version/tag of vmaf master, which did not supply a dynamic library back then. But not 100%sure if it is still the case (we can check). Given the new API changes, this part of the readme (along with others) might indeed require more changes.

kylophone:
Closing for now, as this seems stale. Feel free to re-open if necessary.

li-zhi:
@christosbampis are we ready to address this one?

christosbampis:
@li-zhi since the new API is coming up, one suggestion is to integrate the PR and the comments into a fresh branch that has usage examples of the filter using ffmpeg+new API (rc). So I think we can first get an ffmpeg patch with the new API and then we can get back to this to add some examples.

PR #427, opened 2020-01-07
Add --pool to print_usage()

ltrudeau-twoorioles:
Make JSON more coherent with XML (which already outputs pool).
Contrary to the XML output, JSON pool parameter will always be
exported.


kylophone:
LGTM, also adding @li-zhi and @christosbampis for review.

PR #428, opened 2020-01-07
Link vmafossexec with static version of libvmaf

ltrudeau-twoorioles:
By default, both_libraries() returns the dynamic version.

### DIFF
--- a/libvmaf/tools/meson.build
+++ b/libvmaf/tools/meson.build
@@ -8,7 +8,7 @@ vmafossexec = executable(
     include_directories : libvmaf_inc,
     c_args : vmaf_cflags_common,
     cpp_args : vmaf_cflags_common,
-    link_with : libvmaf,
+    link_with : libvmaf.get_static_lib(),
     install : true,
 )
 


kylophone:
Is there a reason why static linking is preferred?

1480c1:
One thing I would add is that in a msys2 mingw-w64 environment with shared libraries hidden, running make right after cloning results in
```bash
[88/88] g++ @tools/vmafossexec.exe.rsp
FAILED: tools/vmafossexec.exe
g++ @tools/vmafossexec.exe.rsp
D:/media-autobuild_suite/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/media-autobuild_suite/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/9.2.0/libgcc_eh.a(unwind-seh.o):(.text+0x3d0): multiple definition of `_Unwind_Resume'; src/libvmaf.dll.a(d000018.o):(.text+0x0): first defined here
collect2.exe: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
make: *** [Makefile:3: all] Error 1
```

While this PR solves the issue

kylophone:
Maybe something like `--default-library=static` (like the `dav1d` build) would be appropriate.

1480c1:
> Maybe something like `--default-library=static` (like the `dav1d` build) would be appropriate.

Solves the linking issue at least

ltrudeau-twoorioles:
> Is there a reason why static linking is preferred?

We would like a standalone executable. If I recall, with the old build system, it was standalone.

> Maybe something like `--default-library=static` (like the `dav1d` build) would be appropriate.

I tried `--default-library=static` and `-Ddefault-library=static` and the executable still required libvmaf. From my limited understanding of meson, "_The returned `buildtarget` [of _both_libraries()_] always represents the shared library._"  https://mesonbuild.com/Reference-manual.html#both_libraries

kylophone:
Alright, merged.

PR #442, opened 2020-01-14
Add SiTiNorefFeatureExtractor and test; fix a noref test case.

li-zhi:


python/src/vmaf/core/noref_feature_extractor.py
```diff
@@ -431,3 +431,67 @@ def mscn_extract_niqe(cls, img, patch_size, mode):
...
+                scores_mtx_list.append(np.hstack(([si], [ti])))
+                i += 1
+            scores_mtx = np.vstack(scores_mtx_list)
+
+            # write scores_mtx to log file
+        log_file_path = self._get_log_file_path(asset)
+        with open(log_file_path, "wb") as log_file:
+            np.save(log_file, scores_mtx)
+
+    def _get_feature_scores(self, asset):
```
christosbampis: The _get_feature_scores() method seems to be used in a number of other places in the file, maybe we can find a way to re-use existing logic?

python/src/vmaf/core/noref_feature_extractor.py
```diff
@@ -4,30 +4,30 @@
 from scipy.special._ufuncs import gamma
 from skimage.util import view_as_windows
 from PIL import Image
+import numpy as np
+import copy
+from scipy import ndimage
 
 from vmaf.core.executor import NorefExecutorMixin
 
 __copyright__ = "Copyright 2016-2018, Netflix, Inc."
```
christosbampis: not directly related to this PR, but maybe in another PR we can replace the copyright with latest dates, e.g. 2016-2020

python/src/vmaf/core/noref_feature_extractor.py
```diff
@@ -431,3 +431,67 @@ def mscn_extract_niqe(cls, img, patch_size, mode):
             assert False
 
         return list_features
+
+
+class SiTiNorefFeatureExtractor(NorefExecutorMixin, FeatureExtractor):
```
christosbampis: maybe we can add a disclaimer that this implementation makes a number of certain assumptions (e.g. the 75 percentile) and that other SI-TI implementations are available too?

kylophone:
Should we implement this in C?

python/src/vmaf/core/noref_feature_extractor.py
li-zhi: Yes. it is generic and we can move it to the abstract class as the default (it is always overridable).

python/src/vmaf/core/noref_feature_extractor.py
li-zhi: Ok. we can do that (in a different PR).

python/src/vmaf/core/noref_feature_extractor.py
li-zhi: Not sure. For now I think we are OK and we should validate it against BT.500-14 Annex 6 at some point of time.

li-zhi:
> Should we implement this in C?

Maybe later if people want it

python/src/vmaf/core/noref_feature_extractor.py
li-zhi: Hmm. i looked into refactoring the code. The base class already have a _get_feature_scores method. Perhaps I'll keep it this way for now, and look for a more systematic way of cleaning up the code in the near future.

python/src/vmaf/core/noref_feature_extractor.py
slhck: Sorry for dropping in on this thread, but are there any test vectors for SI/TI at all? Annex 6 of BT.500 only specifies the equations.

I think your implementation is functionally equivalent to what I did [here](https://github.com/slhck/siti), so I guess it's fine. There's also an [OpenCV implementation](https://github.com/Telecommunication-Telemedia-Assessment/SITI/).

li-zhi: @slhck as you said, I don't have any test vectors. Are you aware of any? At least I plan to run through the equations to make sure the math matches.

slhck: No, I am not aware of any. Maybe we can cross-check with Margaret at NTIA if she has something from earlier VQEG FR tests. If all our three implementations matched, that would also be good. (I *did* cross-check the OpenCV code with my Python code though.)

PR #447, opened 2020-01-16
Appveyor: Add mingw-w64 builds

1480c1:
Since MSVC builds are still broken due to pthread dependencies, I've left the code to enable them, but commented them out.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,41 +1,55 @@
-os: Visual Studio 2017
+os: Visual Studio 2019
 
 environment:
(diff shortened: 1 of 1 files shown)


kylophone:
> Since MSVC builds are still broken due to pthread dependencies, I've left the code to enable them, but commented them out.

Great! I can't wait to see a green Appveyor build again. Eventually `ptools` will go away and the `pthreads` dependency will become a much smaller issue. At that point we'll write a Win32 threads wrapper and MSVC will no longer be an issue.

PR #457, opened 2020-01-19
Parse CLI to populate model names and other user-defined parameters.

christosbampis:


libvmaf/include/libvmaf/model.config.h
```diff
@@ -0,0 +1,19 @@
+#ifndef __VMAF_MODEL_CONFIG_H__
+#define __VMAF_MODEL_CONFIG_H__
+
+enum VmafModelFlags {
+    VMAF_MODEL_FLAG_DISABLE_CLIP = (1 << 0),
+    VMAF_MODEL_FLAG_ENABLE_TRANSFORM = (1 << 1),
+    VMAF_MODEL_FLAG_ENABLE_CONFIDENCE_INTERVAL = (1 << 2),
+    VMAF_MODEL_FLAG_DEFAULT = VMAF_MODEL_FLAG_DISABLE_CLIP,
```
kylophone: Call this `VMAF_MODEL_FLAGS_DEFAULT` since we may OR other flags later.

libvmaf/include/libvmaf/model.config.h
```diff
@@ -0,0 +1,19 @@
...
+};
+
+typedef struct VmafModelConfig {
+    enum VmafModelFlags flags;
+    char *name;
+    char *path;
+} VmafModelConfig;
+
+void vmaf_model_config_destroy(VmafModelConfig *config);
```
kylophone: Probably unnecessary, I'll elaborate below...

libvmaf/include/libvmaf/model.config.h
kylophone: Also, just put this in `model.h`, no need for the extra header file.

libvmaf/include/libvmaf/model.h
```diff
@@ -1,9 +1,11 @@
 #ifndef __VMAF_MODEL_H__
 #define __VMAF_MODEL_H__
 
+#include <libvmaf/model.config.h>
+
 typedef struct VmafModel VmafModel;
 
-int vmaf_model_load_from_path(VmafModel **model, const char *path);
+int vmaf_model_load_from_path(VmafModel **model, VmafModelConfig *config);
```
kylophone: `cfg` instead of `config`, not really a big problem but this is how `VmafConfiguration`s are referred to, I think.

libvmaf/src/model.c
```diff
@@ -47,9 +51,18 @@ void vmaf_model_destroy(VmafModel *model)
...
     free(model->path);
+    free(model->name);
     svm_free_and_destroy_model(&(model->svm));
     for (unsigned i = 0; i < model->n_features; i++)
         free(model->feature[i].name);
     free(model->feature);
     free(model);
 }
+
+void vmaf_model_config_destroy(VmafModelConfig *config)
```
kylophone: Probably unnecessary, I'll elaborate below...

libvmaf/src/unpickle.cpp
```diff
@@ -22,7 +25,8 @@ static int unpickle(VmafModel *model, const char *pickle_path)
 
     if (!((VAL_IS_NONE(score_clip)) || VAL_IS_LIST(score_clip)))
         return -EINVAL;
-    model->score_clip.enabled = !(VAL_IS_NONE(score_clip));
+    model->score_clip.enabled = !(VAL_IS_NONE(score_clip))
+                                && !(flags & VMAF_MODEL_FLAG_DISABLE_CLIP);
```
kylophone: put the `&&` at the end of the first line

libvmaf/test/test_model.c
```diff
@@ -8,8 +7,16 @@ static char *test_model_load_and_destroy()
     int err;
 
     VmafModel *model;
-    err = vmaf_model_load_from_path(&model, "../../model/vmaf_v0.6.1.pkl");
+    VmafModelConfig config = {
```
kylophone: If you called `vmaf_model_config_destroy()` on this it would probably segfault since these are statically allocated.

libvmaf/tools/cli_parse.c
```diff
@@ -78,6 +79,75 @@ static unsigned parse_unsigned(const char *const optarg, const int option,
     return res;
 }
 
+static int parse_config(VmafModelConfig **config,
```
kylophone: I think it's possible to write this in a way that it requires no extra memory allocation. The strings are part of `argc[]`, so no extra `memcpy` should be required. You can return a `VmafModelConfig` directly to avoid having to pass it as a pointer.

kylophone: Also, call this `parse_model_config()`.

libvmaf/include/libvmaf/model.config.h
kylophone: Also, make `VMAF_MODEL_FLAGS_DEFAULT ` a `#define`, since it doesn't make sense for it to be part of the `enum`.

christosbampis:
Thanks a lot for the comments! Addressed all the PR comments except the VmafModelConfig allocation, which we can discuss further.

PR #463, opened 2020-01-25
libvmaf: implement vmaf_use_features_from_model

kylophone:
Implements `vmaf_use_features_from_model()`. Function call used for registering all feature extractors required by a specific model.


kylophone:
@1480c1, do you know what's with the clang-10 build?

1480c1:
It seems clang-10 was removed globally or something is wrong since svt-av1 is also failing

PR #467, opened 2020-01-27
travis: use most recent clang package

kylophone:
Since `clang-10` doesn't seem to be an installable package any more, let's just use the most recent `clang` package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
       gcc-8) sudo apt-get install -qq gcc-8 g++-8 ;;
       gcc-9) sudo apt-get install -qq gcc-9 g++-9 ;;
       clang-6.0) sudo apt-get install -qq clang-6.0 ;;
-      clang-10) sudo apt-get install -qq clang-10 ;;
+      clang) sudo apt-get install -qq clang ;;
       esac
     fi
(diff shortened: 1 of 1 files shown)


1480c1:
https://github.com/OpenVisualCloud/SVT-AV1/pull/980/files#diff-354f30a63fb0907d4ad57269548329e3
It seems that you would need to change line 11 to `sourceline: "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main"` since they moved clang-10 off testing and into its own folder

kylophone:
> https://github.com/OpenVisualCloud/SVT-AV1/pull/980/files#diff-354f30a63fb0907d4ad57269548329e3
> It seems that you would need to change line 11 to `sourceline: "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main"` since they moved clang-10 off testing and into its own folder

Both of these seem to work. Do you think there's any reason to prefer one over the other?

1480c1:
clang-10 might be able to catch more bugs than clang-6, although clang-6 is the default, so it might be more common

kylophone:
OK, going to merge this as-is. If we see there are issues, I'm happy to adjust this.

PR #483, opened 2020-02-05
fix libvmaf_rc regression

kylophone:
Fixes the regression introduced when 10-bit input was implemented. PSNR, SSIM, MS-SSIM feature extractors do not use a pixel value offset.


kylophone:
@li-zhi, all tests are passing now with the exception of this one. I don't think this is a regression, this probably never passed as written.

```
>       self.assertAlmostEqual(results[0]['VMAFRC_motion2_score'], 0.7523685999999999, places=8)
E       AssertionError: 0.752368 != 0.7523685999999999 within 8 places (5.999999999062311e-07 difference)
```

PR #491, opened 2020-02-11
travis: build libvmaf with sanitizers enabled

kylophone:
Sets the Travis jobs to build all `libvmaf` libs/bins and unit tests with sanitizers enabled. With this enabled, we should catch undefined behavior, address problems, and memory leaks on CI. The tradeoff is execution speed, but penalty is < x2.


kylophone:
None of these artifacts are archived for distribution. If that was desired in the future, we could have a separate release build.

kylophone:
Looks like there are still a couple linking problems for a couple of compilers.

kylophone:
This would have helped avoid #762.

PR #500, opened 2020-02-18
libvmaf_rc: expose adm scores across scales.

christosbampis:


libvmaf/src/feature/float_adm.c
```diff
@@ -68,6 +86,8 @@ static int close(VmafFeatureExtractor *fex)
 
 static const char *provided_features[] = {
     "'VMAF_feature_adm2_score'",
+    "adm2_scale0", "adm2_scale1",
```
li-zhi: should it be called adm_scale0 instead of adm2_scale0 to be consistent with the Python implementation?

Do we need to expose it in the C code? In the Python code it was meant to be an experimental elementary metric, which was not eventually adopted in the current model.

kylophone: Is there any reason not to? We've already done the work of computing the value.

PR #519, opened 2020-03-09
libvmaf/output: add width, height, fps to xml log

kylophone:
Adds width, height, and fps stats to the XML output file.

```xml
<VMAF version="RELEASE_CANDIDATE">
  <params width="576" height="324" />
  <fyi fps="66.44" />
  <frames>
    <frame frameNum="0" motion2="0.000000" adm2="0.962171" adm_scale0="0.946342" adm_scale1="0.939103" adm_scale2="0.957538" adm_scale3="0.981014" vif_scale0="0.505393" vif_scale1="0.878155" vif_scale2="0.937589" vif_scale3="0.964358" ../model/vmaf_v0.6.1.pkl="83.868099" />
    <frame frameNum="1" motion2="4.215693" adm2="0.949842" adm_scale0="0.929862" adm_scale1="0.916670" adm_scale2="0.938517" adm_scale3="0.978180" vif_scale0="0.411091" vif_scale1="0.811943" vif_scale2="0.894037" vif_scale3="0.937140" ../model/vmaf_v0.6.1.pkl="82.661630" />
    <frame frameNum="2" motion2="4.070274" adm2="0.943535" adm_scale0="0.912681" adm_scale1="0.902340" adm_scale2="0.927663" adm_scale3="0.981115" vif_scale0="0.381398" vif_scale1="0.797620" vif_scale2="0.888710" vif_scale3="0.936691" ../model/vmaf_v0.6.1.pkl="81.053467" />
    <frame frameNum="3" motion2="3.825236" adm2="0.948045" adm_scale0="0.915596" adm_scale1="0.909747" adm_scale2="0.937121" adm_scale3="0.982282" vif_scale0="0.390273" vif_scale1="0.800916" vif_scale2="0.891099" vif_scale3="0.938725" ../model/vmaf_v0.6.1.pkl="81.935266" />
    <frame frameNum="4" motion2="3.825236" adm2="0.937779" adm_scale0="0.908689" adm_scale1="0.892088" adm_scale2="0.922616" adm_scale3="0.977720" vif_scale0="0.349535" vif_scale1="0.760889" vif_scale2="0.862231" vif_scale3="0.919330" ../model/vmaf_v0.6.1.pkl="77.519625" />
  </frames>
</VMAF>

```

### DIFF
--- a/libvmaf/src/feature/feature_collector.c
+++ b/libvmaf/src/feature/feature_collector.c
@@ -129,6 +129,9 @@ int vmaf_feature_collector_append(VmafFeatureCollector *feature_collector,
     pthread_mutex_lock(&(feature_collector->lock));
     int err = 0;
 
+    if (!feature_collector->timer.begin)
+        feature_collector->timer.begin = clock();
+
     FeatureVector *feature_vector =
         find_feature_vector(feature_collector, feature_name);
 
@@ -157,6 +160,7 @@ int vmaf_feature_collector_append(VmafFeatureCollector *feature_collector,
     err = feature_vector_append(feature_vector, picture_index, score);
 
 unlock:
+    feature_collector->timer.end = clock();
     pthread_mutex_unlock(&(feature_collector->lock));
     return err;
 }
--- a/libvmaf/src/feature/feature_collector.h
+++ b/libvmaf/src/feature/feature_collector.h
@@ -21,6 +21,7 @@
 
 #include <pthread.h>
 #include <stdbool.h>
+#include <time.h>
 
 typedef struct {
     char *name;
@@ -34,6 +35,7 @@ typedef struct {
 typedef struct VmafFeatureCollector {
     FeatureVector **feature_vector;
     unsigned cnt, capacity;
+    struct { clock_t begin, end; } timer;
     pthread_mutex_t lock;
 } VmafFeatureCollector;
 
--- a/libvmaf/src/libvmaf.rc.c
+++ b/libvmaf/src/libvmaf.rc.c
@@ -23,6 +23,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 
 #include <libvmaf/libvmaf.rc.h>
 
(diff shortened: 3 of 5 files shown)


kylophone:
@christosbampis, did you notice the attribute name changes? Are you OK with that?

li-zhi:
I think based on previous discussion, it requires a fix to give a default
name "vmaf" if there is only one model and the model name is not provided;
if there are two or more models, the model names must be provided for each
model.

kylophone:
> I think based on previous discussion, it requires a fix to give a default name "vmaf" if there is only one model and the model name is not provided; if there are two or more models, the model names must be provided for each model.

Sure, that's not what I meant, though. Rationale mostly has to do with the `scaled` prefix for width and height. Calculating VMAF on a non-scaled inputs should be a pretty normal use case for this library. `fps` for now is wall clock, whether or not to include a cpu time fps is another question. I prefer to leave this out since there are other external tools (`time`) to measure this.

```xml
<fyi ... execFps="66.44" />
<params ... ScaledWidth="576" ScaledHeight="324" />
```
becomes...

```xml
<fyi ... fps="66.44" />
<params ... width="576" height="324" />
```

christosbampis:
@kylophone Yep, the attribute change from scaledWidth to width looks ok. If we wanted another alternative, qualityWidth (meaning the width/height that quality is calculated at) could be an option, but maybe simpler is better :)
@li-zhi as @kylophone mentioned, there is another PR fixing the default model name to "vmaf" and avoid using the path of the model as the name. That PR also explains as part of the command line that a name is required if >1 models exist. We can also add an error log as part of the library, if an existing feature name exists.

li-zhi:
Thanks for the clarification.

On "scaledWidth -> width": I am leaning toward keeping the name
"scaledWidth" to differentiate it from the input video's width. Another
option is "targetWidth" (as in reference/distorted/target), but I think
scaledWidth captures what's being done the best.

christosbampis:
@li-zhi One thought about scaledWidth is that the "scaled" prefix might not be very clear, since one could be scaling to the source's resolution or the encode's resolution, or let's say has the flexibility to do so.

Another alternative (besides targetWidth or qualityWidth) is: predictionWidth or computationWidth.

kylophone:
Unless we add internal scaling into the library, `libvmaf` has no way to differentiate any of the above cases.

li-zhi:
I am thinking about the ffmpeg+libvmaf case, example. The output log needs
to be explicit about what the width represents. scaledWidth could avoid the
confusion.

Issue #406, opened 2019-12-06
Changes in how to build libvmaf.a static library


ariffinsetya:
Hi, is there any changes in how to build libvmaf.a, because I'm unable to build it following the guide in the readme..

1480c1:
```bash
git clone https://github.com/Netflix/vmaf.git
cd vmaf/libvmaf
mkdir build
cd build
meson .. --buildtype=release
ninja
```
You should be able to find `libvmaf.a` under `vmaf/libvmaf/build/src/libvmaf.a`

1480c1:
They recently changed to a meson build system, but a lot of the docs haven't been updated yet

jryzkns:
I got pretty confused just now when I was trying to build vmafossexec as well, please update the documentations on these!

ryanlmcmurray:
> ```shell
> git clone https://github.com/Netflix/vmaf.git
> cd vmaf/libvmaf
> mkdir build
> cd build
> meson .. --buildtype=release
> ninja
> ```
> 
> You should be able to find `libvmaf.a` under `vmaf/libvmaf/build/src/libvmaf.a`

Tried this, but build dir already exists:
```
$ pwd 
/home/ryan/vmaf/libvmaf 

$ ls -al build | awk '{ print $9 }'
.
..
doc
include
meson-logs
meson-private
src
test
tools
```

So I:

```
$ cd build
$ meson .. --buildtype=release
The Meson build system
Version: 0.49.2
Source dir: /home/ryan/vmaf/libvmaf
Build dir: /home/ryan/vmaf/libvmaf/build
Build type: native build
Project name: libvmaf
Project version: 1.3.16
Native C compiler: cc (gcc 7.4.0 "cc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0")
Native C++ compiler: c++ (gcc 7.4.0 "c++ (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0")
Build machine cpu family: x86_64
Build machine cpu: x86_64
Dependency threads found: YES
Configuring version.h using configuration
Program doxygen found: YES (/usr/bin/doxygen)
Configuring Doxyfile using configuration

test/meson.build:3:0: ERROR:  Include directory to be added is not an include directory object.

A full log can be found at /home/ryan/vmaf/libvmaf/build/meson-logs/meson-log.txt
```
How can I work around this?
Thank you!

1480c1:
[...]

PR #520, opened 2020-03-12
fix building error

XuShiSheng:
fix building error. "stdatomic.h" is missing in GCC-4.8. We use gnu&msc buildin fuctions instead of atomic_int in stdatomic.h. Problem mentioned in (#406)


kylophone:
I think a prefer an approach similar to what `dav1d` has done on this: https://code.videolan.org/videolan/dav1d/-/blob/master/include/compat/gcc/stdatomic.h

XuShiSheng:
ok, I changed modifications and follow what has been done in dav1d.

kylophone:
Thanks for taking the time to fix that! I will take a look at this today.

XuShiSheng:
hi, is there something awkward need to be change?

kylophone:
> hi, is there something awkward need to be change?

Nope. Sorry about the delay. Merged.

PR #534, opened 2020-03-30
Feature/frame 0 motion fix advanced lkrasula

krasuluk:
Assigning motion2[0] = motion[1] together with removing redundant calculation of difference from the previous frame and storing the previous frame.

### DIFF
--- a/libvmaf/src/feature/all.c
+++ b/libvmaf/src/feature/all.c
@@ -46,6 +46,7 @@ int all(int (*read_frame)(float *ref_data, float *main_data, float *temp_data, i
 {
     double score = 0;
     double score2 = 0;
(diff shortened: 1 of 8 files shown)


kylophone:
I believe that we should not modify the pre-existing feature `motion2` and instead store this new feature as `motion3`. Reproducibility of vmaf scores is important, and we can't guarantee that when one of the underlying features is modified.

Another thing, all these changes are all on the old libvmaf, my preference is to leave the old libvmaf alone, and make these changes on the new feature extractors only. The reason for this is that the old library will be deprecated soon.

Both these comments apply to both of the PRs.

kylophone:
Closing for now.

PR #536, opened 2020-03-31
VMAF Integer Motion Score - Submitted by Ittiam

IttiamVijayakumarGR:
Modification to floating point implementation of VMAF:
- Yuv file fopen modification  (fopen in "rb" mode)

Addition of Integer Motion score
- Added integer point implementation of motion as a new feature to Vmaf Feature Extractor (vmaf_rc api)


kylophone:
Great. Thanks for splitting this into a new PR. Dealing with each feature extractor separately should make the review process a lot less noisy.

For the coming PRs, please access the picture buffers directly. This extra allocation/copy was needed to wrap the existing the float implementations, but should not be needed for the integer feature extractors. No extra allocation and copy should be necessary.

Since motion is a feature extractor that needs buffered input, you may allocate your 3 blurred buffers with `vmaf_picture_alloc`. Instead of `integer_picture_copy`, the `VmafPicture` api should be extended with a similar function `vmaf_picture_copy`.

Is there a reason for 16-bit convolution only?

Any functions or data that you do not plan to use anywhere else, please add these directly to `integer_motion.c` and mark as `static`. Since `vmafossexec` will be deprecated soon, please do not spend time implementing these feature extractors in the old library.

After we address these items we can address style. I will provide a .`clang-format` file. Do not worry about this now.

IttiamVijayakumarGR:
Hi Kyle,

We have modified the code to incorporate the following changes.

1. Allocated a Buffer in VMAFPicture to hold Offset added values (-128 for 8 bit and -512 for 10 bit)
   offsetting is done once and will be used in Integer_motion, Integer_adm and Integer_vif functions.
2. Original buffer without offset is retained as it is since modules like PSNR and SSIM would need the data in 0-255 form.
3. Integer picture copy is removed.
4. Convolution is done in place since the offseted buffers, since same buffers would be needed by other feature modules. hence offset buffer is treated as read only within modules.
5. 3 Blur buffers are used to store  Convolution output of 3 frames and used inside Motion in ring buffer form . hence these buffers are retained motion structure.
6. Moved all functions related to integer convolution into integer_motion_functions.c file. integer_motion file is kept clean with only high level APIs (init,extarct, flush,close). 

Additionally regarding to your query on "Is there a reason for 16-bit convolution only?" 
We have experimented with 8 bit and 10 bit YUV inputs and we concluded 16 bit was sufficient to be very precise with Float convolution. 32 bit was not needed and would have been computationally more complex.

We will submit one more commit in the same PR with these changes. Please review.

kylophone:
> 1. Allocated a Buffer in VMAFPicture to hold Offset added values (-128 for 8 bit and -512 for 10 bit)
>    offsetting is done once and will be used in Integer_motion, Integer_adm and Integer_vif functions.

Should not be necessary to do this. By adding another 16-bit buffer in `VmafPicture` for these, we're allocating x3 the memory we need. Please revert the changes you made to `VmafPicture`. Also, avoid changes to `vmaf_rc`, since this library needs to work when linked elsewhere. FFmpeg for example.

> 2. Original buffer without offset is retained as it is since modules like PSNR and SSIM would need the data in 0-255 form.

There is no reason to use offset anymore, I think. This pixel offset was something that was done in the floating point feature extractors, but should not be necessary for new feature extractors. Especially just to do a Gaussian filtering and SAD computation. Like I said this results in an overhead of 3x the memory required as well as extra time spent copying and offsetting the pixel values.

> 3. Integer picture copy is removed.

I still see two versions of it in the diff.

> 4. Convolution is not done in place since the offseted buffers would be needed by other feature modules. hence offseted buffers are treated as read only buffers within feature modules.

OK. I meant first copy into your ring buffer and then convolve in place. Your way does both of these things at the same time and is probably faster.

> 5. 3 Blur buffers are used to store  Convolution output of 3 frames and used inside Motion in ring buffer form . hence these buffers are retained motion structure.

Yep, this is good. Instead of using raw buffers, you may also want to just allocate your own `VmafPicture`s here to hold the blur buffers as well as associated pixel strides, etc.

> 6. Moved all functions related to integer convolution into integer_motion_functions.c file. integer_motion file is kept clean with only high level APIs (init,extarct, flush,close).

If some piece of data or a function is only used in this feature extractor and not anywhere else, please add them to this file and mark them as static. As certain functions get reused elsewhere or become candidates for SIMD versions, we can move them up.

> 
> Additionally regarding to your query on "Is there a reason for 16-bit convolution only?"
> We have experimented with 8 bit and 10 bit YUV inputs and we concluded 16 bit was sufficient to be very precise with Float convolution. 32 bit was not needed and would have been computationally more complex.

What I meant by this was: why not write both 8/16 bit versions of these functions?

IttiamVijayakumarGR:
> > 1. Allocated a Buffer in VMAFPicture to hold Offset added values (-128 for 8 bit and -512 for 10 bit)
> >    offsetting is done once and will be used in Integer_motion, Integer_adm and Integer_vif functions.
> 
> Should not be necessary to do this. By adding another 16-bit buffer in `VmafPicture` for these, we're allocating x3 the memory we need. Please revert the changes you made to `VmafPicture`. Also, avoid changes that require modifications at the application level (`vmaf_rc`), since this library needs to work when linked elsewhere. FFmpeg for example.
>
IttiamVijay : Ok. the reason for allocating extra buffer was to reuse the offset data across features. We will revert the changes.

> > 1. Original buffer without offset is retained as it is since modules like PSNR and SSIM would need the data in 0-255 form.
> 
> There is no reason to use offset anymore, I think. This pixel offset was something that was done in the floating point feature extractors, but should not be necessary for new feature extractors. Especially for motion which is just a Gaussian filtering and SAD computation. Like I said this results in an overhead of 3x the memory required as well as extra time spent copying and offsetting the pixel values.
> 
IttiamVijay : If offsetting is not done and dynamic range of YUV is not reduced to [-128 to 127] then the deviation of integer Features (motion, vif, adm) would be much higher than float features. This may lead to deviations in overall VMAF value more than agreed values as per Agreement. All our changes are inline to keep the deviation of Integer functions as minimal as possible. 
**Please suggest how do we proceed.**. This is very crucial,

> > 1. Integer picture copy is removed.
> 
> I still see two versions of it in the diff.
> 
IttiamVijay: We will take care of this in next commit.
> > 1. Convolution is not done in place since the offseted buffers would be needed by other feature modules. hence offseted buffers are treated as read only buffers within feature modules.
> 
> OK. I meant first copy into your ring buffer and then convolve in place. Your way does both of these things at the same time and is probably faster.
>
IttiamVijay: OK. 
> > 1. 3 Blur buffers are used to store  Convolution output of 3 frames and used inside Motion in ring buffer form . hence these buffers are retained motion structure.
> 
> Yep, this is good. Instead of using raw buffers, you may also want to just allocate your own `VmafPicture`s here to hold the blur buffers as well as associated pixel strides, etc.
> 
IttiamVijay: OK we will change the raw buffers to VMAFPicture . 
> > 1. Moved all functions related to integer convolution into integer_motion_functions.c file. integer_motion file is kept clean with only high level APIs (init,extarct, flush,close).
> 
> If some piece of data or a function is only used in this feature extractor and not anywhere else, please add them to this file and mark them as static. As certain functions get reused elsewhere or become candidates for SIMD versions, we can move them up.
> 
Ittiam Vijay: Motion has very less sub functions and moving all sub functions to one file and making sttci would be fine. But VIF and ADM has lots of functions and 1 file having all sub functions would make the file very big and lots of code. 
**Please suggest if it OK to move all the sub functions to one file even if there are lots of code**.
> > Additionally regarding to your query on "Is there a reason for 16-bit convolution only?"
> > We have experimented with 8 bit and 10 bit YUV inputs and we concluded 16 bit was sufficient to be very precise with Float convolution. 32 bit was not needed and would have been computationally more complex.
> 
> What I meant by this was: why not write both 8/16 bit versions of these functions?
IttiamVijay: 8 bit convolution was not sufficient to maintain the deviation with offset data. if we have to avoid offseting then we have to re evaluate the deviation.

kylophone:
> IttiamVijayUpdate: We did some experiments to check for data overflow scenarios, seems to have been an data overflow when the offset was not added. We have corrected the data over flow and checked for output, we see that Float and integer motion scores are matching and even in VIF and ADM functions offset removal has been tested. We will go ahead with Offset buffer and offset logic removal

Cool, thanks!

IttiamVijayakumarGR:
Hi Kyle,

We are pushing a new commit with following changes
1. Removal of offset Logic, offset data population and extra buffers for Offset data
2. Integer_copy functions removal
3. 8 bit and 16 bit convolution for 8 bitdepth and 10/12 bit depth respectively.
4. Usage of VMAFPicture buffer for storing blur data instead of raw buffers.

Please review

kylophone:
We're getting close. A few more questions:

Did you try the convolution directly on the 8-bit buffers? What was the result, and why is it necessary to upsample this to 16-bit?
Can the convolution be done without the use of an additional temporary buffer?
Does your SAD computation work for bitdepths higher that 8? The accumulator is always divided by 256.0.

IttiamVijayakumarGR:
> We're getting close. A few more questions:
> 
> Did you try the convolution directly on the 8-bit buffers? What was the result, and why is it necessary to upsample this to 16-bit?

IttiamVijay: We have tried with 8 bit convolution. Motion scores were differing by **0.48%** MAX,
Where as using 16 bit convolution resulted in  **0.00075%**. Also impact of motion score deviation on final VMAF value was in very high > 1%  when 8 bit convolution was used. Hence we used 16 bit convolution. Please let us know if any further details are needed.

> Can the convolution be done without the use of an additional temporary buffer?

IttiamVijay: Since convolution is 5 tap filter (2 previous, 1 current, 2 next) based process, it cannot be done on the input buffer inplace. Also completing Y convolve for full frame followed by X convolve would help in better SIMD optimization of each modules separately. We feel using temporary buffer to store intermediate data is better. At least we would need a row level temp buffer to store Y convolve output and the perform X convole at row level (SIMD efficiency will be less in this approach). Please suggest if row level temp buffer approach  needs to be implemented.

> Does your SAD computation work for bitdepths higher that 8? The accumulator is always divided by 256.0.

IttiamVijay: -	Yes SAD convolution works for higher bit depths, 
We are passing input data as it is without normalizing data to 8 bit format (like divide by 4 for 10 bit), we are taking care of it in x convolution and normalizing the data there based on input bitdepth. The output of convolution function is in normalized form for all bitdepths. The final output / 256.0 is done to change the output of SAD calculation from Q8 (left shifted by 8) to float format.

Q8 format is used in compute_motion to preserve accuracy with floating point calculation. 
Hope this clarifies your query. please let us know if any further details are needed.

kylophone:
Closing, replaced by #560.

PR #542, opened 2020-04-07
improve libvmaf compilation doc

slhck:


slhck:
@kylophone I'd appreciate your review on these installation notes. I think this is where users would naturally look for guidance on which dependencies are needed and how to get them before building.

libvmaf/README.md
```diff
@@ -1,15 +1,60 @@
...
+
+- [Meson](https://mesonbuild.com/) (0.47 or higher)
+- [Ninja](https://ninja-build.org/)
+
+Install these dependencies under Ubuntu with:
+
+```
+sudo apt update -qq && \
+sudo apt install python3 python3-pip python3-setuptools python3-wheel ninja-build
+pip3 install --user meson
```
1480c1, line 15: Would it be better to recommend installing ninja through pip too?

slhck: I don't have any preference here. I chose the distro-packaged version since it "just worked".

kylophone: Me neither. If there's not an official "best practice", then I'm fine with either.

kylophone:
> @kylophone I'd appreciate your review on these installation notes. I think this is where users would naturally look for guidance on which dependencies are needed and how to get them before building.

Looks good. Thanks for taking the time to update this. I wonder if we should also list clang/gcc as a build dependency, since technically a C/C++ compiler is also required. I think it's OK to forego this for now since a fair bit of C++ will be removed in the future, as well as a C99 requirement defined. I think I can update just update this when the time comes.

Issue #541, opened 2020-04-07
v1.5.1 & master: ImportError with vmaf python tools


dvrogozh:
Please, clarify whether you officially support installing vmaf python module via python setuptools (https://github.com/Netflix/vmaf/blob/master/python/setup.py)? If you do, can you, please, document that in https://github.com/Netflix/vmaf/blob/master/resource/doc/VMAF_Python_library.md?

I observe "ImportError" using vmaf-1.5.1 python tools. For example:
```
# cd $(vmaf_cloned_copy)/python
# python3 setup.py build
# python3 setup.py install --user
# python3 -c 'import vmaf; print(vmaf.__license__);'
ImportError
ImportError
ImportError
ImportError
ImportError
ImportError
ImportError
BSD+Patent
# python3 --version
Python 3.6.8
```

Same issue happens if I directly follow setup instructions on https://github.com/Netflix/vmaf/blob/master/resource/doc/VMAF_Python_library.md (`export PYTHONPATH=$(pwd)/python/src`)

I believe that's regression comparing to v1.3.15 where I don't see this issue.

slhck:
> Please, clarify whether you officially support installing vmaf python module via python setuptools

This is not supported at the moment. At least it was never intended to be installed that way.

> If you do, can you, please, document that (…)

The documentation should be updated soon.

> I observe "ImportError" using vmaf-1.5.1 python tools. For example:

Yes, there have been other reports of this as well (e.g. [here](https://github.com/Netflix/vmaf/issues/530#issuecomment-605325717)).

> I believe that's regression comparing to v1.3.15 where I don't see this issue.

That's interesting. I haven't yet checked from which files/packages these errors originate.

dvrogozh:
>> I believe that's regression comparing to v1.3.15 where I don't see this issue.

> That's interesting. I haven't yet checked from which files/packages these errors originate.

Here we go. If I did not mistake:

```
$ git bisect start
$ git bisect bad v1.5.1
$ git bisect good v1.3.15

$ git bisect log
git bisect start
# bad: [35c6044d8483a3ab9528f1cc81f75738efe4f4c0] Merge pull request #509 from Netflix/feature/update-version-changelog-readme
git bisect bad 35c6044d8483a3ab9528f1cc81f75738efe4f4c0
# good: [e434247f8bf8f5ec36ba87062656583c073394fd] Feature/update doc (#361)
git bisect good e434247f8bf8f5ec36ba87062656583c073394fd
[...]

PR #543, opened 2020-04-07
remove unnecessary ImportError, addresses #541

slhck:
The error is printed when externals (e.g. psnr) are not found. At the moment it is not very helpful and should be removed.

That said, I'm not sure what happens when these externals are not found and still called by the Python code? But this should be handled when it happens, not via this `ImportError` message.

### DIFF
--- a/python/src/vmaf/config.py
+++ b/python/src/vmaf/config.py
@@ -29,8 +29,6 @@ class VmafExternalConfig(object):
                 return path
 
         except ImportError:
-            print('ImportError')
-
             pass
 
         return None


dvrogozh:
This one works for me. No "ImportError", but as we discussed in #541 error reporting will likely needs to be improved going forward.

li-zhi:
Fixed ImportError in commit:
https://github.com/Netflix/vmaf/commit/a1a9837dd209d7fc40dca4910bcded0f98df4cb7
Not the cleanest way, but should be ok for now.

PR #544, opened 2020-04-08
WIP: restructure python project and documentation

slhck:
:warning: **Do not merge yet!**

The rationale behind this PR is to make the VMAF Python code more standalone, as it currently relies on dependencies specified outside of the Python source. By moving the Python code to a more reasonable module structure, it can be installed via `pip`. This gives us a one-line installation for all the Python stuff, including the console scripts, which means the user does not need to modify `PYTHONPATH ` or `PATH` — instead, a simple `pip install` is enough.

This move however requires re-factoring several helper functions (which assume some needed files to be outside of the Python directory) and the unit tests.

What is changed:

### VMAF Module

- Move the VMAF python code into a separate module named `vmaf`
- Add required installation dependencies to `setup.py`
- Fetch version from `__init__.py` rather than a file outside the python project
- Fix helper functions so that the paths are correctly resolved

### Console Scripts

- Move the console scripts to a submodule, and make them installable via setup.py
- Removed top level console scripts (symlinks)

### Documentation

- Improve the documentation w.r.t. python code
- Update Dockerfile to match instructions and new package format

### Still TODO:

- Check installation procedure under macOS (required Homebrew libraries)
- Check if any apt dependencies are missing


slhck:
@li-zhi @kylophone I'd appreciate if you could have a look at this PR. I'm sure it's not 100% final, but I'd like to get your thoughts on that before I continue.

I do not believe that there are any breaking changes in this, and that restructuring the package will make the code more future-proof.

I think mainly we need to make sure that the required dependencies are documented properly, and that the libraries are installed correctly. The tests seem to run fine.

libtest
```diff
@@ -1,3 +0,0 @@
-#!/usr/bin/env sh
-
-PYTHONPATH=python/src python -m unittest discover -s python/test/lib -p '*_libtest.py'
```
li-zhi, line 3: Have swe moved this test somewhere? Is it still run in travis?

python/vmaf/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 __copyright__ = "Copyright 2016-2020, Netflix, Inc."
 __license__ = "BSD+Patent"
-
+__version__ = "1.5.1"
```
li-zhi, line 6: In this case, there are two places where the version is specified: here, and in the VERSION file in the root directory. We can probably remove the outer file. Or you prefer to keep both?

python/setup.py
```diff
@@ -31,25 +30,38 @@ def get_version():
...
+        "pandas>=1.0.3",
+        "scikit-learn>=0.22.2",
+        "scikit-image>=0.16.2",
         "h5py>=2.6.0",
-        "sureal>=0.1.0",
-    ]
+        "sureal>=0.4.2",
+        "dill>=0.3.1",
+    ],
+    entry_points = {
```
li-zhi, line 52: This is nice. Thank you.

li-zhi:
@slhck I pushed some minor fix and added some minor comments. It looks good to me.

python/vmaf/__init__.py
slhck: It's necessary for the version specification to at least be somewhere inside the Python repository once you try to pip-install it. So I would suggest keeping it only there, or bumping it in two places if you have an easy way to do that via a release script.

libtest
slhck: File restored, thanks for catching this.

slhck:
I squashed your commit on top of mine (to make the history cleaner) and added the missing libtest file. I'm waiting for the CI to tell me if everything runs fine. There was a weird error with the GCC 9 build before.

li-zhi:
I just found from @kylophone that libtest is no longer need. I'll simply remove it.

li-zhi:
@slhck oops I merged. I thought it was ready to go. Wan't it?

slhck:
I think it's fine! I just wanted to double check the needed apt and Homebrew libraries. But I can still do that now. Thanks!

PR #545, opened 2020-04-09
Feature/refactor individual executables

li-zhi:


kylophone:
We should also remove the C executables (and related code) that's no longer being used.

li-zhi:
[WIP]

PR #551, opened 2020-04-17
VMAF Integer ADM : Submitted by Ittiam

IttiamVijayakumarGR:
- YUV File open modified to "rb" mode
- Integer ADM feature and related functions added .
- Offset "-128" logic removal has been accommodated in adm_dwt2 functions with a constant subtraction of add_bef_shift_round_VP


libvmaf/src/feature/float_adm.c
```diff
@@ -28,6 +28,12 @@
 #include "mem.h"
 #include "picture_copy.h"
 
+#define PROFILE_FLOAT_ADM 1
+
+#if PROFILE_FLOAT_ADM
```
1480c1, line 33: You might want to add `&& ! defined(_WIN32)` to this since `sys/time.h` doesn't exist on windows. Although `timeval` is declared in [`winsock2.h`](https://docs.microsoft.com/en-us/windows/win32/api/winsock/ns-winsock-timeval), [`gettimeofday`](https://stackoverflow.com/questions/10905892/equivalent-of-gettimeday-for-windows) is not available on windows

libvmaf/src/feature/integer_adm.c
```diff
@@ -0,0 +1,211 @@
...
+#include "mem.h"
+#include "adm_options.h"
+
+#ifndef NUM_BUFS_ADM
+#define NUM_BUFS_ADM 30
+#endif
+
+#define PROFILE_INTEGER_ADM 1
+
+#if PROFILE_INTEGER_ADM
```
1480c1, line 36: Same here

libvmaf/src/feature/float_adm.c
IttiamVijayakumarGR: Thanks for review, we will update the code and push a new commit

libvmaf/src/feature/integer_adm.c
IttiamVijayakumarGR: Thanks for review, we will update the code and push a new commit

libvmaf/src/feature/float_adm.c
```diff
@@ -28,6 +28,12 @@
 #include "mem.h"
 #include "picture_copy.h"
 
+#define PROFILE_FLOAT_ADM 1 && (defined(__MINGW32__) || (!defined(_WIN32)))
```
1480c1, line 31: You put the preprocessor check in the wrong spot

1480c1, line 31:
```suggestion
#if defined(__MINGW32__) || !defined(_WIN32)
#define PROFILE_FLOAT_ADM 1
#endif
```

IttiamVijayakumarGR: Sure, we will modify the code and push. Thanks.

1480c1: Ah I see why you had this here, although it works, it's still a bit ugly appending the conditionals at the end of a constant like this since if it was like this, you could have just done
```c++
#define PROFILE_FLOAT_ADM (defined(__MINGW32__) || !defined(_WIN32))
```
without the 1 constant while making sure this is obvious as just checking the conditionals

kylophone:
Rebased and merged in PR #625

PR #554, opened 2020-04-18
improve installation instructions

slhck:
- Fix usage of executables: These can now be called without a `./` prefix.
- Fix installation of libvmaf


slhck:
@li-zhi As these are using the YUV test resources which have just been moved to another repo, you may want to modify this README to explain how to obtain these test files.

PR #555, opened 2020-04-18
add main function to crossval script

slhck:
This is required to use it as a command line script.


li-zhi:
@slhck this script was not meant to be used as a command line script. Actually it is quite stale, and I think it can be saftely removed. Let me go ahead and do this.

PR #558, opened 2020-04-27
Actions: Add libvmaf.yml

1480c1:
Replicates some of the jobs in travis-ci

### DIFF
--- a/.github/workflows/libvmaf.yml
+++ b/.github/workflows/libvmaf.yml
@@ -1,0 +1,77 @@
+name: libvmaf
+on:
+  push:
+  pull_request:
+
+env:
+  DEBIAN_FRONTEND: noninteractive
+
+jobs:
+  libvmaf-build:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-18.04
+            CC: ccache gcc
+            CXX: ccache g++
+          - os: ubuntu-18.04
+            CC: ccache gcc-9
+            CXX: ccache g++-9
+          - os: ubuntu-18.04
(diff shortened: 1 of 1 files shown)


1480c1:
How it looks on my repository https://github.com/1480c1/vmaf/actions/runs/88645912

li-zhi:
It looks like it ran the extratests (meant for tests with external dependency such as FFmpeg and Matlab). I added conditional skip for them. Please try again after rebasing master.

1480c1:
Should I change this to always run on push and prs? Right now it's only running if there was a change to any of the workflows or under the libvmaf subfolder

li-zhi:
I am not familiar with the github workflow. How does it do differently than the travis CI? why do we need it since we already have travis CI?

1480c1:
I am mainly thinking about moving some of the jobs from travis-ci to github ci, with the main purpose of spreading the number of jobs that are currently being run. Plus, I would consider github actions to have slightly better integration with pull requests since it will display a notification for each job separately rather than a single `continuous-integration/travis-ci/pr`


Also, I have a feeling that the reason why the python job is failing is that it can't find ffmpeg

1480c1:
For example, from a different repo
image

li-zhi:
In this case, we can migrate to github ci and retire travis ci.

I pushed some changes to master. Hopefully it is going to fix the issue.

li-zhi:
@1480c1 would you also help add a "build" badge to the README.md for github ci?

1480c1:
> @1480c1 would you also help add a "build" badge to the README.md for github ci?

Sure, I can do that

PR #559, opened 2020-04-27
Actions: Add an ffmpeg build

1480c1:
Currently has macOS disabled because of errors with assembly with and without libvmaf and x86-asm

https://github.com/1480c1/vmaf/runs/623473257?check_suite_focus=true

Fixed macOS build thanks to @ePirat

https://github.com/1480c1/vmaf/actions/runs/89530778

### DIFF
--- a/.github/workflows/ffmpeg.yml
+++ b/.github/workflows/ffmpeg.yml
@@ -1,0 +1,71 @@
+name: ffmpeg
+on:
+  push:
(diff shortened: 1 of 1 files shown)


kylophone:
@1480c1, is this ready to go? I can merge if so.

1480c1:
Yeah, it's ready

PR #568, opened 2020-05-08
Mutex unlock before destroy

titardrew:
Mutex should be unlocked before destruction. Otherwise, this may lead to undefined behaviour (see https://pubs.opengroup.org/onlinepubs/009695399/functions/pthread_mutex_destroy.html)


kylophone:
Thank you. Merged.

PR #572, opened 2020-05-12
Update meson.build to next release version (1.5.2) to sync libvmaf.pc

Oxalin:
libvmaf.pc version wasn't properly sync when we moved from 1.3.15 to 1.5.1.
Increase meson.project_version() to resync release and .pc file.

Signed-off-by: Alexandre Demers <alexandre.f.demers@gmail.com>


kylophone:
Will merge this when the new version is cut. I'm wondering if we should just cut it now, @li-zhi?

PR #573, opened 2020-05-12
Add meson options to disable tests and documentation

Oxalin:
Since tests and documentation are not always required (multilib | 32 bit on
64 bit), allow disabling building and running tests or generating documentation.

Signed-off-by: Alexandre Demers <alexandre.f.demers@gmail.com>

### DIFF
--- a/libvmaf/meson_options.txt
+++ b/libvmaf/meson_options.txt
@@ -1,0 +1,9 @@
+option('tests', 
+    type: 'boolean', 
+    value: true, 
+    description: 'Enable libvmaf tests')
+
+option('documentation', 
+    type: 'boolean', 
+    value: true, 
+    description: 'Build documentation')
--- a/libvmaf/meson_options.txt
+++ b/libvmaf/meson_options.txt
@@ -7,3 +7,4 @@ option('documentation',
     type: 'boolean', 
     value: true, 
     description: 'Build documentation')
+    
--- a/libvmaf/doc/meson.build
+++ b/libvmaf/doc/meson.build
@@ -1,3 +1,8 @@
+# Leave subdir if documentation is disabled
+if not get_option('documentation')
+    subdir_done()
+endif
+
 doxygen = find_program('doxygen', required: false)
 
 if doxygen.found()
--- a/libvmaf/test/meson.build
+++ b/libvmaf/test/meson.build
@@ -1,3 +1,8 @@
+# Leave subdir if tests are disabled
+if not get_option('tests')
+    subdir_done()
+endif
+
 test_inc = include_directories('.')
 
 test_picture = executable('test_picture',


libvmaf/meson_options.txt
```diff
@@ -0,0 +1,9 @@
+option('tests', 
+    type: 'boolean', 
+    value: true, 
+    description: 'Enable libvmaf tests')
+
+option('documentation', 
+    type: 'boolean', 
+    value: true, 
+    description: 'Build documentation')
```
1480c1, line 9: Can you add a newline at the end of this file?

Oxalin: Done.

1480c1: Thanks

kylophone:
Can you explain why this is necessary? It seems to me that these can already be built independent of one another:

`ninja -vC build`
`ninja -vC build test`
`ninja -vC build doc`

1480c1:
It would be helpful to be able to completely skip a whole meson directory parsing, especially in the case that the person building is sure they have absolutely zero interest in building those components. Although it might be only a few seconds, in the case they are building multiple other projects, each second building might add up to a lot.

Oxalin:
This adds the foundation to any desired meson options. @1480c1 we could do that: add an option for any or all components that you have in mind. What would that be?

@kylophone : the point is to not build them at all. You just skip them entirely according to whatever you need.

> ninja -vC build
This will build everything that was configured, which includes tests (and run them) and doc.
The two other propositions will build specifically everything and run tests or only focus on the documentation.

Moreover, you are targeting the build level, while this series acts at the configuration level. In other words, right now, you configure everything without discernment, then you either build everything (19 build targets) or you build test or doc, but you can't build everything without test (-6 build targets) and doc (-1 build target). Concretely, I'm maintaining some multilib packages. I'm not interested in running tests, which are either already covered by the native package or skipped. Nor am I interested in the doc, again already provided by the native package.

1480c1:
The options I was thinking about are already covered in this PR, tests and docs

1480c1:
Also, question, would it be better to do the check at the adding subdirectory step, or inside the subdirectory's meson.build? Would there be any problems with just not running the `subdir` command in the first place?

Oxalin:
I looked at some other projects and both locations are used. This could be moved to whichever location the maintainers prefer.

kylophone:
@Oxalin, what is the `meson` command line used to disable the configuration of these subdirectories?

kylophone:
> @Oxalin, what is the `meson` command line used to disable the configuration of these subdirectories?

@1480c1, do you know?

1480c1:
> > @Oxalin, what is the `meson` command line used to disable the configuration of these subdirectories?
> 
> @1480c1, do you know?

`meson setup build libvmaf -Ddocumentation=false -Dtests=false` I believe

Oxalin:
Using "meson configure" will give you the available options. Then using "-D theOptionVariable", you are able to set them as answered by @1480c1

kylophone:
Cleaned up and pushed as 7134da68a7d8a3f8cf68c3df7d4ad77d1708a36e

Note the option names have changed slightly. Thanks for this.

PR #575, opened 2020-05-18
Use commit hash for versioning.

christosbampis:


libvmaf/src/libvmaf.rc.c
```diff
@@ -360,7 +361,7 @@ int vmaf_score_pooled(VmafContext *vmaf, VmafModel *model,
 
 const char *vmaf_version(void)
 {
-    return "RELEASE_CANDIDATE";
+    return VMAF_VERSION;
```
li-zhi, line 364: How is this versioning different from a major.minor.fix convention?

1480c1: This prints out just an abbreviated hash, and nothing similar to semver

```none
./libvmaf/build/tools/vmaf_rc --version
311ea4c
```

christosbampis: Hi, yes this corresponds to the abbreviated commit hash rather than the semantic version, which allows us to point to the exact commit id used.

PR #577, opened 2020-05-27
Add adm dwt2 cy

li-zhi:


kylophone:
Looks like you're only using `adm_dwt2_d` in Python via CDEF, not in the lib. IMO, it would be better to compile the library as usual and link to Python via CFFI. CDEF is for extending or optimizing Python code with C, whereas CFFI is for linking libraries.

This version of ADM will be replaced soon, so do we need to add this functionality to  #551?

PR #580, opened 2020-05-28
Add some documentation to compute_vmaf in libvmaf.h

TheVaffel:
I struggled to make VMAF's C API work, so I have written down a few observations in comments to be parsed by Doxygen in `libvmaf.h`.

The comment I have written describes usage of `compute_vmaf`. It does not cover every parameter, but most importantly covers `read_frame`, whose exact requirements I found to be hopelessly ambiguous from the function signature alone.

This documentation could definitely use more work, but I believe having a basic description of the function and a few of its parameters will be of great value to new users.

### DIFF
--- a/libvmaf/include/libvmaf/libvmaf.h
+++ b/libvmaf/include/libvmaf/libvmaf.h
@@ -23,6 +23,39 @@
 extern "C" {
 #endif
 
+ /**
+  *
+  *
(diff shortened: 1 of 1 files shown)


li-zhi:
@TheVaffel thanks for doing this. We also have a rc API which hopefully is a much better version than this one:
https://github.com/Netflix/vmaf/blob/master/libvmaf/include/libvmaf/libvmaf.rc.h
@kylophone

PR #588, opened 2020-06-16
libvmaf: VmafDictionary, VmafOption, parameterized feature extraction

kylophone:
Simple key/value dictionary. This dictionary will be used to implement filter initialization options. See `test/test_dict.c` for api usage examples.

### DIFF
--- a/libvmaf/src/dict.c
+++ b/libvmaf/src/dict.c
@@ -1,0 +1,107 @@
+#include <errno.h>
+#include <string.h>
+
+#include "dict.h"
+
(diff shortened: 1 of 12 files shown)


kylophone:
An update, since there are a few new things happening here. I will squash this into 3 commits before merging.

- `VmafDictionary`: simple key/value dictionary
- `VmafOption`: Options API for arbitrary types. Includes nice helpers like default, min, max, etc.
- internal libvmaf_rc API gets support for feature extractor initialization with options.

Issue #585, opened 2020-06-05
error: command 'x86_64-linux-gnu-gcc' failed with exit status 1


yuanCabaal:
Hi,
when I was trying to run the command "pip3 install --user ."  get the error in the title for: Building wheel for vmaf (setup.py) ... error
Running setup.py install for vmaf ... error

I followed closely the instruction of vmaf. Are there other packages that I have to install besides the ones mentioned in the documentation?

Thanks

yuanCabaal:
python-wheel is already installed. Are you suggesting to use a different package instead of python-wheel?

ChengHan111:
I also got the same question about that. I wandering if there is a more specific answer to that? Thank you

1480c1:
my best answer would be to try to see if your system differentiates between pip, pip2, and pip3, and make sure you have the right package for the pip you are using, for example, if `pip --version` returns something like `(python 3.*)` at the end, that signals to me that your distro's default python version is 3, so you would just look for `python-wheel` in your package manager, however, if your system does not default to python 3 with its packages, you would need to look for something like `python3-wheel`.

I personally do not use the python side of vmaf and pretty much only use the libvmaf and vmafossexec provided, which only requires running make or using meson+ninja directly.

@li-zhi might have a better answer for your problem since he seems to be the main person modifying the python scripts

ChengHan111:
I actually thinking I am facing this problem then I double check to make sure I have got python3-wheel already. Also I am in python 3.6.9 dev as well. Though my pip --version is in python2.7 I use pip3 all the time. Still facing the problem though

1480c1:
Can you give me details such as what distro+version and what exact commands you used starting from git clone or download so I can try to reproduce?

ChengHan111:
Thank you for helping me.
Here place the code:

git clone https://github.com/Netflix/vmaf.git
cd vmaf
cd python

sudo apt-get update -qq && \
sudo apt-get install -y \
  python3 \
  python3-dev \
  python3-pip \
  python3-setuptools \
  python3-wheel \
  python3-tk

pip3 install --user .

I get stuck at --user . The error code is above the same as @yuanCabaal mentioned four days ago.
Running in Ubuntu 18.04.4 LTS with python 3.6-dev (in python3.6.9)

1480c1:
```bash
apt update
apt install git g{cc,++} python3{,-{dev,pip,setuptools,wheel,tk}}
git clone https://github.com/Netflix/vmaf.git
[...]

PR #589, opened 2020-06-16
force-remove file that possibly does not exist

slhck:
This file does not actually exist during build, so it fails when running "rm".
Running "rm -f" is a workaround in case it might exist.


slhck:
Please disregard; this is a symptom of https://github.com/Netflix/vmaf/issues/585

PR #592, opened 2020-06-22
libvmaf feature options: public api, cli

kylophone:
Updates the `vmaf_use_feature()` api call to take an options dict, which allows passing of feature options from the public api. This is a breaking change to the `libvmaf_rc` api, but this is still an unpublished release candidate api, so not an issue. 

If you'd like to specify feature options from the `vmaf_rc` command line, usage is as follows. key/value options are set with a `=` and each individual option is delimited with a `:`. This syntax should be familiar if you are used to `libavfilter` filter strings. 

```bash
# register the "psnr" feature extractor
# set values for three separate options (enable_chroma, second_param, third_param)
./vmaf_rc --feature psnr=enable_chroma=false:second_param=123:third_param=1.414 ...`
```

Specifically setting options is not a requirement and thus can be omitted if not required. If a feature extractor has an available option and it is not set, the default value will be used.

```bash
# do not set any specific options, default values will be used.
./vmaf_rc --feature psnr
```

Before this is merged, I still need to verify multithreaded feature extractor options.

### DIFF
--- a/libvmaf/include/libvmaf/feature.h
+++ b/libvmaf/include/libvmaf/feature.h
@@ -1,0 +1,26 @@
+/**
+ *
+ *  Copyright 2016-2020 Netflix, Inc.
+ *
+ *     Licensed under the BSD+Patent License (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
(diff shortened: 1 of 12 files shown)


kylophone:
`strsep()` is not available on Windows, looks like. Will see about a shim for that.

PR #593, opened 2020-06-23
vmaf_rc: provide a strsep compatibility shim

kylophone:
While `strsep()` is not as portable as `strtok()`, it is a much better alternative. This PR provides a compatibility shim, and should fix the Windows build.


1480c1:
@kylophone would it be possible to mark the compatibility `strsep` as static to make it so that any it won't conflict with other libs who might also want to provide a compatibility `strsep` in the future?

kylophone:
> @kylophone would it be possible to mark the compatibility `strsep` as static to make it so that any it won't conflict with other libs who might also want to provide a compatibility `strsep` in the future?

Oops, thanks. Fixed.

PR #594, opened 2020-06-23
libvmaf/motion: support motion in addition to motion2

kylophone:
Some published VMAF models require a feature called `motion` which is similar, but has no look-ahead like `motion2`. This PR adds the `motion` feature to both the float/integer motion feature extractors.

```xml
<VMAF version="85b492c">
  <params qualityWidth="1920" qualityHeight="1080" />
  <fyi fps="60.80" />
  <frames>
    <frame frameNum="0" motion="0.000000" motion2="0.000000" integer_motion="0.000000" integer_motion2="0.000000" />
    <frame frameNum="1" motion="4.913140" motion2="4.828127" integer_motion="4.913157" integer_motion2="4.828149" />
    <frame frameNum="2" motion="4.828127" motion2="4.678761" integer_motion="4.828149" integer_motion2="4.678786" />
    <frame frameNum="3" motion="4.678761" motion2="4.678761" integer_motion="4.678786" integer_motion2="4.678786" />
  </frames>
</VMAF>

```

PR #595, opened 2020-06-23
Refactor motion to motion2

li-zhi:


libvmaf/src/feature/feature_extractor.c
```diff
@@ -30,9 +30,9 @@ extern VmafFeatureExtractor vmaf_fex_float_psnr;
 extern VmafFeatureExtractor vmaf_fex_float_ansnr;
 extern VmafFeatureExtractor vmaf_fex_float_adm;
 extern VmafFeatureExtractor vmaf_fex_float_vif;
-extern VmafFeatureExtractor vmaf_fex_integer_motion;
+extern VmafFeatureExtractor vmaf_fex_integer_motion2;
```
kylophone, line 33: if #594 is merged, I don't think it makes sense to rename these `*_motion2`, since they provide both. We could also use the same feature extractor for `motion3` in the future.

kylophone:
See my inline comment. Also, squash before merging.

PR #599, opened 2020-06-24
Fix test include directories in older meson versions

### LINKED ISSUE #518: Can't install vmaf
I'm trying to install vmaf on LMDE4 (debian 10 based) but meson fails and I need some hints to fix the error message. meson-log.txt attached

[meson-log.txt](https://github.com/Netflix/vmaf/files/4307584/meson-log.txt)

(can the verbosity of meson-log.txt be increased? )

PeacefullyDisturbed:
This should fix #518 which affects building with meson versions before 0.50.0.


kylophone:
Thank you. Merged.

PR #608, opened 2020-07-01
Add INSTALL_PREFIX to top level Makefile

KevinJW:
The top level Makefile did not provide a way to override the meson default installation path prefix. This allows the build process to override the default by passing in a variable on the make execution if required.

Signed-off-by: Kevin Wheatley <kevin.j.wheatley@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
+INSTALL_PREFIX?=/usr/local
+
 all:
 	cd third_party/libsvm && make lib
 
-	meson setup libvmaf/build libvmaf --buildtype release && \
+	meson setup libvmaf/build libvmaf --buildtype release --prefix=${INSTALL_PREFIX} && \
 	ninja -vC libvmaf/build
 	cd python && python3 setup.py build_ext --build-lib .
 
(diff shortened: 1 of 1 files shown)


1480c1:
Would there be a way to make it so that `--prefix=...` is not there at all if `INSTALL_PREFIX` is not defined to accommodate cases in which `/usr/local` is not the user's or meson's default non-usr prefix?

Similar to doing 
```bash
meson setup libvmaf/build libvmaf --buildtype release ${INSTALL_PREFIX:+--prefix="$INSTALL_PREFIX"}
```
in the shell

KevinJW:
Sure I guess rather than specify the path, the variable could expand to the full extra parameters to pass to meson "--prefix blah", or if there is a clean way to ask meson for the default install path, the variable could initialise to that.

I was basing what I did on the comments in  https://mesonbuild.com/Builtin-options.html where is says:
`prefix defaults to C:/ on Windows, and /usr/local otherwise. You should always override this value.`

Especially that last part.

Kevin

1480c1:
If it's too hard/impossible, it's fine to ignore that suggestion since it seems `/usr/local` would be the default for meson and I don't think it might change anytime soon.

KevinJW:
It is possible if you want to require a specific flavour of make e.g. GNU make has an if directive.

I haven't figure out a macro expansion that can do it in a portable manner, I'm not an expert in either make or meson though so happy if somebody can show me!

Kevin

PR #615, opened 2020-07-09
Add rev_target to libvmaf_rc for proper building order

### LINKED ISSUE #614: Build fails if run in parallel
This may be a regression introduced in 1.5.2, as previous versions don't seem to have the issue; see https://hydra.nixos.org/job/nixpkgs/trunk/libvmaf.x86_64-linux

When run on highly multicore system, the build typically fails due to `vcs_version.h` not being present. Since this file is generated as part of the build process, I'd assume it's an undeclared dependency and that, on smaller systems, it's typically generated before any file depending on it is processed.

I've confirmed this both by disabling parallel builds, and by disabling most CPU cores. The former made the build work every time; the latter made it work _most_ of the time.

The commit that broke the build is probably https://github.com/Netflix/vmaf/commit/9a27f448002d12c5c5f4a405aca613b52fdd91ce#diff-bd4a20f180c662df62c5a52f21a65ac7. Reordering subdir processing only ensures the include-file jobs are started before the src jobs, not that they finish before the src jobs are started.

HolyWu:
Close #614

### DIFF
--- a/libvmaf/src/meson.build
+++ b/libvmaf/src/meson.build
@@ -199,7 +199,7 @@ libvmaf_rc_sources = [
 
 libvmaf_rc = both_libraries(
     'vmaf_rc',
-    libvmaf_rc_sources,
+    [libvmaf_rc_sources, rev_target],
     include_directories : [vmaf_include, libvmaf_inc],
     c_args : vmaf_cflags_common,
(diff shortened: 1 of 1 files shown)


kylophone:
Thanks, merged.

PR #619, opened 2020-07-10
[WIP] libvmaf_rc: add a `compute_vmaf()` wrapper

kylophone:
This PR provides a `compute_vmaf()` api into `libvmaf_rc`. This function is identical to the `compute_vmaf()` provided in the old library, but this implementation is a simple wrapper for the newly introduced `libvmaf_rc` APIs. 

This API has been marked as deprecated, and is introduced here as a courtesy for application developers. It will be removed in a future version, once most users have migrated.

Also in this PR, another executable has been added to the build: `vmafossexec_rc`. This is the same `vmafossexec` application code, but it links the new `compute_vmaf()` function provided by `libvmaf_rc` instead of `libvmaf`.

This PR needs more testing before merging. Testing strategy will be to run our old `vmafossexec` regression tests on `vmafossexec_rc`


li-zhi:
@kylophone I would suggest simply replace old vmafossexec with the new one, instead of creating vmafossexec_rc. If everything works fine, all the tests should pass (the test coverage should be fairly comprehensive). Plus, remove the old code including combo.c.

kylophone:
> @kylophone I would suggest simply replace old vmafossexec with the new one, instead of creating vmafossexec_rc. If everything works fine, all the tests should pass (the test coverage should be fairly comprehensive). Plus, remove the old code including combo.c.

This is just a temporary binary for regression testing. Not part fo the published install, and will go away.

kylophone:
I've marked this as WIP

kylophone:
Alright, we now have several failing tests with vmafossexec linked with the new lib. Will start working on those now.

Issue #570, opened 2020-05-12
Atomics and C++


titardrew:
Hi,
One particularly annoying thing with using libvmaf from C++ code is atomic. Suppose, I'd like to implement my own reader. It has to fill `VmafPicture`'s payload with picture data. However, to do that one has to include `picture.h` header file which includes `stdatomic.h` and uses `atomic_int` in the definition of `VmafPicture` struct. In C++ this leads to compilation errors. See [this](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=60932) and [this](https://stackoverflow.com/questions/45342776/how-to-include-c11-headers-when-compiling-c-with-gcc) for example. There's also a [proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0943r1.html) to make C++ compatible with C atomics, but this is not the case now, to the best of my knowledge.
Currently, I found 2 workarounds:
1. Manually force libvmaf to compile with its `stdatomic.h` from `compat` and add 
```c
typedef int atomic_int
typedef unsigned int atomic_uint
```
before include.

2. Add:
```c
typedef int atomic_int
typedef unsigned int atomic_uint
```
or
```C
typedef std::atomic<int> atomic_int
typedef std::atomic<unsigned int> atomic_uint
```
and include `stdatomic.h` from `compat` to CMake to avoid including standard `stdatomic.h` from `picture.h`. This works, because compat stdatomic.h does nothing if included from c++ (obviously with extern "C" as well): https://github.com/Netflix/vmaf/blob/master/libvmaf/include/compat/gcc/stdatomic.h
However, I don't claim that it works properly. It just compiles well. By the way, what's the purpose of `#if !defined(__cplusplus)`? It does not seem to solve the issue, because we also have to define atomc_int in C++ code somehow.

---

The problem is that I can't come up with a proper way to include `VmafPicture` struct without a headache with atomics.
The same issue occurs if I include feature extractors to be able to get extracted features in addition to vmaf_score. Although, this seems to be illegal currently because feature extractors are not visible from header files from `include`.

kylophone:
> ```c
> typedef int atomic_int
> typedef unsigned int atomic_uint
> ```

The above won't work, since the type needs to actually be atomic for thread safety.

Another technique might be to hide the atomic inside an anonymous struct `VmafRef`. If we did that `<stdatomic.h>` should not be required from any libvmaf API headers. What do you think about that?

titardrew:
@kylophone,
[...]

PR #623, opened 2020-07-13
 libvmaf: implement VmafRef, remove stdatomic dependency from public API

kylophone:
This PR removes the `stdatomic.h` dependency from public API. This should be helpful when libvmaf is linked from C++, and should address #570. VmafPicture reference counting is now handled inside an anonymous type called `VmafRef`. See unit test for usage example.


1480c1:
```c++
../src/feature/feature_extractor.h:121:9: error: unknown type name ‘atomic_int’
         atomic_int capacity, in_use;
         ^~~~~~~~~~
```

kylophone:
@titardrew, does this address #570?

titardrew:
@kylophone, sure, that seems to solve the issue.

Issue #635, opened 2020-08-03
Travis: libvmaf clang job is failing during setup


kylophone:
This job is not being run because it is failing during the package installation phase during setup.

Example: https://travis-ci.org/github/Netflix/vmaf/jobs/714600716

```sh
$ sudo chown -R travis: $HOME/.ccache
install.1
0.01s$ ccache -s
install.2
4.56s$ pip3 install meson
0.81s$ if type apt-get; then
    case "$CC" in
    gcc-7) sudo apt-get install -qq gcc-7 g++-7 ;;
    gcc-8) sudo apt-get install -qq gcc-8 g++-8 ;;
    gcc-9) sudo apt-get install -qq gcc-9 g++-9 ;;
    clang-6.0) sudo apt-get install -qq clang-6.0 ;;
    clang) sudo apt-get install -qq clang ;;
    esac
  fi
  
apt-get is /usr/bin/apt-get
E: Unable to correct problems, you have held broken packages.
The command "if type apt-get; then
    case "$CC" in
    gcc-7) sudo apt-get install -qq gcc-7 g++-7 ;;
    gcc-8) sudo apt-get install -qq gcc-8 g++-8 ;;
    gcc-9) sudo apt-get install -qq gcc-9 g++-9 ;;
    clang-6.0) sudo apt-get install -qq clang-6.0 ;;
    clang) sudo apt-get install -qq clang ;;
    esac
  fi
  " failed and exited with 100 during .
Your build has been stopped.
```

kylophone:
Fixed by #637.

PR #637, opened 2020-08-04
Travis: fix broken clang build

kylophone:
Attempted fix for #635.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ install:
   - ccache -s
   - pip3 install meson
   - |
+    sudo apt-get update
     if type apt-get; then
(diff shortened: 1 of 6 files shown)


kylophone:
@1480c1, any clue? (You may know better than me) `apt` is saying that `clang` is broken package.

1480c1:
Could you try adding a
```yaml
compiler: clang
```
under the clang build?

1480c1:
> clang : Depends: clang-11 (>= 11~) but it is not installable

seems this is the reason. Clang 11 is not yet out yet to the main repos either way.

Submitted fix in #640

Issue #584, opened 2020-06-05
when does vmaf can support an arm-sverser which don't have avx instrics


ROYduan:
[1/18] cc -Isrc/25a6634@@convolution_and_psnr_avx@sta -Isrc -I../src -I../src/feature/common -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=c99 -O3 -fPIC -mavx -pedantic -DOC_NEW_STYLE_INCLUDES -MD -MQ 'src/25a6634@@convolution_and_psnr_avx@sta/feature_common_convolution_avx.c.o' -MF 'src/25a6634@@convolution_and_psnr_avx@sta/feature_common_convolution_avx.c.o.d' -o 'src/25a6634@@convolution_and_psnr_avx@sta/feature_common_convolution_avx.c.o' -c ../src/feature/common/convolution_avx.c
FAILED: src/25a6634@@convolution_and_psnr_avx@sta/feature_common_convolution_avx.c.o 
cc -Isrc/25a6634@@convolution_and_psnr_avx@sta -Isrc -I../src -I../src/feature/common -fdiagnostics-color=always -pipe -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=c99 -O3 -fPIC -mavx -pedantic -DOC_NEW_STYLE_INCLUDES -MD -MQ 'src/25a6634@@convolution_and_psnr_avx@sta/feature_common_convolution_avx.c.o' -MF 'src/25a6634@@convolution_and_psnr_avx@sta/feature_common_convolution_avx.c.o.d' -o 'src/25a6634@@convolution_and_psnr_avx@sta/feature_common_convolution_avx.c.o' -c ../src/feature/common/convolution_avx.c
cc: error: unrecognized command line option ‘-mavx’
[...]

PR #638, opened 2020-08-05
libvmaf: Add cross-platform CPU detection

kylophone:
This PR adds cross-platform CPU detection at both compile time as well as runtime. Should compile just fine on non-x86 platforms, and for x86 platforms adds an API for querying and masking CPU capabilities. This is easily extended to other platforms like ARM if we want to add some NEON in the future. This implementation borrows heavily from dav1d.

It should be noted that this adds an additional build dependency for x86, `nasm`. 

For example usage, see the following [test](https://github.com/Netflix/vmaf/compare/vmaf_init_cpu?expand=1#diff-3447e17508e4736b1d8768e040bfdb1eR32-R42). Capability querying and masking API is [here](https://github.com/Netflix/vmaf/pull/638/commits/50fb101eee495dff0d418c92ee61b6b8d12697aa#diff-fdc84b2396c69d94b2f41c05b90997c7).

Next steps: 
- Fix the old CPU detection and intrinsics code so we can compile on non-x86 again. (#584)
- Start bringing the x86 feature extractor optimizations.

### DIFF
--- a/libvmaf/src/x86/cpu.c
+++ b/libvmaf/src/x86/cpu.c
@@ -1,0 +1,80 @@
+/*
+ * Copyright © 2018, VideoLAN and dav1d authors
+ * Copyright © 2018, Two Orioles, LLC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright notice, this
+ *    list of conditions and the following disclaimer.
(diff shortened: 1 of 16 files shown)


libvmaf/src/x86/cpu.c
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright © 2018, VideoLAN and vmaf authors
```
kylophone, line 2: oops, fix this copyright message, should say dav1d.

PR #641, opened 2020-08-07
add missing includes

cosmin:
Trying to build the latest version lead to a few errors about implicit function declaration, such as

```
../src/darray.c:26:2: error: implicitly declaring library function 'memset' with type 'void *(void *, int, unsigned long)' [-Werror,-Wimplicit-function-declaration]
```

Rather than try to work around it I tracked them down and added the missing includes which allows this to build cleanly for me.

### DIFF
--- a/libvmaf/src/darray.c
+++ b/libvmaf/src/darray.c
@@ -17,6 +17,7 @@
  */
 
 #include <stdlib.h>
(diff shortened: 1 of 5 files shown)


kylophone:
Thank you. Merged.

PR #648, opened 2020-08-13
bug fix:  No module named 'vmaf' and module 'vmaf.svmutil' has no attribute 'svm_load_model'

OnTheWay111:
Sorry, I have updated the pull request information.
1.  bug fix:  "No module named 'vmaf'"

```
# the error info:
from vmaf.config import VmafConfig, DisplayConfig
ModuleNotFoundError: No module named 'vmaf'
```

2. bug fix: "module error 'svmutil'  "
```
File /vmaf/python/vmaf/core/train_test_model.py, line 905, in _from_info_loaded
    model = svmutil.svm_load_model(filename + '.model')
AttributeError: module 'vmaf.svmutil' has no attribute 'svm_load_model'
```

### DIFF
--- a/python/vmaf/script/run_vmaf.py
+++ b/python/vmaf/script/run_vmaf.py
@@ -6,8 +6,12 @@ matplotlib.use('Agg')
 import sys
 import os
 
-import numpy as np
+curpath = os.path.abspath(os.path.dirname(__file__))
+root_path = os.path.split(os.path.split(curpath)[0])[0]
+if root_path not in sys.path:
+    sys.path.append(root_path)
 
+import numpy as np
 from vmaf.config import VmafConfig, DisplayConfig
 from vmaf.core.asset import Asset
 from vmaf.core.quality_runner import VmafQualityRunner
@@ -32,10 +36,12 @@ def print_usage():
 
 
 def main():
+    # 检查参数数量
     if len(sys.argv) < 6:
         print_usage()
         return 2
 
+    # 获取各参赛
     try:
         fmt = sys.argv[1]
         width = int(sys.argv[2])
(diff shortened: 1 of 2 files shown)


1480c1:
Could you make the commit title (the first line) a bit more clearer or add details? Having the first thing I see in a git log being `bug fix:` isn't very explanatory of what it's supposed to be

1480c1:
Some suggestions:

- `Fix "No module named 'vmaf'"`
- `bug fix: Python module error`

etc

python/vmaf/script/run_vmaf.py
```diff
@@ -32,10 +36,12 @@ def print_usage():
 
 
 def main():
+    # 检查参数数量
```
li-zhi, line 39: Would prefer not to have non-English comments.

li-zhi:
@OnTheWay111 i wonder if prefixing the PYTHON_PATH environment variable to your command line would solve the problem. Have you tried something like:
https://github.com/Netflix/vmaf/blob/master/resource/doc/VMAF_Python_library.md#run_vmaf----running-vmaf-in-single-mode

Admittedly the VMAF python library is not in its best state at this moment. Some investment on properly packaging and publishing it would be needed.

python/vmaf/script/run_vmaf.py
OnTheWay111: > Would prefer not to have non-English comments.

ok, thanks for your reply

OnTheWay111:
> @OnTheWay111 i wonder if prefixing the PYTHON_PATH environment variable to your command line would solve the problem. Have you tried something like:
> https://github.com/Netflix/vmaf/blob/master/resource/doc/VMAF_Python_library.md#run_vmaf----running-vmaf-in-single-mode
> 
> Admittedly the VMAF python library is not in its best state at this moment. Some investment on properly packaging and publishing it would be needed.

I had tryed, the problem is still exist

nilfm:
Cleaning up this very old PR. If you still face this problem, please open an issue with details about your environment and we will investigate a fix.

PR #649, opened 2020-08-14
Add motion_force_zero option; add tests.

li-zhi:


kylophone:
Two thoughts.

- If there are other values besides 0 which might be useful to force, may the option can be a float.
- Instead of the 3 `if (s->motion_force_zero)` checks, just do this once at the beginning of `.extract()`. Also, you should be able to set `.flush()` to NULL in `.init()` since there will be no buffer to drain.

li-zhi:
1. Not I can think of. forcing zero is really to force VMAF to behave like
a static image metric.
2. Not sure if I understand the flushing mechanism you wrote. If I set the
flush to NULL, would it yield one less frame, which is the last frame?

li-zhi:
More on 2: if I avoid 3 if (s->motion_force_zero) by doing it once at the
beginning, I'll have to repeat other checks (like if (index == 0) and if (s
->debug)) in the beginning block. Is this right? Then it wouldn't help
simplify the code.

kylophone:
My suggestion is to just do the following once at the top of `.extract()`, and set `.flush()` to NULL.

```c
if (s->motion_force_zero) {
    return vmaf_feature_collector_append(feature_collector,
            "'VMAF_feature_motion2_score'",
            0., index);
}

...
```

kylophone:
> 2. Not sure if I understand the flushing mechanism you wrote. If I set the flush to NULL, would it yield one less frame, which is the last frame?

Because there is no reason to buffer frames anymore, flush is not necessary.

li-zhi:
Ok. I updated the commit. Let me know if this one is better.

PR #650, opened 2020-08-15
Fix inverted height and width

ltrudeau-twoorioles:
No impact on the final score, changed just for consistency.


kylophone:
Thank you. Merged.

PR #658, opened 2020-08-26
Add my simple shell script to the readme.

Audition-CSBlock:
It took me a while to fully install libvmaf, so i thought of making a [shell script](https://gist.github.com/Audition-CSBlock/bef34e553132efad883c0f128c46d638) for easier usage.

### DIFF
--- a/resource/doc/ffmpeg.md
+++ b/resource/doc/ffmpeg.md
@@ -12,6 +12,7 @@ make install
 Using FFmpeg+libvmaf is very powerful, as you can create complex filters to calculate VMAF directly on videos of different encoding formats and resolutions. For the best practices of computing VMAF at the right resolution, refer to our [techblog](https://medium.com/netflix-techblog/vmaf-the-journey-continues-44b51ee9ed12). 
 
 We provide a few examples how you can construct the FFmpeg command line and use VMAF as a filter. Note that you may need to download the test videos from [vmaf_resource](https://github.com/Netflix/vmaf_resource/tree/master/python/test/resource).
+If all of this is too confusing and you've already installed libvmaf (easiest through homebrew or any package manager) and configured ffmpeg you could use [this](https://gist.github.com/Audition-CSBlock/bef34e553132efad883c0f128c46d638) shell script.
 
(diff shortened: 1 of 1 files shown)


slhck:
But this doesn't help with installation, right? It is just a wrapper for running the command line with ffmpeg. I am also not sure what you mean by "too confusing".

While certainly helpful, I think it would be better as an external resource listed at the bottom of the page. I know there are similar scripts or tools such as yours. We could list them there as well.

slhck:
To give more context, I think users would be better served by showing them how the raw command works in ffmpeg, not abstracting it away in a script. Teach a man how to fish etc. If the ffmpeg command needs more explanation that could certainly be added.

resource/doc/ffmpeg.md
```diff
@@ -12,6 +12,7 @@ make install
 Using FFmpeg+libvmaf is very powerful, as you can create complex filters to calculate VMAF directly on videos of different encoding formats and resolutions. For the best practices of computing VMAF at the right resolution, refer to our [techblog](https://medium.com/netflix-techblog/vmaf-the-journey-continues-44b51ee9ed12). 
 
 We provide a few examples how you can construct the FFmpeg command line and use VMAF as a filter. Note that you may need to download the test videos from [vmaf_resource](https://github.com/Netflix/vmaf_resource/tree/master/python/test/resource).
+If all of this is too confusing and you've already installed libvmaf (easiest through homebrew or any package manager) and configured ffmpeg you could use [this](https://gist.github.com/Audition-CSBlock/bef34e553132efad883c0f128c46d638) shell script.
```
li-zhi, line 15: Would you like to add this script to the repo instead of providing an external link? This way it is easier to maintain.

li-zhi:
@slhck I agree with your point.

Audition-CSBlock:
Ah yeah i see your perspective now, I'll just keep it to my gist or make a youtube tutorial about the whole thing 👍

slhck:
I think it would be a good compromise to simply add a link to your script (and possibly others') at the bottom of the page under an "External Resources" heading or similar. That way it's clear that those are third-party contributions.

PR #660, opened 2020-08-27
read_pictures: allocate and copy data into translated pics

kylophone:
This PR allocates and copies data into translated pictures during each call to `vmaf_read_pictures()`. These new pictures are exact copies of the originals, just mirrored and rotated 90 degrees. The reason this is a useful thing is because, provided these translated pictures, allows a feature extractor to do operations in Y direction in continuous memory. This means convolutions in the Y direction could be SIMD'd just as X convolutions already have been. This PR is still in draft form, and is mostly useful at this point for benchmarking purposes.  My benchmarking for the overhead of this is given a 500 frame 1080p sequence shows that the addition of this patch brings a speed overhead of **3%**. The command I've used for testing is as follows:

```sh
./build/tools/vmaf_rc \
    --model path=../model/vmaf_v0.6.1.pkl \
    --output c_motion.xml -q
```

### DIFF
--- a/libvmaf/src/libvmaf.rc.c
+++ b/libvmaf/src/libvmaf.rc.c
@@ -275,6 +275,18 @@ static int validate_pic_params(VmafContext *vmaf, VmafPicture *ref,
     return 0;
 }
 
+static void translate_picture(VmafPicture *pic_a, VmafPicture *pic_b)
+{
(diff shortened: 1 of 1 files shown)


kylophone:
Replaced by #671.

PR #666, opened 2020-09-15
Modifications to accept yuv 12 bit inputs, removed pixel normalization

ramabyg:
Signed-off-by: Rama <career4rama@gmail.com>
Removed normalization to 8 bit from higher bit depth, not sure this is simple change or we need to change all filter weights, re train etc.

### DIFF
--- a/libvmaf/src/combo.c
+++ b/libvmaf/src/combo.c
@@ -365,6 +365,18 @@ void* combo_threadfunc(void* vmaf_thread_data)
                 // max psnr 72.0 for 10-bit per Ioannis
                 ret = compute_ansnr(ref_buf, dis_buf, w, h, stride, stride, &score, &score_psnr, 255.75, 72.0);
             }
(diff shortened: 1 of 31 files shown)


kylophone:
There are a few failed Travis tests, please have a look and see what you find.

ramabyg:
Hi Kylophone,

10 bit tests are failing because I removed, normalization 
row_ptr[j] = tmp_buf[j] / 4.0 + off; to row_ptr[j] = tmp_buf[j] + off;

So wanted to understand what is needed to use 10/12 bit pixels instead of converting them to 10 bit. So that VMAF can be used for 10 or 12 bit HDR files?

-
Rama

li-zhi:
@ramabyg I'll look into this in detail tomorrow but a quick reply to your question above: you should keep row_ptr[j] = tmp_buf[j] / 4.0 + off; as this is to normalize a 10-bit input to the range of [-128, 127] (same as 8-bit) so that the same pipeline can be used to handle 10-bit. Note that no precision is lost since tmp_buf is represented by a floating-point number instead of integer. And for 12-bit input, you need to divide by 16.0 instead of 4.0.

PR #669, opened 2020-09-22
libvmaf: introduce VmafModelCollection

kylophone:
This PR introduces a new type called `VmafModelCollection`. It's exactly what it sounds like, a collection of `VmafModel`s. This will be used in conjunction with bootstrap and/or random forest models, i.e. whenever a single prediction involves more than one input models, this type is useful. 

Also of note for RC users, this PR includes a breaking change:`vmaf_load_model_from_path()`. This is likely the very last breaking change to the API before `libvmaf_rc` becomes `libvmaf`.


kylophone:
Replaced by #682.

PR #672, opened 2020-09-24
Added external resources to vmaf for ffmpeg guide

Audition-CSBlock:
As said in #658 that it would be a better idea to make a header for all the external resources.
Now other people might be able to add their ideas too 👍

### DIFF
--- a/resource/doc/ffmpeg.md
+++ b/resource/doc/ffmpeg.md
@@ -12,6 +12,7 @@ make install
 Using FFmpeg+libvmaf is very powerful, as you can create complex filters to calculate VMAF directly on videos of different encoding formats and resolutions. For the best practices of computing VMAF at the right resolution, refer to our [techblog](https://medium.com/netflix-techblog/vmaf-the-journey-continues-44b51ee9ed12). 
 
 We provide a few examples how you can construct the FFmpeg command line and use VMAF as a filter. Note that you may need to download the test videos from [vmaf_resource](https://github.com/Netflix/vmaf_resource/tree/master/python/test/resource).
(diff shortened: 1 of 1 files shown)


kylophone:
Hmm, I'm torn on whether this is necessary. I'll think about it and let others comment too. Calling FFmpeg directly instead of wrapping it allows for much more flexibility, especially when it comes to custom filtergraphs, etc. In any case, I'd prefer if helper scripts like these were in-tree instead of linked out to an external gist.

slhck:
In general I would like to have a community-curated list of third-party tools using VMAF. Running via ffmpeg directly has its benefits, but from my experience people often struggle with putting together a command and maybe want a more friendly interface. And often you don't need extra filters or options.
Of course, anything more than just a helper script wouldn't be made part of this repo anyway. For instance, I'd like to link to [my Python module](https://github.com/slhck/ffmpeg-quality-metrics), but of course it shouldn't be part of VMAF itself.

For the particular script in question, I can see how it could be vendored because it's just a simple wrapper, not more, not less. I could also see how options related to scaling could be added.

May I suggest toning down the description and making it more objective? Instead of:

> Shell script for running your ffmpeg and vmaf installation with ease

Something like:

> Bash wrapper script for running `libvmaf` through `ffmpeg`

li-zhi:
I think link-outs to third-party tools will be useful, despite the risk of potential broken links. The contributors just need to be more diligent in maintaining the links.

I am debating on if we should have a centralized page titled "Third-party Resources" and have individual pages to link to sections in there, or just have distributed sections in different pages. Perhaps starting from distributed, and later we can move things around.

PR #678, opened 2020-10-06
Doc/ffmpeg.md: Add note about using another folder and also for Windows

1480c1:
Adds some notes about Windows and paths passed to `model_path` #413


1480c1:
My English skills are falling off as of lately due to not having an English class recently and also due to the diminished amount of physical human contact I have had recently, so please do be critical of the wording to make sure it makes sense to a layman Windows user.

li-zhi:
This looks perfect. Thanks for the detailed writeup.

Issue #374, opened 2019-10-08
Fails to build on x86 without SSE


jbeich:
i386 target on FreeBSD, NetBSD, OpenBSD (and maybe some Linux distributions) lacks SSE unless `-msse` is passed. For one, FreeBSD i386 targets `-march=i486` by default but [may switch](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=230888#c12) to `-march=i586` soon.

```c++
$ c++ -v
FreeBSD clang version 8.0.0 (tags/RELEASE_800/final 356365) (based on LLVM 8.0.0)
Target: i386-unknown-freebsd11.3
Thread model: posix
InstalledDir: /usr/bin

$ meson _build
$ ninja -C _build
[...]
FAILED: vmaf@sta/src_libvmaf_src_adm_tools.c.o
cc -Ivmaf@sta -I. -I.. -I.././src/ptools/opencontainers_1_8_4/include -I.././src/libvmaf/src/ -I.././src/libvmaf/src/common -I.././src/ptools/ -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -g -fPIC -std=c99 -w -Wextra -pedantic -D_FILE_OFFSET_BITS=64 '-D MULTI_THREADING' -DOC_NEW_STYLE_INCLUDES -MD -MQ 'vmaf@sta/src_libvmaf_src_adm_tools.c.o' -MF 'vmaf@sta/src_libvmaf_src_adm_tools.c.o.d' -o 'vmaf@sta/src_libvmaf_src_adm_tools.c.o' -c ../src/libvmaf/src/adm_tools.c
../src/libvmaf/src/adm_tools.c:36:16: error: always_inline function '_mm_cvtss_f32' requires target feature 'sse', but would be inlined into function 'rcp_s' that is compiled without support for 'sse'
    float xi = _mm_cvtss_f32(_mm_rcp_ss(_mm_load_ss(&x)));
               ^
../src/libvmaf/src/adm_tools.c:36:30: error: always_inline function '_mm_rcp_ss' requires target feature 'sse', but would be inlined into function 'rcp_s' that is compiled without support for 'sse'
    float xi = _mm_cvtss_f32(_mm_rcp_ss(_mm_load_ss(&x)));
                             ^
../src/libvmaf/src/adm_tools.c:36:41: error: always_inline function '_mm_load_ss' requires target feature 'sse', but would be inlined into function 'rcp_s' that is compiled without support for 'sse'
[...]

PR #681, opened 2020-10-17
build: unbreak x86 builds

jengelh:
SSE is not guaranteed to be present on x86 hosts, so i386..i686
compilers default to -mno-sse. That however makes the build fail
because vmaf unconditionally uses emmintrin functions which are
not available under no-sse. ppc64le builds however are just fine, so...

Change the macro guarding emmintrin to not match on x86-like
architectures, but to match on the enablement of SSE2. The macro
"__SSE2__" is the gcc name and is true whenever -msse2 is enabled
(explicitly or implicitly); clang should behave the same. I have not
tested this under other compilers such as Microsoft, though.

Fixes: #374

### DIFF
--- a/libvmaf/src/feature/adm_tools.c
+++ b/libvmaf/src/feature/adm_tools.c
@@ -35,7 +35,7 @@
 #define MIN(x, y) (((x) < (y)) ? (x) : (y))
 #define MAX(x, y) (((x) > (y)) ? (x) : (y))
 
-#if ARCH_X86
+#ifdef __SSE2__
 #ifdef ADM_OPT_RECIP_DIVISION
 
 #include <emmintrin.h>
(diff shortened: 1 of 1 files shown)


kylophone:
Merged. Ideally we could add a i386 CI runner to avoid breaking this in the future, but I'm not sure that is supported by our CI without self-hosting.

1480c1:
I believe it's possible to do so by using the `i686-linux-gnu-g++` and `i686-w64-mingw32-g++` compiler, the second one is available in the windows image. By the way, for the mismatch between 32-bit and 64-bit libvmaf outputs, I fixed it locally by adding `-msse2 -mfpmath=sse -mstackrealign` to the cflags as that's what fedora does for their 32-bit packages

PR #695, opened 2020-11-02
Actions/windows: use the new syntax for setting environment variables

1480c1:
Also changes the flags "used" for the 32-bit jobs, still disabled since the output is `93.656205` instead of `93.656203`


kylophone:
Thanks, merged.

PR #697, opened 2020-11-03
libvmaf/feature: add vif_enhn_gain_limit param to integer vif

kylophone:
This PR adds the `vif_enhn_gain_limit` parameter to the integer vif feature extractor. Also in this PR is a replacement of the `log2f_approx` function with `log2` from `math.h`. `log2f_approx` was necessary to speed up the floating point version of this feature, but it is both not useful and less accurate in the integer version.

```
./build/tools/vmaf_rc \
    --feature float_vif=vif_enhn_gain_limit=1.0 \
    --feature vif=vif_enhn_gain_limit=1.0 \
    --no_prediction \
    --output vif_neg.xml
```


li-zhi:
@kylophone do you want me to add some tests cases to this or you could add them?

kylophone:
> @kylophone do you want me to add some tests cases to this or you could add them?

I'm going to experiment with one more thing (fixing the symmetry of the mirrored edges) that will slightly change the numeric values. Let's wait for that first.

kylophone:
I've added one more commit which fixes the OOO padding issues. Still needs to be numerically validated.

PR #701, opened 2020-11-05
libvmaf/feature: add adm_enhn_gain_limit param to integer adm

kylophone:
This PR adds the vif_enhn_gain_limit parameter to the integer vif feature extractor.

```
./build/tools/vmaf_rc \
    --feature float_adm=adm_enhn_gain_limit=1.0 \
    --feature adm=adm_enhn_gain_limit=1.0 \
    --no_prediction \
    --output adm.xml
```


kylophone:
I'm not sure if this check is better before or after the integer round. When we validate, I guess we should try both.

https://github.com/Netflix/vmaf/pull/701/files#diff-71cc41fe7048251bf53ce46f9e4cc29a32898c0115ce2282836fcd16d39b21b0R725-R739

PR #718, opened 2020-11-19
libvmaf/x86: add AVX-512 optimizations for vif and motion

kylophone:
This PR adds AVX-512 optimizations for the VIF and motion feature extractors. To build with AVX-512 you must supply the meson option `enable_avx512`, for example: `meson setup -Denable_avx512=true build`. To mask AVX-512 execution at runtime set `--cpumask 16` when calling `vmaf_rc`.


kylophone:
All of the tox tests are passing on AVX-512.

PR #723, opened 2020-11-23
libvmaf/feature: rename features according to float/integer model requirements

kylophone:
This PR renames the feature names written by the integer vif/adm/motion feature extractors to match what is published in the new model files. Additionally, features written by the floating point feature extractors are now logged with the `float_` prefix and the integer versions of these features no longer have the `integer_` prefix.


kylophone:
Pushed another commit which does not invert the `_float`/`_integer` suffixes. Will possibly pursue this in another PR, in which case 3a4794915d49a42e33b7b4282852fc416cc792d6 could just be reverted.

Issue #709, opened 2020-11-12
[Feature Request]: Would it be possible to embed a default model file inside libvmaf?


1480c1:
It would be nice to be able to embed a default model file inside the libvmaf as something like a `static char array` that libvmaf would read from instead of a file if the model_path is NULL or zero-length so it would be easier to use on systems with no good defaults for where to read from (windows).

I know on Linux, you can use `xxd -i file` to print out an `unsigned char` array composed of hexadecimal, but I am not sure what would be a good equivalent on windows and macOS

kylophone:
Yes, this is actually a plan of mine. We will compile in a few models.

kylophone:
#724 merged.

1480c1:
@kylophone how hard would it be to add something like a "keyword" to use the internal model files when passed as the `model_path` through ffmpeg and others (including aomenc) until proper support has been added on their side? Something like `libvmaf=model_path="NULL"` or similar to use the default model instead of looking for a file called `NULL`?

kylophone:
I would rather approach this from the `libaom` side. They were already linking our release candidate library, so they'll have to rewrite a portion of that code anyway. I can reach out to them and report back here. Obviously the main advantage to using the built-in models is the fact that everything is self contained and nothing auxiliary is required. The way we approached this in `vmaf` was to just always default to `v0.6.1` whenever `--model` was not passed.

1480c1:
I see, thanks

PR #724, opened 2020-11-24
libvmaf: add support for built-in models

kylophone:
Models can now be compiled into the library and read directly from the executable. This is a feature which is in addition to the current way of doing things, i.e an external model file which must be read from a filesystem. The way that this works, is via meson the *.json model files are turned into c headers via `xxd --include`. These headers are then loaded and read with `pdjson` in buffer mode instead of stream mode. Optionally, one could compile `libvmaf` without built-in models by setting the meson compilation option `built_in_models` to false (`-Dbuilt_in_models=false`). The binary size increases with built-in models compiled in, and for that reason one may want to disable this if that was an application constraint. For example, on my machine, the binary sizes with this turned on/off are as follows: 2545056 bytes vs. 1651112 bytes.

The new built-in models are read with the `version=` parameter and model files are read with the `path=` parameter (as usual).
```sh
    --model version=vmaf_float_v0.6.1
    --model path=../model/vmaf_float_v0.6.1.json
```

@1480c1, this should solve #709. Do you think you could try getting this working on Windows? I believe all that would be required is a working `xxd`. The build won't fail without `xxd`, but you will not get built-in models without it. Hopefully there's no other cross-platform wonkiness.


kylophone:
FYI, for Windows `xxd`: https://github.com/msys2/MINGW-packages/issues/3518#issuecomment-376788576

1480c1:
Would it be possible to do something like "importing" https://github.com/vim/vim/blob/master/src/xxd/xxd.c as a third-party program file and if `xxd` is not found, compile that as a helper program and run that instead? They seem to have code for MSVC as well

I'm not entirely sure about licensing since I am not close enough to be a lawyer but they seem to use
```c
 * I hereby grant permission to distribute and use xxd
 * under X11-MIT or GPL-2.0 (at the user's choice).
 *
 * Small changes made afterwards by Bram Moolenaar et al.
 *
 * Distribute freely and credit me,
 * make money and share with me,
 * lose money and don't ask me.
```

kylophone:
Hmm. I'm not sure. I think it is better to avoid that and use the pre-packaged command line tool. mingw-w64 should have a package, right?

1480c1:
indeed msys2 has xxd packaged with vim, but this is for people who don't use msys2 and want to try to use vmaf inside visual studio or similar

kylophone:
Looks like ad10de6 works as expected for msys2 builds. I pulled down the build artifact, and it works. I'm open to some other solution that makes life easier for native Windows builds in Visual Studio. @1480c1, any concerns if I just merge this as-is for now?

1480c1:
LGTM personally since msvc is a low prio for me

libvmaf/src/meson.build
```diff
@@ -94,6 +96,34 @@ if is_asm_enabled
     endif
 endif
 
+json_model_c_sources = []
+built_in_models_enabled = get_option('built_in_models') == true
+
+if built_in_models_enabled
+    xxd = find_program('xxd')
```
1480c1, line 103: just found out that `find_program` has `required` set to true by default https://mesonbuild.com/Reference-manual.html#find_program

kylophone: Do we need to change anything? If `required` was true shouldn't the msys2 build have failed before I added the `xxd` package to the CI build?

1480c1: I believe we do since it seems it will automatically fail on that line if `xxd` is not found https://gitlab.com/Luigi311/encoders-docker/-/jobs/870743576#L837

/cc @luigi311

1480c1:
```suggestion
    xxd = find_program('xxd', required : false)
```
I believe the required change is

luigi311: I can confirm that this is an issue, i added xxd to my dependencies and it built successfully.

kylophone: Pushed aa85f6db2e2357221ac0bb339633c03e3cea9f9f, should not be an issue anymore. Let me know if it is.

luigi311: It no longer instantly fails but it still does fail during the build process. Here is the build log https://gitlab.com/Luigi311/encoders-docker/-/jobs/870828946 i dont see any actual error messages though

1480c1:
```c
../src/model.c:36:31: error: zero or negative size array 'built_in_models'
   36 | static const VmafBuiltInModel built_in_models[] = {
      |                               ^~~~~~~~~~~~~~~
```

kylophone: Oh, interesting. I tested this and my compiled did not mind this.

1480c1:
```none
C compiler for the host machine: cc (gcc 9.3.0 "cc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0")
C linker for the host machine: cc ld.bfd 2.34
C++ compiler for the host machine: c++ (gcc 9.3.0 "c++ (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0")
C++ linker for the host machine: c++ ld.bfd 2.34
```
comes from the `ubuntu:20.04` container

luigi311: This is with gcc-9-base:amd64 (9.3.0-17ubuntu1~20.04)

kylophone: Nevermind, It's always wrong even if the compiler doesn't blow up. I will fix.

ISO 9899:2011 6.7.6.2:
> If the expression is a constant expression, it shall have a value greater than zero.

kylophone: PR #729 should fix this. I'll merge when CI is green.

luigi311: I am not longer getting build errors. Thanks @kylophone

1480c1:
Just to make sure, is there a way to access the internal libraries from ffmpeg? I'm assuming no

kylophone:
> Just to make sure, is there a way to access the internal libraries from ffmpeg? I'm assuming no

I will have a patch coming to `ffmpeg-devel` that will add support for built-in models and some of the other new features.

PR #756, opened 2020-12-08
output pooled_metrics as JSON object instead of array

ltrudeau-twoorioles:
The vmaf_feature_name_alias is used as the key, this simplifies
the parsing code. For example, with the previous code, extracting
VMAF from the output file required something along the lines of

```
  for m in vmaf_json["pooled_metrics"]:
    if m["metric"] == "vmaf":
      print("VMAF %f" % m["pooling_methods"]["mean"])
```

it can now be achieved as follows

`  print("VMAF %f" % vmaf_json["pooled_metrics"]["vmaf"]["mean"])`

### DIFF
--- a/libvmaf/src/output.c
+++ b/libvmaf/src/output.c
@@ -163,30 +163,26 @@ int vmaf_write_output_json(VmafContext *vmaf, VmafFeatureCollector *fc,
     }
     fprintf(outfile, "\n  ],\n");
 
-    fprintf(outfile, "  \"pooled_metrics\": [");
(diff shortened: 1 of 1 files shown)


kylophone:
While technically a schema break, I think this is a reasonable change with a good rationale. I am not aware of anyone who is using this new JSON log, so I am inclined to merge this and include it in the v2.0.1 release.

slhck:
👍 for that change. I can imagine that the format as-is would cause a few headaches.

kylophone:
Alright, tested by myself and merged. Thanks, @ltrudeau-twoorioles.

PR #757, opened 2020-12-08
Don't print empty line when --quiet

ltrudeau-twoorioles:


kylophone:
Merged, thank you.

ltrudeau-twoorioles:
@kylophone: Any interest in not printing anything at all with `--quiet`? or when `--quiet ` and `--output`. I can write a quick patch for that.

PR #764, opened 2020-12-11
Add --neg to command line interface

ltrudeau-twoorioles:
This facilitates the use of no enhancement gain and reduces the
chance of user error. This could also be useful if the features
associated with neg change in the future.

### DIFF
--- a/libvmaf/tools/cli_parse.c
+++ b/libvmaf/tools/cli_parse.c
@@ -24,6 +24,7 @@ enum {
     ARG_SUBSAMPLE,
     ARG_CPUMASK,
     ARG_AOM_CTC,
+    ARG_NEG,
 };
 
 static const struct option long_opts[] = {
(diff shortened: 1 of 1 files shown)


kylophone:
I think this is a good addition, although I'm currently trying to get something working where `vmaf` and `vmaf_neg` could possibly be computed simultaneously. They share about 95% of the same computation, so it's worthwhile to do so. I'll probably merge this patch or something similar which accomplishes the same thing. I'm looking at this now, so I'll keep you updated.

ltrudeau-twoorioles:
Thanks for looking into this, it would be an awesome feature.

li-zhi:
Sounds good to me.

kylophone:
Hi, @ltrudeau-twoorioles

Wanted to give you an update on this. Since https://github.com/Netflix/vmaf/pull/789 was merged, we now support templated feature names, which means that VMAF and VMAF NEG can be computed at the same time. As a side effect, it also means that the old way of globally overloading a feature via the `--feature` flag is not backwards compatible. I know you were a user of that, so I just wanted to make sure you saw this. While we want to be sure to maintain as much backwards compatibility, this was just a case where we were fundamentally broken for this use case. 

As for this patch and the `--neg` flag. I think it is still useful. I would like to modify your patch to essentially set the overloads for all the provided models, any concerns with that? I can run with this and do the work unless you have any concerns.

Thanks,
Kyle

ltrudeau-twoorioles:
Hi @kylophone,

Thanks for the update, we are very happy that we can run both VMAF and VMAF NEG in the same run.

Feel free to run with this patch, my only concern for --neg is that if it applies to all models it won't be useful to use it to run both model. Maybe using this new templating feature we could have something like `version=vmaf_b_v0.6.3:name=vmaf_neg:neg=1`

The idea is to have a shorthand for NEG so that it is able to change in the future and to reduce mistakes when specifying it.

PR #793, opened 2020-12-30
Update instructions for accessing vqeghd3

cosmin:
The CDVL search for `vqeghd3` is a bit confusing. Searching for this keyword returns 168 results of mostly the individual files, but they are no longer available for download, returning errors for each link. I contacted @mpinsonntia who provided me with updated instructions to access the full dataset, which has been combined into a single ~84 GB ZIP file. I updated the text using these instructions.


li-zhi:
Thanks for updating!

PR #798, opened 2021-01-06
Do not display collection scores if quiet and output are set

ltrudeau-twoorioles:
The scores will be available in the output file and are probably
unwanted in the standard output if the user specified --quiet


kylophone:
Thank you, merged.

PR #801, opened 2021-01-13
Remove old source files from tree, move float code into its own directory

kylophone:
This PR has two commits. The first commit removes a number of unused source files from the tree. The second commit moves all float code into a separate directory.


li-zhi:
@kylophone is this one still active?

PR #803, opened 2021-01-17
Compilation fix on FreeBSD

da2x:
Didn't compile on FreeBSD, since _SC_NPROCESSORS_ONLN  is undefined there. Instead, use the hw.ncpu sysctl.

### DIFF
--- a/libvmaf/src/cpu_info.c
+++ b/libvmaf/src/cpu_info.c
@@ -10,6 +10,9 @@
 #elif MACOS
 #include <sys/param.h>
 #include <sys/sysctl.h>
+#elif __FreeBSD__
+#include <sys/types.h>
+#include <sys/sysctl.h>
 #else
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


jbeich:
`_SC_NPROCESSORS_ONLN` is [supported](https://github.com/freebsd/freebsd-src/commit/4d44f0b5a3a126151a7bb12d8b8fd89abb95bd52) on FreeBSD for decades and is actually [more efficient](https://github.com/freebsd/freebsd-src/commit/a87855124ca1076938598c469b52c4cd95cb4ae8) than poking `hw.ncpu`. See [downstream fix](https://github.com/freebsd/freebsd-ports/blob/master/multimedia/vmaf/files/patch-meson.build).

li-zhi:
@da2x do you plan to work on this as suggested by @jbeich ? I'll close the PR first as it seems it is not actively worked on.

PR #811, opened 2021-01-29
Add links to FFmpeg binaries that support libvmaf.

BassThatHertz:
Useful for those who are too lazy to compile FFmpeg.


li-zhi:
Recommendation is to add the links to the external resource page:
https://github.com/Netflix/vmaf/blob/master/resource/doc/external_resource.md

PR #812, opened 2021-01-29
include_dirs do not appear to be getting passed properly on OS X 

cosmin:
I'm unable to build latest VMAF on OS X, using Python 3.9.0 from homebrew (although the same problem repro'd with 3.7.5), and another coworker had the exact same problem on another machine.

The error is `unable to find "numpy/arrayobject.h"` in Cython, even though numpy is correctly installed, and `numpy.get_include()` returns the correct path (where I verified the headers exist).

For some reason `include_dirs` in setup.py is not being passed down. After thinking I'm losing my mind I did some searching and I see that other projects are experiencing the same problem, see for example https://github.com/hmmlearn/hmmlearn/issues/43 and the various other issues linking to it.

So I used the same workaround, instead manually setting the include dirs on the ext_module, and that seems to work just fine. No idea why `include_dirs` is not being respected properly though. It feels somewhat dirty but this is the only way I was able to get unblocked with building on OS X.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -28,6 +28,8 @@ def get_version():
         pass
     return "0.0-dev"
 
+ext_module = cythonize(['vmaf/core/adm_dwt2_cy.pyx'])
+ext_module[0].include_dirs = [numpy.get_include(), '../libvmaf/src']
 
 setup(
     name="vmaf",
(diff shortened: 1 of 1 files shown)


li-zhi:
Thanks for noticinng this and provide a solution. It seems both github and travis CI are still py3.8. I'll merge it.

PR #813, opened 2021-01-29
add five-parameter logistic function as a model option

cosmin:
Several papers such as [1][2] have mentioned that the following five-parameter logistic function (5PL) from [3] provide a good fit for linearizing non-linear VQA scores such as SSIM

image

I wanted to make this an option for training single feature models such as only based on SSIM. Based on tests in NFLX_public dataset this does appear to linearize slightly better than the current sigmoid, improving PCC from 0.750 to 0.770. Remains to be seen how this works on other data sets, but it feels like a reasonable option to have for evaluation purposes.

[1] A. K. Venkataramanan, C. Wu, A. C. Bovik, I. Katsavounidis and Z. Shahid, "A Hitchhiker's Guide to Structural Similarity". Submitted to IEEE Access.

[2] S. Athar and Z. Wang, "A Comprehensive Performance Evaluation of Image Quality Assessment Algorithms," in IEEE Access, vol. 7, pp. 140030-140070, 2019, doi: 10.1109/ACCESS.2019.2943319.

[3] H.R. Sheikh, M.F. Sabir, and A.C. Bovik, “A Statistical Evaluation of Recent Full Reference Image Quality Assessment Algorithms,” IEEE Transactions on Image Processing, vol. 15, pp. 3441-3452, November 2006.

### DIFF
--- a/python/vmaf/core/train_test_model.py
+++ b/python/vmaf/core/train_test_model.py
@@ -740,7 +740,7 @@ class TrainTestModel(TypeVersionEnabled):
             os.remove(filename)
 
     @classmethod
-    def get_xs_from_results(cls, results, indexs=None, aggregate=True):
+    def get_xs_from_results(cls, results, indexs=None, aggregate=True, features=None):
         """
         :param results: list of BasicResult, or pandas.DataFrame
         :param indexs: indices of results to be used
@@ -756,8 +756,11 @@ class TrainTestModel(TypeVersionEnabled):
             # or get_ordered_list_scores_key. Instead, just get the sorted keys
             feature_names = results[0].get_ordered_results()
 
-        feature_names = list(feature_names)
-        cls._assert_dimension(feature_names, results)
+        if features is not None:
+            feature_names = [f for f in feature_names if f in features]
+        else:
+            feature_names = list(feature_names)            
+            cls._assert_dimension(feature_names, results)
 
         # collect results into xs
         xs = {}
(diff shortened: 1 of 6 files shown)


li-zhi:
@cosmin perhaps add a test for this TrainTestModel class?

cosmin:
Will do. Otherwise does the general approach seem sound or is there anything else I should change about the implementation?

li-zhi:
Looks good to me. I'll take a second look once you have a test case. You may follow an example like this:
https://github.com/Netflix/vmaf/blob/master/python/test/train_test_model_test.py#L84

cosmin:
@li-zhi I added a test, and also extended the `./unittest` utility to make it easier to run a single test by name.

Since the `MomentNorefFeatureExtractor` no-ref feature extractor returns 3 features, I need to pick a single one for the test. I didn't know the best way to manipulate the returned `xs` and `xys` objects within the test, so I added an option to `get_xs_from_results()` to get specific features back, and then I'm manually constructing the `xys` so I don't have to also modify `get_xys_from_results()`.

Is there a better way I should be doing this instead?

cosmin:
I’ll take a look and see why the tests are failing.

cosmin:
I don't see any relationship to my change, the tests appear to be failing because the doctests in `ListStats` are being over precise with float formatting. AFAICT the way to deal with this in doctest is to instead output a string format of the float value where we can specify the precision. I have a change to do just this, but as it's unrelated to this change I'm going to submit a separate pull request for that.

cosmin:
Fixed the doctests in #817

li-zhi:
Let me restart the master branch to see if the issue can be reproduced.

python/vmaf/core/train_test_model.py
```diff
@@ -740,7 +740,7 @@ def _delete(filename, **more):
             os.remove(filename)
 
     @classmethod
-    def get_xs_from_results(cls, results, indexs=None, aggregate=True):
+    def get_xs_from_results(cls, results, indexs=None, aggregate=True, features=None):
```
li-zhi, line 743: Can you also update the signature of the derived class `NiqeTrainTestModel` under niqe_train_test_model.py?

li-zhi:
It can be reproduced in master. Just merged the other PR created. Thanks for fixing it.

cosmin:
rebasing to latest master

li-zhi:
@cosmin FYI. I moved 5PL_v1.py to resource/model_param.

cosmin:
Thank you, that’s where I intended to put it but apparently I managed to screw that up and didn’t catch it.

PR #816, opened 2021-01-30
Corrected default model_path location.

BassThatHertz:
When using libvmaf with FFmpeg, if you do not use the `model_path` option, you get the error:
 `problem loading model file: /usr/local/share/model/vmaf_v0.6.1.pkl`

 Therefore, the default location of `model_path` is `/usr/local/share/model/vmaf_v0.6.1.pkl`, not `/usr/local/share/model/vmaf_float_v0.6.1.json`.

I confirmed this by going to FFmpeg's GitHub repository and looking the code for the libvmaf filter: https://github.com/FFmpeg/FFmpeg/blob/069d2b4a50a6eb2f925f36884e6b9bd9a1e54670/libavfilter/vf_libvmaf.c#L75

### DIFF
--- a/resource/doc/ffmpeg.md
+++ b/resource/doc/ffmpeg.md
@@ -14,7 +14,7 @@ We provide a few examples how you can construct the FFmpeg command line and use 
 
 Below is an example on how you can run FFmpeg+libvmaf on a pair of YUV files. First, download the reference video [`src01_hrc00_576x324.yuv`](https://github.com/Netflix/vmaf_resource/blob/master/python/test/resource/yuv/src01_hrc00_576x324.yuv) and the distorted video [`src01_hrc01_576x324.yuv`](https://github.com/Netflix/vmaf_resource/blob/master/python/test/resource/yuv/src01_hrc01_576x324.yuv). `-r 24` sets the frame rate (note that it needs to be before `-i`), and `PTS-STARTPTS` synchronizes the PTS (presentation timestamp) of the two videos (this is crucial if one of your videos does not start at PTS 0, for example, if you cut your video out of a long video stream). It is important to set the frame rate and the PTS right, since FFmpeg filters synchronize based on timestamps instead of frames.
 
-The `log_path` is set to standard output `/dev/stdout`. It uses the `model_path` at location `/usr/local/share/model/vmaf_float_v0.6.1.json` (which is the default and can be omitted).
+The `log_path` is set to standard output `/dev/stdout`. It uses the `model_path` at location `/usr/local/share/model/vmaf_v0.6.1.pkl` (which is the default and can be omitted).
 
 ```shell script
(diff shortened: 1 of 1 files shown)


kylophone:
pkl models are deprecated since v2.0.0, I will update the ffmpeg filter soon.

BassThatHertz:
I was actually thinking about doing that, I'm assuming I just need to edit the following line?: https://github.com/FFmpeg/FFmpeg/blob/069d2b4a50a6eb2f925f36884e6b9bd9a1e54670/libavfilter/vf_libvmaf.c#L75

I've never made a contribution to FFmpeg before but I believe you have to clone the repository, make the change, commit it and then do `git-send-email ffmpeg-devel@ffmpeg.org`? 

Also, should I set the default path to `/usr/local/share/model/vmaf_float_v0.6.1.json` or `/usr/local/share/model/vmaf_v0.6.1.json`? What's the difference between the "float" version and the regular version?

kylophone:
> I was actually thinking about doing that, I'm assuming I just need to edit the following line?: https://github.com/FFmpeg/FFmpeg/blob/069d2b4a50a6eb2f925f36884e6b9bd9a1e54670/libavfilter/vf_libvmaf.c#L75
> 
> I've never made a contribution to FFmpeg before but I believe you have to clone the repository, make the change, commit it and then do `git-send-email ffmpeg-devel@ffmpeg.org`?

Yes, but I can take care of it. This will be part of a much larger update, pretty much a re-write to make use all of the nice API things introduced in v2.0.0. 

> What's the difference between the "float" version and the regular version?

Numerically, not much. Speedwise, more than twice as fast.

BassThatHertz:
> Numerically, not much. Speedwise, more than twice as fast.

Which version is supposed to be more than twice as faster? I just ran a test and the float version is slower, and from reading your comment it seems like you're saying that the float version is more than twice as fast.

Also, are the differences between the float and regular version documented somewhere?

kylophone:
> > Numerically, not much. Speedwise, more than twice as fast.
> 
> Which version is supposed to be more than twice as fast? I just ran a test and the float version is slower, and from reading your comment it seems like you're saying that the float version is more than twice as fast.
> 
> Also, are the differences between the float and regular version documented somewhere?

The integer implementation is faster than float, which was in itself the reason for the update. See https://github.com/Netflix/vmaf/releases/tag/v2.0.0 and https://netflixtechblog.com/toward-a-better-quality-metric-for-the-video-community-7ed94e752a30 for details. If you'd like to run both models at the same time to see the difference yourself, you may do so with the following command. Make sure you compile with `-Denable_float=true` if you want to include those feature extractors. You should also know that floating-point VMAF is deprecated and could go away in a future release, use the new models.

```sh
./build/tools/vmaf \
    --reference reference.y4m \
    --distorted distorted.y4m \
    --model=version=vmaf_v0.6.1:name=vmaf \
    --model=version=vmaf_float_v0.6.1:name=vmaf_float \
    --output /dev/stdout
```

PR #829, opened 2021-02-20
clarify python installation instructions

cosmin:
on latest CentOS `yum` has also been replaced by `dnf`. Also at least on CentOS `ninja` is available as `ninja-build`.  `python-dev` is also called `python-devel` on CentOS. Also adding some text to clarify the use of virtualenv for those less familiar with the nuances of Python development tools.


cosmin:
After helping a few people less familiar with Python try to get up and running with VMAF model development I noticed a few common stumbling blocks when trying to follow along the instructions. I'm hoping this will help make it a bit more clear.

PR #834, opened 2021-03-01
add script to verify accuracy of alternate feature implementation

cosmin:
As part of re-implementing a feature extractor (from say optimized C code to a reference Matlab code) I added this script to cross-check that two feature extractors return matching values.

For example assuming we have a reference re-implementation of the default VMAF features in `VMAF_feature` in a new feature extractor called `VQA_feature`, then we can cross-check the implementations as follows.

```
PYTHONPATH=python ./python/vmaf/script/validate_alt_implementation.py resource/dataset/NFLX_dataset_public.py VMAF_feature VQA_feature "adm2:dlm"
```

will check all features in `VMAF_feature` and `VQA_features` that have the same atom name in both, along with a list of ref:test pairs such as `adm2:dlm` to handle cases where features don't have the same name.

```
Comparing the following feature pairs:

    VMAF_feature_motion_score => VQA_feature_motion_score
    VMAF_feature_vif_scale0_score => VQA_feature_vif_scale0_score
    VMAF_feature_vif_scale1_score => VQA_feature_vif_scale1_score
    VMAF_feature_vif_scale2_score => VQA_feature_vif_scale2_score
    VMAF_feature_vif_scale3_score => VQA_feature_vif_scale3_score
    VMAF_feature_adm2_score => VQA_feature_dlm_score
```

and then for each asset it will output something like

```
/data/NFLX_dataset_public/dis/BigBuckBunny_20_288_375.yuv
VQA_feature_motion_score: PCC 0.995 SRCC 1.000 RMSE 0.747
VQA_feature_vif_scale0_score: PCC 1.000 SRCC 1.000 RMSE 0.000
VQA_feature_vif_scale1_score: PCC 1.000 SRCC 1.000 RMSE 0.000
VQA_feature_vif_scale2_score: PCC 1.000 SRCC 1.000 RMSE 0.000
VQA_feature_vif_scale3_score: PCC 1.000 SRCC 1.000 RMSE 0.001
VQA_feature_dlm_score: PCC 1.000 SRCC 1.000 RMSE 0.000
```

This for example shows that the VIF and DLM features seem to be fine, but there's a slight difference in the `motion` feature implementation, as an example.

### DIFF
--- a/python/vmaf/script/validate_alt_implementation.py
+++ b/python/vmaf/script/validate_alt_implementation.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+import os
+import sys
+from vmaf.routine import read_dataset
+from vmaf.tools.misc import import_python_file
+from vmaf.core.feature_extractor import FeatureExtractor
+from vmaf.core.result_store import FileSystemResultStore
+from vmaf.core.perf_metric import RmsePerfMetric, SrccPerfMetric, PccPerfMetric
+
+def print_usage():
+    print("usage: " + os.path.basename(sys.argv[0]) + " dataset ref_feature_extractor test_feature_extractor [ref_feature:test_feature]*")
+
+
+def main():
+    if len(sys.argv) < 4:
+        print_usage()
+        return 2
+
+    try:
+        dataset_filepath = sys.argv[1]
+        ref_feature_extractor_type = sys.argv[2]
+        test_feature_extractor_type = sys.argv[3]
+    except ValueError:
+        print_usage()
+        return 2
+
+    test_dataset = import_python_file(dataset_filepath)
+
+    test_assets = read_dataset(test_dataset)
+    # print(test_assets)
+
+    ref_fextractor_class = FeatureExtractor.find_subclass(ref_feature_extractor_type)
+    test_fextractor_class = FeatureExtractor.find_subclass(test_feature_extractor_type)
+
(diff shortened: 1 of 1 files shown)


python/vmaf/script/validate_alt_implementation.py
```diff
@@ -0,0 +1,91 @@
...
+        print(f"    {ref_feature} => {test_feature}")
+
+    for asset in test_assets:        
+        if asset.dis_path == asset.ref_path:
+            # skip self comparisons
+            continue
+        print("")
+        print(asset.dis_path)
+
+        ref_result = result_store.load(asset, ref_extractor.executor_id)
```
li-zhi, line 64: This script assumes that the result has already been stored in result_store from a previous run?

Instead of calling result_store.load() and ref_extractor._post_process_result() directly, it is probably better to call the blessed routine:
```
        fextractor = ref_extractor(
            assets,
            None, fifo_mode=True,
            result_store=result_store
        )
        fextractor.run(parallelize=True)
        results = fextractor.results
```
to achieve similar effect. It doesn't assume that the features are pre-computed and stored.

cosmin: Makes sense, I will update the script as suggested.

li-zhi: @cosmin any update on this PR? If it is not actively worked on, I suggest you close it first.

PR #852, opened 2021-04-12
Fix the bug: DisplayConfig.show() cannot generate the result picture …

wangwei1237:
Fix the bug: DisplayConfig.show() cannot generate the result picture when the parameter write_to_dir is missing.

In `python/vmaf/script/run_vmaf_training.py`, the script set the Matplotlib default backends to `non-interactive backends`: Agg.

```
#!/usr/bin/env python3

import matplotlib
matplotlib.use('Agg')
```

But, for `DisplayConfig.show()`, the script will use the interactive backends when the parameter `write_to_dir` is missing.

### DIFF
--- a/python/vmaf/config.py
+++ b/python/vmaf/config.py
@@ -209,12 +209,22 @@ class DisplayConfig(object):
     @staticmethod
     def show(**kwargs):
         from vmaf import plt
-        if 'write_to_dir' in kwargs:
-            format = kwargs['format'] if 'format' in kwargs else 'png'
-            filedir = kwargs['write_to_dir'] if kwargs['write_to_dir'] is not None else VmafConfig.workspace_path('output')
-            os.makedirs(filedir, exist_ok=True)
-            for fignum in plt.get_fignums():
-                fig = plt.figure(fignum)
-                fig.savefig(os.path.join(filedir, str(fignum) + '.' + format), format=format)
+        import matplotlib
+
+        if matplotlib.rcParams['backend'] == 'agg':
+            if 'write_to_dir' in kwargs:
+                format = kwargs['format'] if 'format' in kwargs else 'png'
+                filedir = kwargs['write_to_dir'] if kwargs['write_to_dir'] is not None else VmafConfig.workspace_path('output')
+                os.makedirs(filedir, exist_ok=True)
+                for fignum in plt.get_fignums():
+                    fig = plt.figure(fignum)
+                    fig.savefig(os.path.join(filedir, str(fignum) + '.' + format), format=format)
+            else:
+                format = 'png'
+                filedir = VmafConfig.workspace_path('output')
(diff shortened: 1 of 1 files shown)


li-zhi:
Not sure if I get the motivation for this PR. My understanding is that you want to run `run_vmaf_training.py` and save the pictures. But why don't you specify the `write_to_dir`, as in:
```
        if save_plot_dir is None:
            DisplayConfig.show()
        else:
            DisplayConfig.show(write_to_dir=save_plot_dir)
```
?

wangwei1237:
> Not sure if I get the motivation for this PR. My understanding is that you want to run `run_vmaf_training.py` and save the pictures. But why don't you specify the `write_to_dir`, as in:
> 
> ```
>         if save_plot_dir is None:
>             DisplayConfig.show()
>         else:
>             DisplayConfig.show(write_to_dir=save_plot_dir)
> ```
> 
> ?

This PR means: we should use `the type of the Matplotlib backends` instead of `write_to_dir` to determine whether to call plt.show().

I want to run run_vmaf_training.py and save the pictures in MacOS. And, I don't set the `--save-plot` for `run_vmaf_training.py`. Then, the code block `DisplayConfig.show()` will execute. 

```
python ../python/vmaf/script/run_vmaf_training.py \
mydata.py \
vmaf_feature_v3.py \
libsvmnusvr_v3.py \
mymodel.json
```

1. In `run_vmaf_training.py`, it use `matplotlib.use('Agg')` to set the matplotlib backends to a `non-interactive backends`. 
2. For `DisplayConfig.show()`, if `if 'write_to_dir' == None`, then `plt.show()` will excute. 
3. For non-GUI backends(non-interactive backends), `plt.show()` will raise an exception, unless running headless (i.e. on Linux with an unset DISPLAY).
4. Thus, I get the following Warning info:
```
/usr/local/lib/python3.8/site-packages/vmaf/config.py:220: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  plt.show()
```

Thus, we should use the type of the Matplotlib backends to determine whether to call plt.show().

li-zhi:
Thanks for the explanation. I see what you mean. But in this case, the problem probably lies in why in MacOS `plt.show()` won't work with `agg` (doesn't seem to be the case for other OS). If we have to, the solution is probably to set the Matplotlib backend based on the OS, instead of changing `DisplayConfig.show()`.

wangwei1237:
> Thanks for the explanation. I see what you mean. But in this case, the problem probably lies in why in MacOS `plt.show()` won't work with `agg` (doesn't seem to be the case for other OS). If we have to, the solution is probably to set the Matplotlib backend based on the OS, instead of changing `DisplayConfig.show()`.

I think you are right, the best solution is to set the Matplotlib backend based on the OS. Thank you for your reply.

PR #854, opened 2021-04-14
Feature/add piecewise linear map to libvmaf

li-zhi:


libvmaf/src/model.h
```diff
@@ -43,6 +43,11 @@ typedef struct {
     VmafDictionary *opts_dict;
 } VmafModelFeature;
 
+typedef struct {
+    double x;
+    double y;
+} Point;
```
kylophone, line 49: Might be a good idea to give this type a Vmaf... prefix. This is to avoid collisions with symbols/types in other libraries.

libvmaf/src/model.h
```diff
@@ -61,6 +66,11 @@ typedef struct VmafModel {
             bool enabled;
             double value;
         } p0, p1, p2;
+        struct {
+            bool enabled;
+            Point *list;
```
kylophone, line 71: Could this be a static allocation instead? i.e.`Point list[3]`?

libvmaf/src/predict.c
```diff
@@ -61,30 +61,143 @@ static int denormalize(const VmafModel *model, double *prediction)
     return 0;
 }
 
+int find_linear_function_parameters(Point p1, Point p2, double *alpha, double *beta) {
```
kylophone, line 64: Use `static` for functions that don't see use outside of this file.

libvmaf/src/predict.c
```diff
@@ -61,30 +61,143 @@ static int denormalize(const VmafModel *model, double *prediction)
...
+    }
+    else {
+        *beta = (p2.y * (p1.y - p1.x)) / (p2.x - p1.x);
+        *alpha = (p1.y - *beta) / p1.x;
+    }
+
+    return 0;
+}
+
+int piecewise_linear_mapping(double x, Point *knots, unsigned n_knots, double *y) {
```
kylophone, line 87: `static`

libvmaf/src/predict.c
```diff
@@ -61,30 +61,143 @@ static int denormalize(const VmafModel *model, double *prediction)
...
     if (!model->score_transform.enabled)
         return 0;
     if (flags & VMAF_MODEL_FLAG_DISABLE_TRANSFORM)
         return 0;
 
-    double value = 0.;
-    double p = *prediction;
+    double y_stage, y_out;
+
+    // polynomial mapping
```
kylophone, line 164: Might be better to have a `VmafTransformMappingType` and use a switch here. Set the default (0) to be `VMAF_TRANSFORM_MAPPING_TYPE_POLYNOMIAL` for backwards compatibility.

libvmaf/src/predict.h
```diff
@@ -22,6 +22,12 @@
 #include "feature/feature_collector.h"
 #include "model.h"
 
+int find_linear_function_parameters(Point p1, Point p2,
+                                    double *alpha, double *beta);
+
+int piecewise_linear_mapping(double x, Point *knots, unsigned n_knots,
+                             double *y);
+
```
kylophone, line 30: Include the `predict.c` file directly in your test, instead of making these functions public.

libvmaf/src/predict.c
```diff
@@ -61,30 +61,143 @@ static int denormalize(const VmafModel *model, double *prediction)
...
+    double slope, offset;
+
+    if (n_knots <= 1)
+        return EINVAL;
+    unsigned n_seg = n_knots - 1;
+
+    *y = 0.0;
+
+    // construct the function
+    for (unsigned idx = 0; idx < n_seg; idx++) {
```
krasuluk, line 98: This loop continues even if y has already been found. We may consider break after *y is assigned.

libvmaf/src/model.h
li-zhi: Could be. But I was following your example of VmafModelFeature *feature. I thought it's better to maintain a consistent style.

libvmaf/src/predict.c
li-zhi: It was used outside in test_predict.c.

libvmaf/src/predict.c
li-zhi: Same as above.

libvmaf/src/predict.c
li-zhi: It is doing cascading of several transforms instead of just one. Probably OK to keep it this way for now (as it is in sync with Python code), and if it grows substantially in the future, we can always refactor to make it better.

libvmaf/src/model.h
li-zhi: Good call.

libvmaf/src/predict.c
li-zhi: Let me keep this for now. Feel free to update the code, but make sure the C and Python version are consistent.

Issue #889, opened 2021-08-04
Problem in reading y4m files


JoshuaEbenezer:
y4m_input.c in libvmaf/tools has an error in reading the headers of y4m files. Line 565 reads the header for 80 characters or until "\n" is reached:
``` 
 /*Read until newline, or 80 cols, whichever happens first.*/
  for(i=0;i<79;i++)
  ```

  The problem is that there is no reason why a y4m header should be less than 80 characters (afaik). I have a y4m file with a long header (more than 80 characters). A quick fix would be increasing this limit to 100.

  ``` 
  char buffer[100];
  int  ret;
  int  i;
  int  xstride;
  /*Read until newline, or 100 cols, whichever happens first.*/
  for(i=0;i<99;i++){ 
    ret=fread(buffer+i,1,1,_fin);
    if(ret<1)return -1;
    if(buffer[i]=='\n')break;
  }
  ```

1480c1:
aom (uses similar code for y4m input) has some different changes

https://aomedia.googlesource.com/aom/+/dfcb5ff4dd7df0c3aa65da1d1b6e697c87292481%5E%21/#F0
https://aomedia.googlesource.com/aom/+/06023aeaff79e6215c69f3b93b9c740f022ee1c4%5E%21/#F0
https://aomedia.googlesource.com/aom/+/cff8ac586150c23fc10712be601af422c41fa8a7%5E%21/#F0
https://aomedia.googlesource.com/aom/+log/refs/heads/master/common/y4minput.c

1480c1:
Also duplicate of https://github.com/Netflix/vmaf/issues/825

JoshuaEbenezer:
Yes. The AOM fix is cleaner. Looks like #825 wasn't resolved so I'm leaving this open until the change is made.

kylophone:
@JoshuaEbenezer Can you please trying reading your y4m with #890 applied? It's a quick fix, but it should work.

JoshuaEbenezer:
yes it works. I didn't do it with #890 , I just defined the header limit as 100 myself. #890 defines the limit as 256, so I did it again with 256 and it works.

PR #890, opened 2021-08-06
tools/y4m_input: increase header buffer size to accomodate certain y4m files.

kylophone:
Lately FFmpeg has been writing y4ms with additional color info in the header. When the old daala y4m parser encounters these files, it overwrites its own buffer and segfaults. This commit is just a quick fix, and simply increases the size of that static buffer.

Should address both #889/#825.


JoshuaEbenezer:
Confirmed that it addresses #889.

PR #894, opened 2021-08-11
Fix a race condition in vmaf_thread_pool_wait.

### LINKED ISSUE #893: Race condition in vmaf / vmafossexec (results in "problem generating pooled VMAF score")
I've observed some random failures / flakiness in the CLI tools when running in testing environments when threads >= 1 (not with threads = 0). Typically the failure looks like:
```
problem generating pooled VMAF score
```
The simplest example that I have to reproduce is as follows. 

First, Create some sample input files (small):
```
k=0; w=32; cnt=2; 
dd bs=$((w*w*3)) count=${cnt} if=/dev/random  of=/tmp/ref.yuv 2>/dev/null
dd bs=$((w*w*3)) count=${cnt} if=/dev/random  of=/tmp/src.yuv 2>/dev/null
```
Run either vmafossexec:
```
vmaf=build/tools/vmafossexec
while ${vmaf} yuv444p $w $w /tmp/ref.yuv /tmp/src.yuv ../model/vmaf_v0.6.1.json --log /tmp/vmaf.log --log-fmt json --thread 1; do k=$((k+1)); echo $k; done; echo "Failed on iteration $k";
```
or vmaf and wait for failure
```
vmaf=build/tools/vmaf

while ${vmaf} -r /tmp/ref.yuv -d /tmp/src.yuv --width $w --height $w --pixel_format 444 -b 8 -m path=../model/vmaf_v0.6.1.json --json --threads 1 -o /tmp/vmaf.log ; do k=$((k+1)); echo $k; done; echo "Failed on iteration $k";
```

Failure may occur randomly after 100s or thousands of iterations:
```
[Trial1]
problem generating pooled VMAF score
Failed on iteration 255

[Trial2]
problem generating pooled VMAF score
Failed on iteration 151

[Trial3]	
problem generating pooled VMAF score
Failed on iteration 25
```

Occurs on any value of threads >= 1 (but doesn't happen with threads=0 as thread pools aren't used there).

I'm fairly sure the termination condition in vmaf_thread_pool_wait is not waiting for the queue to be drained. I'll send a pull request.

nbirkbeck:
Fixes #893. 

There may still be work in the queue but n_working=0 before the vmaf_thread_pool_runner loop fetches another job in the queue.  When calling vmaf_thread_pool_wait (and not stopping), we will still want to drain the queue. So only exit the wait loop when n_working = 0 && head == nullptr

### DIFF
--- a/libvmaf/src/thread_pool.c
+++ b/libvmaf/src/thread_pool.c
@@ -156,7 +156,8 @@ int vmaf_thread_pool_wait(VmafThreadPool *pool)
     if (!pool) return -EINVAL;
 
     pthread_mutex_lock(&(pool->queue.lock));
-    while((!pool->stop && pool->n_working) || (pool->stop && pool->n_threads))
+    while((!pool->stop && (pool->n_working || pool->queue.head)) ||
+          (pool->stop && pool->n_threads))
         pthread_cond_wait(&(pool->working), &(pool->queue.lock));
     pthread_mutex_unlock(&(pool->queue.lock));
     return 0;


nbirkbeck:
No rush. It is possible to avoid the issue while using the non-threaded
code path (using threads=0).

And yes, I did encounter this issue. I noticed some high test flakiness
that hadn't been there in 1.x versions. In one case it was close to 40%
test failure, depending on compilation options, sanitizer options, etc (in
general, rate went down in debug and other sanitizer modes). I'm not
entirely sure why the test environment failures were so much more common
than the shell commands, however. I tested over 20K runs with the patch in
our test environment, and they all succeed.

Has any one else encountered / reported any similar flakiness?

libvmaf/src/thread_pool.c
```diff
@@ -156,7 +156,8 @@ int vmaf_thread_pool_wait(VmafThreadPool *pool)
     if (!pool) return -EINVAL;
 
     pthread_mutex_lock(&(pool->queue.lock));
-    while((!pool->stop && pool->n_working) || (pool->stop && pool->n_threads))
+    while((!pool->stop && (pool->n_working || pool->queue.head)) ||
```
nbirkbeck, line 159: Just noting that it is probably logically sufficient to just check pool->queue.head, although I'll wait for someone closer to the code to comment.

(If there are other potential fixes, I can test in an environment where failures are likley)

Issue #753, opened 2020-12-08
ffmpeg support for v2.0 API and Homebrew


slhck:
I see that FFmpeg does not support the v2.0 API yet. Are there plans to add support for it, perhaps in a backwards-compatible manner? (That is, switching to the old API if libvmaf is <2.0.)

Only if FFmpeg supports it, `libvmaf` can be bumped in Homebrew. If there's no support, bumping in Homebrew will mean that compiling ffmpeg with `--enable-libvmaf` will fail, which would break [custom builds](https://github.com/homebrew-ffmpeg/homebrew-ffmpeg).

So IMO the steps would be:

- [ ] Add support for libvmaf 2.0 in FFmpeg
- [ ] Wait for stable FFmpeg release
- [ ] Bump FFmpeg in Homebrew
- [ ] [Bump libvmaf in Homebrew](https://github.com/Homebrew/homebrew-core/pull/66477)
  - [ ] Fix the applied patch

kylophone:
The current libvmaf ffmpeg works as-is still since the old api call `compute_vmaf()` Is still supported. That being said, I will be working on an updated filter this week to take advantage of v2.0.0 which I will be sending to `ffmpeg-devel`. When that is in-progress/done, I will provide updates via this issue.

Packaging libvmaf now should hopefully be a lot less painful now. We are not providing default model files as part of the installation anymore, since built-in models are available. This should solve problems like this: https://github.com/Netflix/vmaf/issues/324#issuecomment-548508736.

slhck:
Thanks for the heads-up! I didn't see that there was support for the old API call. That means we can safely update libvmaf on Homebrew though. I'll check out the failing PR later.

> Packaging libvmaf now should hopefully be a lot less painful now

Yeah, I saw that – very nice! Good that the models are JSON now, too.

vans163:
To add to this, it seems VMAF changed the build system around and I cannot seem to build a version of FFMPEG with VMAF statically compiled, it keeps trying to load the libvmaf.so.1 when ffmpeg or ffprobe run now.  I even tried the new instructions to build the static lib and indeed i get the static lib and shared built, but FFMPEG keeps trying to want the dynamic one.

```
RUN cd ~/ffmpeg_sources && \
    wget -q https://github.com/Netflix/vmaf/archive/v2.0.0.zip -O vmaf-2.0.0.zip && \
    unzip -qq vmaf-2.0.0.zip && cd vmaf-2.0.0 && \
[...]

PR #897, opened 2021-08-13
Fix some backward compatibility issue

quink-black:


kylophone:
Can you please give some context for these changes?

quink-black:
> Can you please give some context for these changes?

It's in the commit message:

1. model: fix misleading log

    Deprecated means it works but will be removed. The support of pkl
    model has been removed. The log message and log level make user
    wondering why it doesn't work.

2.     compute_vmaf: use default format when log_path is specified but log_fmt is missing

    For backward compatibility. libvmaf has default log_fmt is not specified in previous version.

3.    compute_vmaf: add sub to log_fmt_map

Support a new format through old API.

quink-black:
By the way, these issues were found when using libvmaf via FFmpeg. I'm wondering if there is some progress in

https://github.com/Netflix/vmaf/issues/753

libvmaf/src/model.c
```diff
@@ -137,12 +137,13 @@ int vmaf_model_load_from_path(VmafModel **model, VmafModelConfig *cfg,
...
-        vmaf_log(VMAF_LOG_LEVEL_WARNING,
+        vmaf_log(VMAF_LOG_LEVEL_ERROR,
                  "could not read model from path: \"%s\"\n", path);
         char *ext = strrchr(path, '.');
         if (ext && !strcmp(ext, ".pkl")) {
-            vmaf_log(VMAF_LOG_LEVEL_WARNING,
-                     "pkl model files have been deprecated, use json\n");
+            vmaf_log(
+                VMAF_LOG_LEVEL_ERROR,
+                "support for pkl model files have been removed, use json\n");
```
slhck, line 146: Minor nit: "support … **has** been removed, …"

libvmaf/src/model.c
```diff
@@ -282,12 +283,13 @@ int vmaf_model_collection_load_from_path(VmafModel **model,
...
-        vmaf_log(VMAF_LOG_LEVEL_WARNING,
+        vmaf_log(VMAF_LOG_LEVEL_ERROR,
                  "could not read model collection from path: \"%s\"\n", path);
         char *ext = strrchr(path, '.');
         if (ext && !strcmp(ext, ".pkl")) {
-            vmaf_log(VMAF_LOG_LEVEL_WARNING,
-                     "pkl model files have been deprecated, use json\n");
+            vmaf_log(
+                VMAF_LOG_LEVEL_ERROR,
+                "support for pkl model files have been removed, use json\n");
```
slhck, line 292: Same comment as above; use "has" instead of "have"

slhck:
This is indeed something rather confusing, also seeing that some precompiled ffmpeg binaries eventually switched back to libvmaf 1.x because end users had issues with the 2.x version and the model files. (I don't know what these issues were. I'm just paraphrasing a maintainer.)

PR #908, opened 2021-10-06
Ffmpeg fix

sb1487:
test


kylophone:
The diff doesn't make sense to me. Can you give some context please?

Issue #863, opened 2021-05-13
PSNR-HVS and CIEDE2000 need a "enable_chroma" option


HolyWu:
When processing yuv400p single channel video, PSNR-HVS generated NaN values like the following. For CIEDE2000, the process just crashed.

```
<VMAF version="2.1.1">
  <params qualityWidth="640" qualityHeight="360" />
  <fyi fps="10.41" />
  <frames>
    <frame frameNum="0" psnr_y="27.435467" integer_motion2="0.000000" integer_motion="0.000000" psnr_hvs_y="30.252694" psnr_hvs_cb="-nan(ind)" psnr_hvs_cr="-nan(ind)" psnr_hvs="-nan(ind)" integer_adm2="0.826119" integer_adm_scale0="0.673305" integer_adm_scale1="0.675514" integer_adm_scale2="0.877188" integer_adm_scale3="0.959387" integer_vif_scale0="0.336571" integer_vif_scale1="0.904926" integer_vif_scale2="0.966646" integer_vif_scale3="0.987702" float_ssim="0.920977" float_ms_ssim="0.982248" vmaf="59.528155" />
    <frame frameNum="1" psnr_y="27.147223" integer_motion2="3.748950" integer_motion="3.937943" psnr_hvs_y="30.074174" psnr_hvs_cb="-nan(ind)" psnr_hvs_cr="-nan(ind)" psnr_hvs="-nan(ind)" integer_adm2="0.829521" integer_adm_scale0="0.662171" integer_adm_scale1="0.681730" integer_adm_scale2="0.873541" integer_adm_scale3="0.960149" integer_vif_scale0="0.327428" integer_vif_scale1="0.903528" integer_vif_scale2="0.967052" integer_vif_scale3="0.988674" float_ssim="0.917704" float_ms_ssim="0.981696" vmaf="64.383290" />
[...]

PR #909, opened 2021-10-07
Feature/add cambi

li-zhi:


kylophone:
Since we're adding `VMAF_PIX_FMT_YUV400P` to the public API, we'll need to bump the minor version whenever we cut the new release. Also, #863 will need to be addressed before the next release as well.

kylophone:
Tested with UBSAN and ASAN, no problems. I noticed that `convert_8b_luma_to_10b` is unused:

```sh
../src/feature/cambi.c:347:13: warning: unused function 'convert_8b_luma_to_10b' [-Wunused-function]  
```

li-zhi:
> Tested with UBSAN and ASAN, no problems. I noticed that `convert_8b_luma_to_10b` is unused:
> 
> ```shell
> ../src/feature/cambi.c:347:13: warning: unused function 'convert_8b_luma_to_10b' [-Wunused-function]  
> ```

Removed convert_8b_luma_to_10b.

li-zhi:
> Since we're adding `VMAF_PIX_FMT_YUV400P` to the public API, we'll need to bump the minor version whenever we cut the new release. Also, #863 will need to be addressed before the next release as well.

@kylophone Adding `VMAF_PIX_FMT_YUV400P` seems fine as @joelsole uses this format only for an internal initialization in cambi. The public API only supports 420/422/444 still. See: https://github.com/Netflix/vmaf/blob/master/libvmaf/tools/cli_parse.c#L127-L144
So we are not breaking anything.

#863 is different, as @HolyWu seems trying to directly access the methods underneath. We may later consider generalizing the API to support 400 (or "gray" in ffmpeg lingo) but it should not block CAMBI release.

kylophone:
> > Since we're adding `VMAF_PIX_FMT_YUV400P` to the public API, we'll need to bump the minor version whenever we cut the new release. Also, #863 will need to be addressed before the next release as well.
> 
> @kylophone Adding `VMAF_PIX_FMT_YUV400P` seems fine as @joelsole uses this format only for an internal initialization in cambi. The public API only supports 420/422/444 still. See: https://github.com/Netflix/vmaf/blob/master/libvmaf/tools/cli_parse.c#L127-L144 So we are not breaking anything.

I am talking about the libvmaf C API. This is an addition to a public enum. It is a backwards compatible change, but not a forwards compatible one. Therefore minor library version should be bumped.

https://github.com/Netflix/vmaf/blob/f5e984375925d88edb8fb613db7496be642f8450/libvmaf/include/libvmaf/picture.h#L33

kylophone:
> > When you squash this, make sure you give the authors attribution in the commit.
> 
> @kylophone I don't plan to squash, but I've updated the commit message to indicate the authors of the code.

You can add multiple authors to the commit if you'd like.

PR #920, opened 2021-10-25
Added support for external thread pool

foosoftsrl:


li-zhi:
Sorry, overlooked this. Is there still an interest to merge?

My understanding is that the code change is just a refactoring of existing code to make an API for your external implementation, which is not made public. Is this right?

kylophone:
> Sorry, overlooked this. Is there still an interest to merge?
> 
> My understanding is that the code change is just a refactoring of existing code to make an API for your external implementation, which is not made public. Is this right?

No this is a public API change. If we want to have something similar, I suggest a new API for this use case specifically. I would prefer a call that took multiple distorted pictures vs support for an external thread pool. I would also be nice to see some throughput (FPS) numbers with each solution. For example, does running each of your `VmafContext`s single threaded accomplish a similar throughput to the external threadpool. How does it affect latency, how does it affect overall CPU usage?

foosoftsrl:
Hi both, 

Kile is right, an API change is required (BTW, this could be easily avoidable if there was a constructor / destructor for VmafConfiguration).
So... client code would not change, but a compilation *would be required* should the pull request be accepted as-is, provided that clients do a zero memset of VMafConfiguration.

Kile is also right that it is not definitely sure that running many VMAF estimations at a time is a winner: the enormous memory footprint could make it worth it to do decode many times and do one (or two, I'd say) VMAFs computation at a time.

But.. I believe that some of the modifications I made in another branch are worth a look:

1) extensive C++ usage except for feature extractors and libvmaf API (code is more compact, portable threads are available)
2) Visual Studio portability (with Microsoft compiler!) 
3) cmake build
4) pipelined VIF filtering and feature extraction --> enormous memory saving
5) partially rewritten AVX2 VIF filtering code 
6) significant performance improvement. I'd say... more than 50% faster, at least on AVX2. Might be even more on AVX512, as VIF feature extraction can be performed on AVX512 (R/W scatter / gather for logarithms + easier integer / double conversion)

Point 4) really makes a difference, performance-wise, and I'm under the impression that preprocessing reference frames in order to avoid some computations is not a great idea as I thought it was, especially due to how VIF is computed.

Is the external thread pool still a good idea? I think so, but again Kile is right: a test application is missing (I'll do it)

So, I definitely took a different path from current libvmaf development.

If you're interested, I can and will contribute to libvmaf, but there are a few choices (essentially C++), which may be unacceptable for you at the framework level.

The modification to VIF feature extraction are much easier to integrate.
Regarding this last point, I think that small changes to VIF integer based feature extraction might make AVX2 / AVX512 implementations faster and easier to write.

kylophone:
> 1. extensive C++ usage except for feature extractors and libvmaf API (code is more compact, portable threads are available)

There are bits of libvmaf which are C++ (libsvm), but all new library code should be C11. 

> 2. Visual Studio portability (with Microsoft compiler!)
> 3. cmake build

Not interested in `cmake`. Is there something you need which is missing with `meson`?

> 4. pipelined VIF filtering and feature extraction --> enormous memory saving
> 5. partially rewritten AVX2 VIF filtering code
> 6. significant performance improvement. I'd say... more than 50% faster, at least on AVX2. Might be even more on AVX512, as VIF feature extraction can be performed on AVX512 (R/W scatter / gather for logarithms + easier integer / double conversion)

I am very interested to see what you've come up with here. Were your modifications done on the integer version of VIF?

foosoftsrl:
> > 1. extensive C++ usage except for feature extractors and libvmaf API (code is more compact, portable threads are available)
> 
> There are bits of libvmaf which are C++ (libsvm), but all new library code should be C11.

If you ask my opinion... C is good for ABI stability, and that's it. Too error prone. But it is just a matter of taste!
This will make a full merge of our code bases not feasible. 
But again, feature extraction merge is still dead easy: on the feature extractor  side I did not change anything on purpose.

> 
> > 1. Visual Studio portability (with Microsoft compiler!)
> > 2. cmake build
> 
> Not interested in `cmake`. Is there something you need which is missing with `meson`?

First Visual Studio integration was via cmake and WSL. I'm thinking of integrating vmaf in a bigger closed source linux targeted project where we're using cmake. Nothing really relevant.

> 
> > 1. pipelined VIF filtering and feature extraction --> enormous memory saving
> > 2. partially rewritten AVX2 VIF filtering code
> > 3. significant performance improvement. I'd say... more than 50% faster, at least on AVX2. Might be even more on AVX512, as VIF feature extraction can be performed on AVX512 (R/W scatter / gather for logarithms + easier integer / double conversion)
> 
> I am very interested to see what you've come up with here. Were your modifications done on the integer version of VIF?

Here our current vif integer implementation

https://github.com/foosoftsrl/vmaf/blob/rewrite/libvmaf/src/feature/integer_vif.c

We filter and compute vmaf in vif_statistic_8. 
As you see,, vif metric accumulation is performed line by line, and the filtered image is not stored.
I tidied up a bit the code of VIF feature extraction, more integer usage, moved stuff inside if's... results should be unchanged

```
...
void vif_statistic_8(struct VifState* s, float* num, float* den, unsigned w, unsigned h) {
...
...


There might be additional memory saving using partially overwrapping between stages mu1 and mu2 blocks

The avx2 implementation is here at vif_statistic_8_avx2. 

https://github.com/foosoftsrl/vmaf/blob/rewrite/libvmaf/src/feature/x86/vif_avx2.c

(Note: I did not care about non multiple of 16 width / height. fallback to pure C is ok for me, in that case)

Here an experimental AVX512 vif implementation at vif_statistic_avx512 (old non pipelined API)

https://github.com/foosoftsrl/vmaf/blob/rewrite/libvmaf/src/feature/x86/vif_avx512.c

Ok, cleanup is needed, but the "rewrite" branch is where we are.

Plans are:
1) feedback about results with concurrent single reference VMAF tasks 
2) do some more optimization (target was 2X, I hope it's feasible)
3) contribute back if possible.  We will stay anyway on our fork for the time being.

**Comment by kylophone** [comment]
> > > 1. pipelined VIF filtering and feature extraction --> enormous memory saving
> > > 2. partially rewritten AVX2 VIF filtering code
> > > 3. significant performance improvement. I'd say... more than 50% faster, at least on AVX2. Might be even more on AVX512, as VIF feature extraction can be performed on AVX512 (R/W scatter / gather for logarithms + easier integer / double conversion)
> > 
> > 
> > I am very interested to see what you've come up with here. Were your modifications done on the integer version of VIF?
> 
> Here our current vif integer implementation
> 
> https://github.com/foosoftsrl/vmaf/blob/rewrite/libvmaf/src/feature/integer_vif.c
> 
> We filter and compute vmaf in vif_statistic_8. As you see,, vif metric accumulation is performed line by line, and the filtered image is not stored. I tidied up a bit the code of VIF feature extraction, more integer usage, moved stuff inside if's... results should be unchanged
> 
> ```
> ...
> void vif_statistic_8(struct VifState* s, float* num, float* den, unsigned w, unsigned h) {
> ...
> ...
> ```
> 
> There might be additional memory saving using partially overwrapping between stages mu1 and mu2 blocks
> 
> The avx2 implementation is here at vif_statistic_8_avx2.
> 
> https://github.com/foosoftsrl/vmaf/blob/rewrite/libvmaf/src/feature/x86/vif_avx2.c
> 
> (Note: I did not care about non multiple of 16 width / height. fallback to pure C is ok for me, in that case)
> 
> Here an experimental AVX512 vif implementation at vif_statistic_avx512 (old non pipelined API)
> 
> https://github.com/foosoftsrl/vmaf/blob/rewrite/libvmaf/src/feature/x86/vif_avx512.c
> 
> Ok, cleanup is needed, but the "rewrite" branch is where we are.

The VIF pipelining work looks very promising. I would like to try it out, but I was wondering if I should I wait for your cleanup, or do you think I should jump in now? Do you think we can track this work in a separate PR? We could also discuss the threadpool and multithreading optimizations separately.

> 
> Plans are:
> 
> 1. feedback about results with concurrent single reference VMAF tasks
> 2. do some more optimization (target was 2X, I hope it's feasible)
> 3. contribute back if possible.  We will stay anyway on our fork for the time being.

**Comment by foosoftsrl** [comment]
Hi Kile,
thanks for the interest.

First of all, I sadly confirm that the shared thread pool is much less interesting than I hoped, and it can make a difference only in a low-memory scenario.
In the current implementations contexts and threads are tied, and to use n CPUs at least n contexts are needed.
A decoupled thread pool implementation can execute in parallel n * m tasks, where m is the number of feature extractors.

Regarding the astonishing :) performance improvements, I talked about, they are there, but there is something I can't understand.

This is the original implementation with 16 threads on a 16 core machine:

```
[luca@c10n2 ~]$ time ./vmaf_org -q --threads 16 -b 8 -p 420 -w 1280 -h 720 -r testsrc.yuv -d testsrc.dec.yuv
...
real    0m1.996s
user    0m25.765s
sys     0m0.678s
```

And this is the "optimized" app:
```
[luca@c10n2 ~]$ time ./vmaf_multi -q --threads 16 -w 1280 --height 720 -r testsrc.yuv -d testsrc.dec.yuv
vmaf: 97.534277

real    0m1.196s
user    0m18.070s
sys     0m0.303s
```

The performance improvement is not bad... 1.2 seconds vs 2 seconds, 60% faster. But it is mostly due to better CPU usage, as it is evident from "user" timing, 

Working with a single thread... Netflix's implementation takes 9.0 seconds, ours 7.5. Just 20% faster. I'd say that 20% is VIF filtering and the other 40%.. no idea... I can only guess that in reriting the code I removed a bottleneck, possibly in thread pool code (which has been rewritten from scratch) without even noticing it.

Back to your suggestion of a PR limited to optimization, it's ok. We can close this PR.

kylophone:
> Working with a single thread... Netflix's implementation takes 9.0 seconds, ours 7.5. Just 20% faster. I'd say that 20% is VIF filtering and the other 40%.. no idea... I can only guess that in reriting the code I removed a bottleneck, possibly in thread pool code (which has been rewritten from scratch) without even noticing it.

Yes, there is a bottleneck when running multithreaded and it has to do with the shared resource `VmafFeatureCollector`. More than likely your thread pool implementation has little to do with it, since it doesn't change this fact. I have [known about this for some time](https://docs.google.com/spreadsheets/d/1NZ-PZq9rbHLcB-R53lulxQCbaakc3B9T9wX_3Y5I2AU/edit#gid=0), but haven't gone after it. Mostly, I have been focused on CPU reductions and overall latency has not been a motivator. There are a number of ways we could go after a fix for this including per feature extractor caches. If we did this, there would be nearly no contention and adding threads should scale linearly. If it is important to you, we can make plans to go after something like this.

In any case, if your VIF code is accurate that is a huge accomplishment. Please send the PR when it is ready. Let me know if you run into any problems or have any questions. I am happy to help.

foosoftsrl:
Closing this PR, opened PR 943

PR #937, opened 2021-11-12
CAMBI: Move mallocs to init to avoid allocating repeatedly

nilfm:
Instead of allocating and freeing memory on each call to `filter_mode`, allocate it once in `init` and free it in `close`. The results are identical on 10 benchmark videos, and the execution time is equal or very slightly faster. The main advantage is centralizing all memory allocations inside `init`.

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -82,6 +82,8 @@ typedef struct CambiState {
     float *c_values;
     uint16_t *c_values_histograms;
     uint32_t *mask_dp;
+    uint8_t *filter_mode_histogram;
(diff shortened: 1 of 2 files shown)


kylophone:
Looks OK, one question: do you need to zero out the buffers each time, or do they always get completely overwritten?

nilfm:
Both the `buffer` and `histogram` arrays are overwritten in the current code and don't need to be zeroed out.

By the way, reading the Travis log, I don't think the failure is related, is that right?

kylophone:
> By the way, reading the Travis log, I don't think the failure is related, is that right?

Nope. Looks like a failed download in one of the tests.

PR #938, opened 2021-11-12
libvmaf: add vmaf_feature_collector_append_formatted

kylophone:
This PR adds a function for writing formatted feature names. See test for example usage.


kylophone:
`vasprintf ` may not be portable. Double check before merging.

PR #941, opened 2021-11-17
Add max_contrast parameter

joelsole:
Add max_contrast parameter to account for larger contrast banding than 4 luma levels @ 10b  (e.g., for non-compression banding generated by bit-depth reduction).

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -36,6 +36,9 @@
 /* Visibilty threshold for luminance ΔL < tvi_threshold*L_mean for BT.1886 */
 #define DEFAULT_CAMBI_TVI (0.019)
 
+/* Max log contrast luma levels */
+#define DEFAULT_MAX_LOG_CONTRAST (2)
+
 #define CAMBI_MIN_WIDTH (320)
 #define CAMBI_MAX_WIDTH (4096)
 #define CAMBI_4K_WIDTH (3840)
@@ -44,13 +47,13 @@
 #define NUM_SCALES 5
 static const int g_scale_weights[NUM_SCALES] = {16, 8, 4, 2, 1};
 
-#define NUM_DIFFS 4
-static const int g_diffs_to_consider[NUM_DIFFS] = {1, 2, 3, 4};
-static const int g_diffs_weights[NUM_DIFFS] = {1, 2, 3, 4};
+/* Suprathreshold contrast response */
+static const int g_contrast_weights[32] = {1, 2, 3, 4, 4, 5, 5, 6, 6, 6, 6, 7, 7, 7, 7, 8,
+                                           8, 8, 8, 8, 8, 8, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9};
 
-#define NUM_ALL_DIFFS (2 * NUM_DIFFS + 1)
-static const int g_all_diffs[NUM_ALL_DIFFS] = {-4, -3, -2, -1, 0, 1, 2, 3, 4};
-static const uint16_t g_c_value_histogram_offset = 4; // = -g_all_diffs[0]
+static uint16_t *g_diffs_to_consider;
+static int *g_diffs_weights;
+static int *g_all_diffs;
 
 #define MAX(x, y) (((x) > (y)) ? (x) : (y))
 #define MIN(x, y) (((x) < (y)) ? (x) : (y))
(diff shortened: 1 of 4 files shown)


libvmaf/src/feature/cambi.c
```diff
@@ -132,6 +136,16 @@ static const VmafOption options[] = {
         .min = 0.0001,
         .max = 1.0,
     },
+    {
+        .name = "max_log_contrast",
+        .help = "Max contrast luma level at 10-bits, set to  2^max_log_contrast \
+                 \t 0 to 5: Default 2 is recommended for banding from compression",
```
kylophone, line 142: You could do a multiline string like this: https://github.com/Netflix/vmaf/blob/master/libvmaf/src/feature/integer_adm.c#L52-L53

libvmaf/src/feature/cambi.c
```diff
@@ -242,6 +256,22 @@ static FORCE_INLINE inline void adjust_window_size(uint16_t *window_size, unsign
     (*window_size) = ((*window_size) * input_width) / CAMBI_4K_WIDTH;
 }
 
+void  set_contrast_arrays(const uint16_t num_diffs, uint16_t **diffs_to_consider,
+                          int **diffs_weights, int **all_diffs)
+{
+    *diffs_to_consider = aligned_malloc(ALIGN_CEIL(sizeof(uint16_t)) * num_diffs, 16);
+    *diffs_weights = aligned_malloc(ALIGN_CEIL(sizeof(int)) * num_diffs, 32);
+    *all_diffs = aligned_malloc(ALIGN_CEIL(sizeof(int)) * (2*num_diffs+1), 32);
```
kylophone, line 264: These are unchecked mallocs.

libvmaf/src/feature/cambi.c
```diff
@@ -720,8 +759,10 @@ static FORCE_INLINE inline double weight_scores_per_scale(double *scores_per_sca
 }
 
 static int cambi_score(VmafPicture *pics, uint32_t *mask_dp, uint16_t window_size, double topk,
```
kylophone, line 761: Fine as is, but maybe you want to just take `s` as a parameter here.

libvmaf/src/feature/cambi.c
joelsole: In the code, some are checked and some unchecked.  Is the preference to use the pattern below for all the aligned mallocs in cambi?

https://github.com/Netflix/vmaf/blob/d70c679c8278d6da054e86975f971e11b8f5050d/libvmaf/src/feature/vif.c#L323-L328

libvmaf/src/feature/cambi.c
joelsole: got it

libvmaf/src/feature/cambi.c
```diff
@@ -36,6 +36,9 @@
 /* Visibilty threshold for luminance ΔL < tvi_threshold*L_mean for BT.1886 */
 #define DEFAULT_CAMBI_TVI (0.019)
 
+/* Max log contrast luma levels */
+#define DEFAULT_MAX_LOG_CONTRAST (2)
```
li-zhi, line 40: Can we add one test case to cover the case when max_log_contrast is not the default?

libvmaf/src/feature/cambi.c
kylophone: That code is from the older float vif. It should be updated, but it is basically deprecated at this point. Better to follow this pattern: https://github.com/Netflix/vmaf/blob/master/libvmaf/src/feature/integer_vif.c#L585-L586

Easiest thing to do is do all your allocs upfront in `.init()` and fail right away if there are problems.

libvmaf/src/feature/cambi.c
joelsole: added

libvmaf/src/feature/cambi.c
joelsole: done

libvmaf/src/feature/cambi.c
joelsole: Yes - @nilfm did something on those lines here: https://github.com/Netflix/vmaf/commit/50d64f97d612fd7f1f01459b2ba3668d717b8e2c
Can be added then.

libvmaf/src/feature/cambi.c
```diff
@@ -132,6 +136,16 @@ static const VmafOption options[] = {
         .min = 0.0001,
         .max = 1.0,
     },
+    {
+        .name = "max_log_contrast",
+        .help = "Max contrast luma level at 10-bits, set to  2^max_log_contrast. "
```
nilfm, line 141: I'd say this phrasing is a bit confusing with the comma, maybe it reads better as "Max contrast luma level at 10-bits is set to 2^max_log_contrast."

li-zhi: Can we re-use what you had in README?
```
Maximum contrast in log luma level (2^max_log_contrast) at 10-bits. Default 2 is equivalent to 4 luma levels at 10-bit and 1 luma level at 8-bit.
```
Better explicit than sorry.

li-zhi (approved):

libvmaf/src/feature/cambi.c
joelsole: done. Thanks.

PR #942, opened 2021-11-18
libvmaf: add vmaf_feature_name_from_options

kylophone:
This PR adds a helper function for building parameterized feature names: `vmaf_feature_name_from_options()`. Parameterized feature names are constructed using a base feature name and a set of `VmafOptions` and a state struct.  Suffixes are added to the base feature name only when an option has a non-default value in the form of `_key_val`. The option name (`VmafOption.name`) is used for `key`, but an alias (`VmafOption.alias`) may also be used. This is a variadic function so feature names can be parametrized using any number of options. This is an allocating function, so normal usage inside a feature extractor would be to generate/allocate once in `.init()` and then free in `.close()`.

For example usage and output, see `test_feature_name_from_options()`.

### DIFF
--- a/libvmaf/test/test_feature.c
+++ b/libvmaf/test/test_feature.c
@@ -157,8 +157,9 @@ static char *test_feature_name_from_options()
                                        &s5.opt_double, &s5.opt_bool, &s5.opt_int);
 
     mu_assert("feature_name should have a suffix with aliases and values, "
-              "ordering should follow the ordering of variadac params",
-              !strcmp(feature_name5, "feature_name_opt_double_alias_4.14_opt_bool_1_opt_int_alias_201"));
+              "ordering should not follow the ordering of variadac params,"
+              "rather it should follow the order of options",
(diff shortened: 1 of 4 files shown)


libvmaf/test/test_feature.c
```diff
@@ -49,8 +50,137 @@ static char *test_feature_name()
...
+    free(feature_name4);
+
+    TestState s5 = s4;
+
+    char *feature_name5 =
+        vmaf_feature_name_from_options("feature_name", &options, &s5, 3,
+                                       &s5.opt_double, &s5.opt_bool, &s5.opt_int);
+
+    mu_assert("feature_name should have a suffix with aliases and values, "
+              "ordering should follow the ordering of variadac params",
```
li-zhi, line 160: "ordering should follow the ordering of variadac params" ==> this could be a source of bug. We would want to use the feature_name as a unique identifier, but in case a user calls vmaf_feature_name_from_options twice with different options ordering, it will result in two feature names, although the feature numerical results are the same. Two possible mitigations: either 1) check to see if the input options are sorted alphabetically; if not, return an error code; 2) sort the input options before constructing the feature name string.

PR #943, opened 2021-11-18
- Pipelined vif feature computation: memory requirements reduced and …

foosoftsrl:
…more performance

### DIFF
--- a/libvmaf/src/feature/feature_collector.c
+++ b/libvmaf/src/feature/feature_collector.c
@@ -44,6 +44,16 @@ static int aggregate_vector_append(AggregateVector *aggregate_vector,
 {
     if (!aggregate_vector) return -EINVAL;
 
+    for (unsigned i = 0; i < aggregate_vector->cnt; i++) {
+        if (!strcmp(feature_name, aggregate_vector->metric[i].name)) {
+            if (aggregate_vector->metric[i].value == score) {
+                return 0;
+            } else {
+                return -EINVAL;
+            }
+        }
+    }
+
     const unsigned cnt = aggregate_vector->cnt;
     if (cnt >= aggregate_vector->capacity) {
         size_t initial_size =
--- a/libvmaf/src/feature/integer_psnr.c
+++ b/libvmaf/src/feature/integer_psnr.c
@@ -200,21 +200,25 @@ static int extract(VmafFeatureExtractor *fex,
     }
 }
 
-static int close(VmafFeatureExtractor *fex)
+static int flush(VmafFeatureExtractor *fex,
+                 VmafFeatureCollector *feature_collector)
 {
     PsnrState *s = fex->priv;
     const char *apsnr_name[3] = { "apsnr_y", "apsnr_cb", "apsnr_cr" };
 
+    int err = 0;
     if (s->enable_apsnr) {
         for (unsigned i = 0; i < 3; i++) {
-            double apsnr =
-                log10(s->peak * s->peak) +
-                log10(s->apsnr.n_pixels[i]) -
-                log10(s->apsnr.sse[i]);
-            (void) apsnr;
+            double apsnr = 10 * (log10(s->peak * s->peak) +
+                                 log10(s->apsnr.n_pixels[i]) -
+                                 log10(s->apsnr.sse[i]));
+            err |=
+                vmaf_feature_collector_set_aggregate(feature_collector,
+                                                     apsnr_name[i], apsnr);
         }
     }
-    return 0;
+
+    return (err < 0) ? err : !err;
 }
 
 static const char *provided_features[] = {
@@ -227,7 +231,8 @@ VmafFeatureExtractor vmaf_fex_psnr = {
     .options = options,
     .init = init,
     .extract = extract,
-    .close = close,
+    .flush = flush,
     .priv_size = sizeof(PsnrState),
     .provided_features = provided_features,
+    .flags = VMAF_FEATURE_EXTRACTOR_TEMPORAL,
 };
--- a/libvmaf/test/test_feature_extractor.c
+++ b/libvmaf/test/test_feature_extractor.c
@@ -52,15 +52,16 @@ static char *test_feature_extractor_context_pool()
     err = vmaf_fex_ctx_pool_create(&pool, n_threads);
     mu_assert("problem during vmaf_fex_ctx_pool_create", !err);
 
-    VmafFeatureExtractor *fex = vmaf_get_feature_extractor_by_name("psnr");
+    VmafFeatureExtractor *fex =
+        vmaf_get_feature_extractor_by_name("float_ssim");
     mu_assert("problem during vmaf_get_feature_extractor_by_name", fex);
 
     VmafFeatureExtractorContext *fex_ctx[n_threads];
     for (unsigned i = 0; i < n_threads; i++) {
         err = vmaf_fex_ctx_pool_aquire(pool, fex, NULL, &fex_ctx[i]);
         mu_assert("problem during vmaf_fex_ctx_pool_aquire", !err);
-        mu_assert("fex_ctx[i] should be ssim feature extractor",
-                  !strcmp(fex_ctx[i]->fex->name, "psnr"));
+        mu_assert("fex_ctx[i] should be float_ssim feature extractor",
+                  !strcmp(fex_ctx[i]->fex->name, "float_ssim"));
     }
 
     for (unsigned i = 0; i < n_threads; i++) {


foosoftsrl:
First set of optimizations with the 20% performance promised here: https://github.com/Netflix/vmaf/pull/920

Code still needs some cleanup

foosoftsrl:
PS. 10 bit VMAF not tested. this PR needs a good review

kylophone:
Looks like a large part of your diff is just re-indentation, which makes it super hard to see what exactly has changed and what is superfluous. Any way you can clean up the PR so it is more readable/reviewable? Which functions have changed?

foosoftsrl:
Hi Kyle,
yes there is somethng superflous, and yes there has been some reindentation.
But it's not about reindentation: lots of things have changed.

First of all there are now two different sets of function:
- vif_statistic_{bpp}_{optimization}
- vif_subsample_rd_{bpp}_{optimization}

For AVX2: 

vif_statistic_8_avx2 is the old filter_8 with brand new horizontal filter, vertical filter brand new, and brand new vif computation
vif_statistic_16_avx2 is the old filter_16 with brand new vertical filter, and brand new vif computation
vif_subsample_rd_8_avx2 is more or less identical to the previous subsampling code, but horizontal filter is made with _mm256_alignr_epi8  and downsampling is performed in-loop
vif_subsample_rd_16_avx2 is more or less identical to the previous subsampling code, with vertical downsampling in-loop.

IMHO, it's not an incremental evolution, and trying to match it line by line does not make much sense.

Right now... if I were you... I'd test the code for regressions on results and perhaps wait for cleanup on our side for an actual merge (for example... you could ask for regression tests)

But as some point in time it should be old implementation out, new implementation in.

kylophone:
> Right now... if I were you... I'd test the code for regressions on results and perhaps wait for cleanup on our side for an actual merge (for example... you could ask for regression tests)

Have a look at the failing regression tests here: https://github.com/Netflix/vmaf/runs/4248309754?check_suite_focus=true

libvmaf/src/feature/integer_vif.c
```diff
@@ -37,758 +53,1033 @@
...
+	int64_t num_val, den_val;
+	int64_t accum_x = 0, accum_x2 = 0;
+	int64_t accum_num_log = 0.0;
+	int64_t accum_den_log = 0.0;
+	int64_t accum_num_non_log = 0;
+	int64_t accum_den_non_log = 0;
+	uint32_t* xx_filt = buf.ref_sq;
+	uint32_t* yy_filt = buf.dis_sq;
+	uint32_t* xy_filt = buf.ref_dis;
+	static const int32_t sigma_nsq = 65536 << 1;
```
igv, line 677: Why is it identical to sigma_nsq from vif_statistic_8? For 10 bit it should be 32 (2 * 4^2), for 12 bit - 512 (2 * 16^2).

foosoftsrl: Hi igv,
really thank you for pointing out.
I did not consider 10/12 bit at all, and this must simply be a bug.
I'm really sorry for the hurried pull request, but for me this is just a proof of concept.
Probably "pull request", "proof of concept", and "regression tests broken" are not good friends.
I'm very new to open source contributions and collaboration with unknown people.
I'll fix that shortly.

foosoftsrl: Hi igv,
I looked at integer_vif implementation, and I really can't get the point:

This is the original implementation, and... can't see anything related to 10 or 12 bits.

```
static void vif_statistic(VifBuffer buf, float *num, float *den,
                          unsigned w, unsigned h, uint16_t *log2_table,
                          double vif_enhn_gain_limit)
{
    uint32_t *xx_filt = buf.ref_sq;
    uint32_t *yy_filt = buf.dis_sq;
    uint32_t *xy_filt = buf.ref_dis;

    //float equivalent of 2. (2 * 65536)
    static const int32_t sigma_nsq = 65536 << 1;
...
```
 
Can you help me understand?

foosoftsrl:
Hi Kyle,
the regression was with non multiple of 128 widths, fixed. Committed to the PR branch.

Ran the regression test on WSL, there is one error which might be related to this PR... test_run_integer_vif_fextractor_with_debug

It might be related to the WSL environment!!!!

Don't know how to schedule a regression test.... on our repo tasks are disabled, on your repo I don't know if tasks are schedule automatically or whatever

There is an 

Let me know what I can do next.

libvmaf/src/feature/integer_vif.c
igv: Was it even compatible with higher bit depths? Original VIF implementation certainly isn't. I'm assuming `vif_statistic_16` means for 16 bit inputs (10-16), right?

foosoftsrl: Hi igv,
vif_statistic_16 is actually both for > 8 bit and for second stage and so on.

I really did not spend enough time with > 8 bit implementation, I'm still at an early stage.
So... there might be bugs, I was superclear about that! 
Let's see what the regression tests say.

But... what do you mean by "Original VIF implementation certainly isn't"? You mean that the head of master branch is not ok?

igv: I'm not very familiar with VMAF code base, just trying to make sure that this new version of VIF is optimally optimized and without bugs.
That sigma_nsq constant (variance of invisible noise) came from matlab version of VIFp, which is for 8bit images only. If your `vif_statistic_16` actually processes >=10bit frames (AFAIK currently VMAF converts everything to 8bit internally) than you need to increase that constant properly (and optionally epsilon).

foosoftsrl:
Hi Kyle,
now this is weird... early fail due to cpu flags not zero at start????

kylophone:
> Hi Kyle, there are very early failures in the regression tests, something related to cpu initialization flags. This is not our stuff, correct?

https://github.com/Netflix/vmaf/pull/943/files#diff-3d6a0aace6c3eef3e98ea6be1543a1c8a6cbe3b15b09bdf620d81829d883a897L22

Revert your changes to 'cpu.c' and the test should pass again.

foosoftsrl:
git is beyond me, seriously.
I deleted the branch.

What I want to do is:

- open a branch which does not start from my master
- commit 4 (four) files
- open a pull request

Is there an easy way to do this?

I'm even thinking of having two forks, one for PRs, and one for my projects.

I'm in serious need of help.

igv:
My rule is to never use `git merge` (I noticed you merged some other branch into vif_optimization branch), instead use `git rebase` and `git pull --rebase`

slhck:
If you have all your changes in a dedicated branch, you can keep one fork of the project.

> open a branch which does not start from my master

Assuming the Netflix remote is called `upstream`, and your fork's remote is called `origin`, you can do:

```
git fetch --all
git checkout -b new-branch upstream/master
```

This creates a new local branch based on the current Netflix master branch.

Then commit and push:

```
# edit files
git add all my four files
git commit -m "my changes"
git push -u origin new-branch
```

This will create the new branch on your fork, and GitHub will ask you whether you want to make a PR against the upstream repo.

If there is a change from the upstream master that you want to integrate into your branch:

```
git fetch --all
git checkout new-branch
git rebase upstream/master
```

foosoftsrl:
Hei Werner,
thanks a lot... I started just recently to make friends with git. Forks... is the whole mess of git to an entirely new level.
I mean... this PR went down the bin because I took some code from master@foosoft/vmaf without even noticing it.
Than I started issuing random git reset's on the damned to-be-pr branch in order to remove unwanted commits.
Something along...

```
git remote add upstream https://github.com/Netflix/vmaf.git
git clone http://my_repo
git checkout my_branch
git reset --hard upstream/master
(copy the 4 files)
git commit -a 
git push origin master --force
```

Correct?
Can you explain to me why git fetch --all is needed before git checkout? At first glance it does not seem logically necessary (ehm... I bet that the correct answer is RTFM :)

slhck:
You have a few things swapped in your commands. Start from scratch:

```bash
git clone https://github.com/foosoftsrl/vmaf.git
cd vmaf
git remote add upstream https://github.com/Netflix/vmaf.git
git fetch --all # download the changes from the upstream
git checkout master # not needed, just to make sure
git reset --hard upstream/master # this resets your fork's master branch to the Netflix one 
git push --force -u origin master # this pushes your reset back to the GitHub repo
git switch -c my_branch # create a new branch and switch over to it
git reset --hard upstream/master # you can do this but it's not needed now, since your master was already reset
(copy the 4 files)
git add .
git commit -m "your commit message"
git push -u origin my_branch # you have to push your new branch
```

In case there are changes on Netflix upstream:

```
git fetch --all
git switch master
git rebase upstream/master
git switch my_branch
git rebase master
```

You need to `fetch` remote changes! Otherwise git doesn't know what happened on the server side. A  `checkout` only works on your local copy and doesn't sync to a remote server.

And yes, Git is complicated.

PR #947, opened 2021-11-24
CAMBI heatmap generation

nilfm:
* Enables string options for VMAF
* Factors out buffers used by CAMBI from `CambiState`  into `CambiBuffers` to avoid clutter
* Adds optional parameter to CAMBI which provides a path to save one 16-bit YUV400 (grayscale) video for each scale, representing the `c_values` at that scale. This allows the user to visualize the bands detected by CAMBI and their temporal movement.

Comments:
* If we agree that the YUV400 option is better than the multiple PFM option, we can squash the last 4 commits into one. I'm keeping it this way for now in case we want to leave the last commit out (in which case, we can squash the last 3).
* It'd be good to test this on Windows, I don't have a Windows system but I'll try to find one to try it on. Mostly worried about the path separator.
* I've tested this on Mac OS by opening the generated .yuv files with `vooya`.
* (minor) I'm not sure if `.yuv` is the right extension for single-channel videos, I haven't found a better one.
* It would be good to check if the provided path is not empty and exit to avoid possibly corrupting user data.

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -24,7 +24,9 @@
 #include "common/macros.h"
 #include "feature_collector.h"
 #include "feature_extractor.h"
+#include "log.h"
 #include "mem.h"
+#include "mkdirp.h"
 #include "picture.h"
 
 /* Ratio of pixels for computation, must be 0 < topk <= 1.0 */
@@ -38,9 +40,6 @@
 
 /* Max log contrast luma levels */
 #define DEFAULT_MAX_LOG_CONTRAST (2)
-
-/* Path to an existing directory where the heatmaps will be dumped. If null, the heatmaps will not be dumped to disk. */
-#define DEFAULT_CAMBI_HEATMAPS_PATH (NULL)
 
 #define CAMBI_MIN_WIDTH (320)
 #define CAMBI_MAX_WIDTH (4096)
@@ -73,12 +72,6 @@ static int *g_all_diffs;
         x = y;               \
         y = temp;            \
     }
-
-#ifdef _WIN32
-    #define PATH_SEPARATOR '\\'
-#else
-    #define PATH_SEPARATOR '/'
-#endif
 
 #define PICS_BUFFER_SIZE 2
 #define MASK_FILTER_SIZE 7
@@ -163,10 +156,10 @@ static const VmafOption options[] = {
     },
     {
         .name = "heatmaps_path",
-        .help = "Path to an existing directory where the heatmaps will be dumped. If null, the heatmaps will not be dumped to disk.",
+        .help = "Path where heatmaps will be dumped.",
         .offset = offsetof(CambiState, heatmaps_path),
         .type = VMAF_OPT_TYPE_STRING,
-        .default_val.s = DEFAULT_CAMBI_HEATMAPS_PATH,
+        .default_val.s = NULL,
     },
     { 0 }
 };
@@ -799,6 +792,12 @@ static void write_uint16(FILE *file, uint16_t v) {
     fwrite((void*)(&v), sizeof(v), 1, file);
 }
 
+#ifdef _WIN32
+    #define PATH_SEPARATOR '\\'
+#else
+    #define PATH_SEPARATOR '/'
+#endif
+
 static int dump_c_values(const float *c_values, int width, int height, int scale, 
                           int window_size, const char *dir_path, const uint16_t num_diffs) {
     int max_diff_weight = g_diffs_weights[0];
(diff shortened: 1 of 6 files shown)


kylophone:
This will probably have merge conflicts with #941.

libvmaf/src/feature/cambi.c
```diff
@@ -719,9 +734,33 @@ static FORCE_INLINE inline double weight_scores_per_scale(double *scores_per_sca
     return score / normalization;
 }
 
-static int cambi_score(VmafPicture *pics, uint32_t *mask_dp, uint16_t window_size, double topk,
-                       const uint16_t *tvi_for_diff, float *c_values, uint16_t *c_values_histograms, 
-                       uint8_t *filter_mode_histogram, uint16_t *filter_mode_buffer, double *score) {
+void write_uint16(FILE *file, uint16_t v) {
```
ltrudeau-twoorioles, line 737: This function could probably be `static`.

libvmaf/src/feature/cambi.c
```diff
@@ -719,9 +734,33 @@ static FORCE_INLINE inline double weight_scores_per_scale(double *scores_per_sca
...
+        }
+    }
+    char path[1024];
+    snprintf(path, sizeof(path), "%s/cambi_heatmap_scale_%d.yuv", dir_path, scale);
+    int max_c_value = max_diff_weight * window_size * window_size / 4;
+    int scaling_value = (1 << 16) - 1;
+    FILE *file = fopen(path, "a");
+    for (int i = 0; i < height; i++) {
+        for (int j = 0; j < width; j++) {
+            uint16_t val = scaling_value * c_values[i * width + j] / max_c_value;
```
ltrudeau-twoorioles, line 755: Might be faster to compute `max_c_value_inv = 1/max_c_value` outside the nested loops and use `* max_c_value_inv` inside the nested loops

libvmaf/src/feature/cambi.c
```diff
@@ -719,9 +734,33 @@ static FORCE_INLINE inline double weight_scores_per_scale(double *scores_per_sca
...
 }
 
-static int cambi_score(VmafPicture *pics, uint32_t *mask_dp, uint16_t window_size, double topk,
-                       const uint16_t *tvi_for_diff, float *c_values, uint16_t *c_values_histograms, 
-                       uint8_t *filter_mode_histogram, uint16_t *filter_mode_buffer, double *score) {
+void write_uint16(FILE *file, uint16_t v) {
+    fwrite((void*)(&v), sizeof(v), 1, file);
+}
+
+static void dump_c_values(float *c_values, int width, int height, int scale, int window_size, char *dir_path) {
```
ltrudeau-twoorioles, line 741: How about `const float *c_values` and `const char *dir_path`

ltrudeau-twoorioles:
> * (minor) I'm not sure if `.yuv` is the right extension for single-channel videos, I haven't found a better one.

How about `.y4m`?

libvmaf/src/feature/cambi.c
nilfm: True, we can even do `scaling_value = ((1 << 16) - 1)/max_c_value` since we're multiplying by that anyway.

nilfm:
@ltrudeau-twoorioles, I have discussed the appropriate file extension with @krasuluk and we found that `.gray` is used elsewhere for this purpose, so we'll be using that. Thanks for the comments!

libvmaf/src/feature/cambi.c
```diff
@@ -719,9 +734,34 @@ static FORCE_INLINE inline double weight_scores_per_scale(double *scores_per_sca
...
+        if (g_diffs_weights[i] > max_diff_weight) {
+            max_diff_weight = g_diffs_weights[i];
+        }
+    }
+    char path[1024];
+    snprintf(path, sizeof(path), "%s/cambi_heatmap_scale_%d_%dx%d_16b.gray", dir_path, scale, width, height);
+    int max_c_value = max_diff_weight * window_size * window_size / 4;
+    int max_16bit_value = (1 << 16) - 1;
+    double scaling_value = (double)max_16bit_value / max_c_value;
+    FILE *file = fopen(path, "a");
```
kylophone, line 753: Your call to `fopen()` can fail, add a check here and bail if it's no good.

libvmaf/src/feature/cambi.c
```diff
@@ -719,9 +734,34 @@ static FORCE_INLINE inline double weight_scores_per_scale(double *scores_per_sca
...
+
+static void dump_c_values(const float *c_values, int width, int height, int scale, int window_size, const char *dir_path) {
+    int max_diff_weight = g_diffs_weights[0];
+    for (int i = 0; i < NUM_DIFFS; i++) {
+        if (g_diffs_weights[i] > max_diff_weight) {
+            max_diff_weight = g_diffs_weights[i];
+        }
+    }
+    char path[1024];
+    snprintf(path, sizeof(path), "%s/cambi_heatmap_scale_%d_%dx%d_16b.gray", dir_path, scale, width, height);
```
kylophone, line 749: "/" is probably not a cross-platform.

libvmaf/src/feature/cambi.c
nilfm: By "bail" you mean abort the whole program, right? I would say that continuing execution without writing to disk is not desirable.

kylophone: No, just return an error and handle it somehow. If fopen fails and your file pointer is NULL you will get a SIGSEGV.

libvmaf/src/feature/cambi.c
nilfm: I have seen some [posts](https://stackoverflow.com/questions/12971499/how-to-get-the-file-separator-symbol-in-standard-c-c-or/30261289) suggesting something like this to set up the separator:
```
const char separator = 
#ifdef _WIN32
    '\\'
#else
    '/'
#endif
;
```

There are a couple references to `_WIN32` in this repo, but none of them is used to determine the separator in this way. Do you think this is fine?

nilfm:
I'll wait for #941 to be pushed and then rebase these changes on top of that.

kylophone:
#941 is merged. This one can be rebased now.

nilfm:
Rebased on top of #941 changes

joelsole (approved):

XuShiSheng:
你好！邮件已收到，我会尽快回复你--许思焱

libvmaf/src/feature/cambi.c
```diff
@@ -911,6 +920,11 @@ static int close_cambi(VmafFeatureExtractor *fex) {
     aligned_free(s->buffers.mask_dp);
     aligned_free(s->buffers.filter_mode_histogram);
     aligned_free(s->buffers.filter_mode_buffer);
+
+    for (int scale = 0; scale < NUM_SCALES; scale++) {
```
kylophone, line 924: maybe add a guard for s->heatmaps_path here.

PR #948, opened 2021-11-24
Fix quick-select for edge case n = k

nilfm:
When `n = k` in quick-select, we can immediately return as the top `n` elements will always be in the first `n` spots (the whole array). This fixes an edge case with this quick-select implementation which previously would have accessed out of bounds and caused UB. This case is never present in the actual CAMBI code with default parameters since `k = 0.6*n`, but could come up if the `topk` parameter is set to 1.


joelsole (approved):

kylophone:
Tested with ASAN, looks good. Pushed.

PR #968, opened 2021-12-15
Full-reference mode for CAMBI

nilfm:
Added an optional flag `full_reference` to CAMBI to allow running it as a full-reference metric, subtracting the banding that is already present in the source and obtaining the "amount of banding added" by the encoding process. Added optional parameters `src_width`, `src_height` to allow different processing resolutions on source vs encode. 

Added a new feature extractor, `CambiFullReferenceFeatureExtractor`, to access this behavior from the Python interface.
Added a new quality runner, `CambiFullReferenceQualityRunner`, to wrap the feature extractor.

Added e2e tests and updated both `cambi.md` and the news section on `README.md`.

Some examples taken from the test AOM sequences where the full-reference CAMBI is significantly different from the no-reference CAMBI:

| Title  | No-Reference CAMBI | Full-Reference CAMBI | Difference | Comments
| :-------------: | :-------------: | :-------------: | :-------------: | :-------------: |
| MissionControlClip3 | 2.07 - 2.23 | 0.00 - 0.13 | ~2.1 | Bands are present on the source clip, no banding is added by compression |
| NewsIntroAnchor  | 4.82 - 5.73 | 0.26 - 1.06 | ~4.6 | Artifacts are definitely present in the source. It's debatable if it's blocking or banding artifacts. No-ref CAMBI overestimates the amount of banding added by encoding. |
| ShakyBaseball | 7.06 - 8.43 |  5.06 - 6.43 | ~2.0 | The camera moves too quickly to appreciate the banding but arguably there is banding on the still frames, especially on the blue background wall |

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -41,6 +41,9 @@
 /* Max log contrast luma levels */
 #define DEFAULT_MAX_LOG_CONTRAST (2)
 
+/* If true, CAMBI will be run in full-reference mode and will use both the reference and distorted inputs */
+#define DEFAULT_CAMBI_FULL_REF_FLAG (false)
+
 #define CAMBI_MIN_WIDTH (320)
 #define CAMBI_MAX_WIDTH (4096)
 #define CAMBI_4K_WIDTH (3840)
@@ -89,12 +92,16 @@ typedef struct CambiState {
     VmafPicture pics[PICS_BUFFER_SIZE];
     unsigned enc_width;
     unsigned enc_height;
+    unsigned src_width;
+    unsigned src_height;
     uint16_t *tvi_for_diff;
     uint16_t window_size;
+    uint16_t src_window_size;
     double topk;
     double tvi_threshold;
     uint16_t max_log_contrast;
     char *heatmaps_path;
+    bool full_ref;
     CambiBuffers buffers;
 } CambiState;
 
@@ -112,6 +119,24 @@ static const VmafOption options[] = {
         .name = "enc_height",
         .help = "Encoding height",
         .offset = offsetof(CambiState, enc_height),
+        .type = VMAF_OPT_TYPE_INT,
+        .default_val.i = 0,
+        .min = 200,
+        .max = 4320,
+    },
+    {
+        .name = "src_width",
+        .help = "Source width. Only used when full_ref=true.",
+        .offset = offsetof(CambiState, src_width),
+        .type = VMAF_OPT_TYPE_INT,
+        .default_val.i = 0,
+        .min = 320,
+        .max = 7680,
+    },
+    {
+        .name = "src_height",
+        .help = "Source height. Only used when full_ref=true.",
+        .offset = offsetof(CambiState, src_height),
         .type = VMAF_OPT_TYPE_INT,
         .default_val.i = 0,
         .min = 200,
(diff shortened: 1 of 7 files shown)


libvmaf/src/feature/cambi.c
```diff
@@ -810,25 +852,59 @@ static int cambi_score(VmafPicture *pics, uint32_t *mask_dp, uint16_t window_siz
...
+        if (err) return err;
+
+        err = vmaf_feature_collector_append(feature_collector, "cambi_source", src_score, index);
+        if (err) return err;
+
+        double combined_score = combine_dist_src_scores(dist_score, src_score);
+        err = vmaf_feature_collector_append(feature_collector, "cambi_full_reference", combined_score, index);
+        if (err) return err;
+    }
+    else {
```
nilfm, line 899: If the `full_reference` flag is not set, this block returns -1 on the `cambi_source` and `cambi_full_reference` metrics. This is to avoid conditionally returning features. There may be a cleaner way to do this, please tell me if that is the case.

libvmaf/src/feature/cambi.c
```diff
@@ -109,6 +116,24 @@ static const VmafOption options[] = {
         .min = 200,
         .max = 4320,
     },
+    {
```
li-zhi, line 119: Not sure if we need this. The input `width` and `height` should apply to both the reference and distorted videos.

Were you try to mimic what `enc_width` and `enc_height` does? I think these parameters were needed because the w/h used for encoding may be different from `width` and `height` (for example, your input distorted video is upsampled from an encode of a smaller resolution). I don't see a similar situation apply here.

@mafonsoN Do you want to comment on this as well?

libvmaf/src/feature/cambi.c
```diff
@@ -147,6 +172,13 @@ static const VmafOption options[] = {
         .min = 0,
         .max = 5,
     },
+    {
+        .name = "full_reference",
```
li-zhi, line 176: Should we abbreviate as `full_ref`?

libvmaf/src/feature/cambi.c
nilfm: We had this discussion with Joel and Mariana at some point, here is the conclusion we reached:
"This (optional) parameter is to account for the cases where a video has been scaled before libvmaf is called. If the source was scaled from, say, 480p to 1080p, we need to have an option to supply the 480p."

cc @joelsole, @mafonsoN

libvmaf/src/feature/cambi.c
nilfm: agreed, `full_reference` is a bit unwieldy

libvmaf/src/feature/cambi.c
li-zhi: Ok. So this is symmetric to `enc_width` and `enc_height`? Just to double check.

We can add these.

mafonsoN: @li-zhi yes, this is added for completeness to handle any use case.

mafonsoN: @nilfm can we add a test case where `src_width` and `src_height` are exercised?

nilfm: Added in one of the follow-up commits.

joelsole (approved):

christosbampis (approved):

li-zhi (approved):

mafonsoN (approved):

ltrudeau-twoorioles:
Great I don't need to do this manually anymore.

PR #969, opened 2021-12-20
CAMBI: support 12-bit and 16-bit inputs by converting them to 10b

nilfm:
CAMBI previously supported 8-bit and 10-bit inputs. This PR adds support for 12-bit and 16-bit inputs. Internally, CAMBI still works at the 10-bit level, and the inputs are upscaled/downscaled appropriately in the preprocessing steps.

Added unit tests to cover these cases, also manually upscaled a 10b video to 12b and 16b and the results were identical.

This PR refactors the upscaling/downscaling code into two functions, one to handle the case where the input is given as `uint8` and the other for `uint16`.

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -428,7 +428,49 @@ static int init(VmafFeatureExtractor *fex, enum VmafPixelFormat pix_fmt,
 }
 
 /* Preprocessing functions */
-static void decimate_generic_10b(const VmafPicture *pic, VmafPicture *out_pic, unsigned out_w, unsigned out_h) {
+
+// For bitdepths <= 8.
+static void decimate_generic_uint8_and_convert_to_10b(const VmafPicture *pic, VmafPicture *out_pic, unsigned out_w, unsigned out_h) {
+    uint8_t *data = pic->data[0];
+    uint16_t *out_data = out_pic->data[0];
+    ptrdiff_t stride = pic->stride[0];
+    ptrdiff_t out_stride = out_pic->stride[0] >> 1;
+    unsigned in_w = pic->w[0];
+    unsigned in_h = pic->h[0];
+
+    int shift_factor = 10 - pic->bpc;
+
+    // if the input and output sizes are the same
+    if (in_w == out_w && in_h == out_h){
+        for (unsigned i = 0; i < out_h; i++) {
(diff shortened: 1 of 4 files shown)


libvmaf/src/feature/cambi.c
```diff
@@ -366,28 +374,41 @@ static void decimate_generic_10b(const VmafPicture *pic, VmafPicture *out_pic) {
...
-    ptrdiff_t stride = pic->stride[0];
+    ptrdiff_t stride = pic->stride[0] >> 1;
     ptrdiff_t out_stride = out_pic->stride[0] >> 1;
     unsigned in_w = pic->w[0];
     unsigned in_h = pic->h[0];
     unsigned out_w = out_pic->w[0];
     unsigned out_h = out_pic->h[0];
 
+    // A bitdepth of 9 is not valid in VMAF, so here it will always be >= 10
+    int shift_factor = bitdepth - 10;
```
kylophone, line 396: Better to read bitdepth from `pic->bpc`.

nilfm:
Rebased on top of the FR CAMBI changes. This is ready for review now.

mafonsoN:
@nilfm for some reason I can't add a comment to the specific lines but in with regards to the handling of bit depth of 9, even though VMAF doesn't support it, it might be good to add this anyway for completeness.
```
// A bitdepth of 9 is not valid in VMAF, so here it will always be >= 10
int shift_factor = pic->bpc - 10;
int rounding_offset = shift_factor == 0 ? 0 : 1 << (shift_factor - 1);
```
Otherwise these changes good to me.

joelsole (commented):
You could add an explanation in cambi.md detailing its behavior for 12 and 16b, since it might not be the expected one for some users.

nilfm:
@mafonsoN The problem with a bit depth of 9 is that it's a pretty special case, as it should be done on the `uint16` function but has to be scaled up (`<<`) instead of down (`>>`), so supporting it would add extra complexity, and it'd probably be best to implement it as a separate function to avoid having extra conditionals on the usual paths. If we agree that this is worth it I'll implement it, just pointing out the downsides.

nilfm:
* Supported 9b inputs for future proofing and completeness
* Updated cambi.md
* Fixed the usage message which had incorrect supported bitdepths listed

mafonsoN (approved):

PR #973, opened 2022-01-14
Factor EOTF out of CAMBI

nilfm:
Since we want to use EOTFs outside of CAMBI, this PR factors out the EOTF functions into a separate file where they can be accessed by other features as well.

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -25,6 +25,7 @@
 #include "feature_collector.h"
 #include "feature_extractor.h"
 #include "log.h"
+#include "luminance_tools.h"
 #include "mem.h"
 #include "mkdirp.h"
 #include "picture.h"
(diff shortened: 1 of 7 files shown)


nilfm:
Some decisions I'm not sure about:
* Naming: `LumaRange` contains the bitdepth, lowest and highest possible values, and `PixelRange` contains either `LIMITED` or `FULL` and is used to construct the `LumaRange`. There may be better names for both.
* Usage of this API: My idea is that a user would call `LumaRange_init` once with the right parameters and then repeatedly call `get_luminance`. If the user has their values normalized already (in the [0, 1] range), they can just call `bt1886_eotf` directly. The other functions are only exposed on the header file so they can be tested, but if there's a way to avoid that it'd be cleaner.
* Extensibility: It should be easy to add a new EOTF just like `bt1886_eotf`. It should also be easy to add a new `PixelRange`, although I don't know of any other option there. If we add other EOTFs, it might be good to have a function `get_eotf` which takes an enum `EOTFType` and returns the appropriate function.

PR #976, opened 2022-01-19
Add flexibility to CAMBI CLI parameters

nilfm:
* Implement PQ EOTF and add a CLI option to use it in CAMBI for the visibility threshold calculations.
* Add a CLI option to CAMBI to provide the encoding bitdepth so it can choose whether to use the anti-dithering filter accordingly. This is for cases where the original bitdepth is < 10 and it has been upscaled to a bitdepth >= 10.
* A few minor style/uniformity changes.

### DIFF
--- a/libvmaf/src/feature/luminance_tools.c
+++ b/libvmaf/src/feature/luminance_tools.c
@@ -66,7 +66,7 @@ int vmaf_luminance_init_luma_range(VmafLumaRange *luma_range, int bitdepth, enum
     return err;
 }
 
-int vmaf_luminance_init_eotf(VmafEOTF *eotf, char* eotf_str) {
+int vmaf_luminance_init_eotf(VmafEOTF *eotf, const char *eotf_str) {
     if (strcmp(eotf_str, "bt1886") == 0) {
         *eotf = vmaf_luminance_bt1886_eotf;
     }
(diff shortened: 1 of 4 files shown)


kylophone:
lgtm

libvmaf/src/feature/luminance_tools.c
```diff
@@ -65,13 +66,38 @@ int vmaf_luminance_init_luma_range(VmafLumaRange *luma_range, int bitdepth, enum
...
 }
 
+double vmaf_luminance_pq_eotf(double V) {
+    double m_1 = 0.1593017578125;
+    double m_2 = 78.84375;
+    double c_1 = 0.8359375;
+    double c_2 = 18.8515625;
+    double c_3 = 18.6875;  // c_3 = c_1 + c_2 - 1
+    double num = MAX(0, pow(V, 1.0 / m_2) - c_1);
+    double den = c_2 - c_3 * pow(V, 1.0 / m_2);
```
1480c1, line 97: These can probably be declared as const, and for `num`, it might be better to separate out the `pow` call from the `MAX` macro, so it's not called multiple times

libvmaf/src/feature/luminance_tools.c
```diff
@@ -65,13 +66,38 @@ int vmaf_luminance_init_luma_range(VmafLumaRange *luma_range, int bitdepth, enum
     return err;
 }
 
+int vmaf_luminance_init_eotf(VmafEOTF *eotf, char* eotf_str) {
```
1480c1, line 69: `const char*` might be better since eotf_str is not being modified

libvmaf/src/feature/luminance_tools.c
nilfm: Thanks for your comments, the last commit addresses them.

christosbampis (approved):

joelsole (approved):

PR #978, opened 2022-01-25
CAMBI: eliminate mutable global variables

nilfm:
These mutable global variables were causing issues when CAMBI was run multi-threaded, using the `--threads` CLI option for VMAF. Each thread would run malloc and free, and it resulted in freeing a non-malloc'd object and crashing. This PR eliminates those global variables and integrates them in `CambiState`. After this fix, running CAMBI with `--threads 8` doesn't crash anymore.

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -55,15 +55,10 @@
 
 #define NUM_SCALES 5
 static const int g_scale_weights[NUM_SCALES] = {16, 8, 4, 2, 1};
-static FILE *g_heatmaps_paths[NUM_SCALES];
 
 /* Suprathreshold contrast response */
 static const int g_contrast_weights[32] = {1, 2, 3, 4, 4, 5, 5, 6, 6, 6, 6, 7, 7, 7, 7, 8,
                                            8, 8, 8, 8, 8, 8, 9, 9, 9, 9, 9, 9, 9, 9, 9, 9};
-
-static uint16_t *g_diffs_to_consider;
-static int *g_diffs_weights;
-static int *g_all_diffs;
 
 #define MAX(x, y) (((x) > (y)) ? (x) : (y))
 #define MIN(x, y) (((x) < (y)) ? (x) : (y))
@@ -90,6 +85,10 @@ typedef struct CambiBuffers {
     uint16_t *c_values_histograms;
     uint16_t *filter_mode_buffer;
     uint8_t *filter_mode_histogram;
+    uint16_t *diffs_to_consider;
+    uint16_t *tvi_for_diff;
+    int *diff_weights;
+    int *all_diffs;
 } CambiBuffers;
 
 typedef struct CambiState {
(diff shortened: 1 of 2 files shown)


libvmaf/src/feature/cambi.c
```diff
@@ -987,8 +987,9 @@ static int preprocess_and_extract_cambi(CambiState *s, VmafPicture *pic, double
     int err = cambi_preprocessing(pic, &s->pics[0], width, height, s->enc_bitdepth);
     if (err) return err;
 
-    err = cambi_score(s->pics, window_size, s->topk, num_diffs, s->tvi_for_diff, 
-                      s->buffers, score, s->heatmaps_path, width, height, is_src);
+    bool write_heatmaps = (s->heatmaps_path != NULL && !is_src);
```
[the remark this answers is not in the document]
kylophone: Either way. I usually omit that comparison and just check the pointer.

kylophone:
lgtm

kylophone:
> Multithreading is not compatible with the `heatmaps_path` feature, as the frames will be written in an arbitrary order and may interleave with each other.

A better approach would be to compute offsets and use mmap, I think.

nilfm:
> A better approach would be to compute offsets and use mmap, I think.

Noted, I'll make this change at some point. I guess this would use extra memory in the order of `num_threads * (size of a frame)`, and the size of a frame is about 33MB in 4k so I'd say this is doable if there aren't too many threads.

libvmaf/src/feature/cambi.c
```diff
@@ -912,36 +912,35 @@ static FORCE_INLINE inline double weight_scores_per_scale(double *scores_per_sca
...
+    FILE *file = heatmaps_files[scale];
+    uint16_t *to_write = malloc(width * sizeof(uint16_t));
     for (int i = 0; i < height; i++) {
         for (int j = 0; j < width; j++) {
-            uint16_t val = (uint16_t)(scaling_value * c_values[i * width + j]);
-            write_uint16(file, val);
+            to_write[j] = (uint16_t)(scaling_value * c_values[i * width + j]);
         }
+        ptrdiff_t offset = (frame * height + i) * width * sizeof(uint16_t);
+        fseek(file, offset, SEEK_SET);
```
nilfm, line 933: So I've tried this `fseek` idea and it seems to work, with `--threads 8` it has generated the correct heatmap in the ~5 times I've tried it so (at least on my machine/os) it is thread-safe. @kylophone any concern?

Is it a problem to have the malloc in line 927? My concern with doing it in `init` is that there are already way too many parameters to these functions and it's quite messy

kylophone: > So I've tried this `fseek` idea and it seems to work, with `--threads 8` it has generated the correct heatmap in the ~5 times I've tried it so (at least on my machine/os) it is thread-safe. @kylophone any concern?

Should be OK. No file pointers (FILE*) are shared between threads, so I don't think there will be any issues. Performance is probably not going to be great, but better than a jumbled mess. 

> Is it a problem to have the malloc in line 927? My concern with doing it in `init` is that there are already way too many parameters to these functions and it's quite messy

It's probably ok, but it's an unchecked malloc right now.

nilfm:
Checked the malloc, tested it on a few sequences with 4 and 8 threads, ready to be merged when the tests pass.

PR #983, opened 2022-01-28
Fix ffmpeg CI workflow

nilfm:
Currently the ffmpeg CI workflow is just failing silently on both [Ubuntu](https://github.com/Netflix/vmaf/runs/4984023460?check_suite_focus=true) and [Mac](https://github.com/Netflix/vmaf/runs/4984023520?check_suite_focus=true).

Ubuntu:
```
93.663925
vmaf Score doesn't match 93.656203
Error: Process completed with exit code 1.
```

Mac:
```
Run cat /proc/cpuinfo
cat: /proc/cpuinfo: No such file or directory
Error: Process completed with exit code 1.
```

This PR removes the line `cat /proc/cpuinfo` and updates the expected score so the test passes. If there is a reason to have the `cat` line, feel free to reject, but it doesn't seem to be doing anything.

### DIFF
--- a/.github/workflows/ffmpeg.yml
+++ b/.github/workflows/ffmpeg.yml
@@ -56,12 +56,11 @@ jobs:
           sudo make -C ffmpeg --quiet -j $(getconf _NPROCESSORS_ONLN 2>/dev/null || sysctl -n hw.ncpu) install
       - name: Test ffmpeg
         run: |
-          cat /proc/cpuinfo
           curl "https://gist.githubusercontent.com/1480c1/0c4575da638ef6e8203feffd0597de16/raw/akiyo_cif.tar.xz.base64" | base64 -d | tar xJ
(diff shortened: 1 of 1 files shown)


1480c1:
iirc, the only reason why the `cat /proc/cpuinfo` line was there was to check if avx2 etc. were present as a long time ago you weren't guaranteed a good CPU (sometimes happened in SVT-AV1 leading to longer CI times), probably fine to remove it now

kylophone:
FYI, I think the reason that the numeric scores have changed is because the new `vf_libvmaf` libavfilter was recently updated.

nilfm:
> This pr seems to contain extraneous commits, please check the commit this was rebased on

Weird, those other commits are already on master. I found this trick of changing the target branch so github figures it out, it seems to have worked.

1480c1:
That's a new feature to me, nice

PR #990, opened 2022-02-21
CAMBI: Change in c_value_pixel

nilfm:
[Pending validation, do not merge]

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -736,12 +736,8 @@ static float c_value_pixel(const uint16_t *histograms, uint16_t value, const int
         if (value <= tvi_thresholds[d]) {
             uint16_t p_1 = histograms[(value + diffs[num_diffs + d + 1]) * histogram_width + histogram_col];
             uint16_t p_2 = histograms[(value + diffs[num_diffs - d - 1]) * histogram_width + histogram_col];
-            if (p_1 > p_2) {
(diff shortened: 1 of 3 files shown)


libvmaf/src/feature/cambi.c
```diff
@@ -736,12 +736,8 @@ static float c_value_pixel(const uint16_t *histograms, uint16_t value, const int
         if (value <= tvi_thresholds[d]) {
             uint16_t p_1 = histograms[(value + diffs[num_diffs + d + 1]) * histogram_width + histogram_col];
             uint16_t p_2 = histograms[(value + diffs[num_diffs - d - 1]) * histogram_width + histogram_col];
-            if (p_1 > p_2) {
-                val = (float)(diff_weights[d] * p_0 * p_1) / (p_1 + p_0);
-            }
-            else {
-                val = (float)(diff_weights[d] * p_0 * p_2) / (p_2 + p_0);
-            }
+            uint16_t p_12 = p_1 + p_2;
```
li-zhi, line 739: @nilfm Any context for this change?

nilfm: There is a thread in the CAMBI channel where I shared a doc with some explanation, I'll link you there. It's meant to help when the window contains 3 different bands [0 | 1 | 2], without this change only the maximum out of bands 0 and 2 is used, and the other one's discarded. I think both should contribute to that pixel's c_value. I still need to validate that this doesn't mess up current scores, though.

PR #997, opened 2022-03-17
CAMBI: performance optimizations

nilfm:
* The mode filter now ignores the edges, which caused a lot of additional computation for no added benefit
* The 3x3 mode filter is substituted for iterated 1x3 and 3x1 mode filters, which aren't exactly equivalent (1) but will be equivalent on a large majority of inputs
* Unrolled a loop in the `c_value` computation to avoid checking for edges in the main part of the loop.

Speedups ranging from +27% (banding-prone videos) to +341% (non-banding-prone videos), speedup % is correlated to the spatial mask being mostly off.

(1) Example of non-equivalence:
```
1 1 1
1 2 2
1 2 2
```

which will give 2 as the mode when it should be 1.

### DIFF
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -77,7 +77,6 @@ typedef struct CambiBuffers {
     uint32_t *mask_dp;
     uint16_t *c_values_histograms;
     uint16_t *filter_mode_buffer;
-    uint8_t *filter_mode_histogram;
     uint16_t *diffs_to_consider;
     uint16_t *tvi_for_diff;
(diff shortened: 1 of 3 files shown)


nilfm:
I need to take care of the failing tests because the CAMBI score does change slightly

nilfm:
This loop unrolling gives ~7% speedup, up to 20% in some sequences. It also makes it easier to SIMD later.

nilfm:
Unrolling this loop gives 3-7% speedup and facilitates the implementation of SIMD later if needed.

nilfm:

PR #999, opened 2022-03-23
CAMBI: SIMD implementation

nilfm:
SIMD implementation of the histogram updates in `calculate_c_values` and changing the window size from 63 to 65 to better use the 16-elements-at-a-time vectorization.


python/test/cambi_test.py
```diff
@@ -45,9 +45,9 @@ def test_run_cambi_fextractor_scaled(self):
 
         # score: arithmetic mean score over all frames
         self.assertAlmostEqual(results[0]['Cambi_feature_cambi_score'],
-                               0.8740225, places=4)
+                               0.79470425, places=4)
```
li-zhi, line 48: Are these numerical change caused by SIMD??

nilfm: No, SIMD keeps the scores the same. But this PR also changes the window size from 63 to 65 which changes the scores. This is because 63 = 16*3+15 so it’s a bad fit for SIMD which operates in chunks of 16 items.

li-zhi: In that case, it is best to do incremental changes: first PR changes the window size with numerical values; second PR changes to SIMD.

Also: if numerical changes, we need to update pycambi in vmaf_oss to make numerical values consistent.

kylophone:
> changing the window size from 63 to 65 to better use the 16-elements-at-a-time vectorization.

Couldn't you just pad 63 one element to make it 64? If you do that, there is no special case remainder to deal with.

nilfm:
@kylophone sorry for editing your comment, I thought I was writing a new one... I've edited it back to where it was. Here's my comment

> Couldn't you just pad 63 one element to make it 64? If you do that, there is no special case remainder to deal with.

I assume you mean iterating 3 times with a full SIMD vector and a fourth time padding it with a zero at the end? If that's the case, I see a few problems
* it's not always 63, because there are edge conditions (we could avoid SIMDing those) and because resolutions other than 4k have smaller windows (this is harder to sidestep). So we can't really hardcode the padding, and generating it dynamically will probably be much slower
* a window size of 65 avoids all these issues, and the scores produced by it seem to be as good or slightly better

Also, I've checked godbolt and both gcc and clang with `-O3` emit SIMD instructions without this PR's changes so that's why it magically speeds up when changing the window size to 65. Not sure what to do with these changes now

nilfm:

PR #1000, opened 2022-03-23
CAMBI: change window_size to 65

nilfm:
Benchmarks show that using a window size of 65 is better than 63 due to SIMD-ability, and results don't change much (and change slightly for the better). In fact, it seems that just by changing the window size to 65, the code is significantly faster without explicitly implementing SIMD, probably because the compiler is emitting those instructions by itself.


kylophone:
See: https://github.com/Netflix/vmaf/pull/999#issuecomment-1075891503

nilfm:

PR #1004, opened 2022-03-31
Fix libvmaf warnings

nilfm:
Fixed some warnings that showed up when building libvmaf. Most of them related to incompatible types (or constness) where a cast is needed.


libvmaf/src/feature/feature_extractor.c
```diff
@@ -287,7 +287,7 @@ get_fex_list_entry(VmafFeatureExtractorContextPool *pool,
     entry.ctx_list = malloc(ctx_array_sz);
     if (!entry.ctx_list) goto fail;
     memset(entry.ctx_list, 0, ctx_array_sz);
-    int err = vmaf_dictionary_copy(&opts_dict, &entry.opts_dict);
+    vmaf_dictionary_copy(&opts_dict, &entry.opts_dict);
```
nilfm, line 290: @kylophone These changes should be mostly safe, this is one that I wasn't sure about, currently `err` is being thrown away but if I add `if (err) goto fail;` the unit tests fail because they pass a null dict to `vmaf_fex_ctx_pool_aquire`. So it's not clear to me if we want to fail on this.

libvmaf/src/dict.c
```diff
@@ -6,7 +6,7 @@
 #include "dict.h"
 #include "libvmaf/feature.h"
 
-const VmafDictionaryEntry *vmaf_dictionary_get(VmafDictionary **dict,
+VmafDictionaryEntry *vmaf_dictionary_get(VmafDictionary **dict,
                                                const char *key, uint64_t flags)
```
1480c1, line 10: Might want to reformat these two lines, I don't know if @kylophone used a clang-format, so you might need to just manually delete the spaces on line 10 to match the open paren

libvmaf/src/dict.h
```diff
@@ -42,7 +42,7 @@ enum VmafDictionaryFlags {
 int vmaf_dictionary_set(VmafDictionary **dict, const char *key, const char *val,
                         uint64_t flags);
 
-const VmafDictionaryEntry *vmaf_dictionary_get(VmafDictionary **dict,
+VmafDictionaryEntry *vmaf_dictionary_get(VmafDictionary **dict,
                                                const char *key, uint64_t flags);
```
1480c1, line 46: same here

nilfm:
After these changes, the following warnings are left (if compiling with `Denable_float=true`):
* `../src/feature/adm.c:52:14: warning: unused function 'init_dwt_band_d' [-Wunused-function]`
  * I can't get rid of this function because there's a [Cython file](https://github.com/Netflix/vmaf/blob/master/python/vmaf/core/adm_dwt2_cy.pyx#L23) that uses it. Not sure how to best deal with this.
* `../tools/main.cpp:195:11: warning: 'compute_vmaf' is deprecated [-Wdeprecated-declarations]`
  * We'll probably get rid of this old API soon so I'm not worried about this.
* 115 warnings coming from the `vif_avx2.c` file.   
  * We have a pending PR (#996) to rewrite this file, so we can deal with this later
* that's it! this PR fixes about 160 warnings.

PR #1009, opened 2022-04-04
libvmaf: fix unused function warning in adm.c

nilfm:
As mentioned in #1004, this function is used by a Cython file, but it's not used by any C code so it throws an "unused function" warning. My fix is to use `__attribute__((unused))`, which works on all compilers that I've tried. There are also a few usages of `__attribute__` in libvmaf already so I think it shouldn't restrict portability.

After this, the only warnings left are the deprecated `compute_vmaf` call and the `vif_avx2` file. There may be other warnings when compiling with `-Denable_avx512=true` which I haven't checked.


kylophone:
lgtm, please feel free to merge after CI is all green

PR #1010, opened 2022-04-04
libvmaf/vmaf: prepare --aom_ctc v3.0 preset

kylophone:
This PR prepares the `--aom_ctc v3.0` preset. The only difference to previous versions is the addition of the `cambi` metric.

### DIFF
--- a/libvmaf/tools/cli_parse.c
+++ b/libvmaf/tools/cli_parse.c
@@ -287,14 +287,39 @@ static void aom_ctc_v1_0(CLISettings *settings, const char *const app)
         parse_feature_config("psnr_hvs", app);
 }
 
+static void aom_ctc_v2_0(CLISettings *settings, const char *app)
+{
+    aom_ctc_v1_0(settings, app);
+}
+
+static void aom_ctc_v3_0(CLISettings *settings, const char *app)
+{
+    aom_ctc_v2_0(settings, app);
+    settings->feature_cfg[settings->feature_cnt++] =
+        parse_feature_config("cambi", app);
+}
+
 static void parse_aom_ctc(CLISettings *settings, const char *const optarg,
                           const char *const app)
 {
     if (!strcmp(optarg, "proposed"))
-        usage(app, "`--aom_ctc proposed` is deprecated. Use `--aom_ctc v1.0`");
-    else if (!strcmp(optarg, "v1.0") || !strcmp(optarg, "v2.0"))
+        usage(app, "`--aom_ctc proposed` is deprecated.");
+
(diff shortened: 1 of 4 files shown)


libvmaf/tools/cli_parse.c
```diff
@@ -287,13 +287,29 @@ static void aom_ctc_v1_0(CLISettings *settings, const char *const app)
...
 {
     if (!strcmp(optarg, "proposed"))
         usage(app, "`--aom_ctc proposed` is deprecated. Use `--aom_ctc v1.0`");
-    else if (!strcmp(optarg, "v1.0") || !strcmp(optarg, "v2.0"))
+    if (!strcmp(optarg, "v1.0"))
         aom_ctc_v1_0(settings, app);
+    if (!strcmp(optarg, "v2.0"))
+        aom_ctc_v2_0(settings, app);
+    if (!strcmp(optarg, "v3.0"))
+        aom_ctc_v3_0(settings, app);
```
1480c1, line 312: wouldn't the removal of the else chain without an early return potentially call `usage()` on `--aom_ctc v1.0`, or am I missing something else?

kylophone:
`--aom_ctc` version parsing issue is fixed now.

libvmaf/tools/cli_parse.c
```diff
@@ -287,14 +287,39 @@ static void aom_ctc_v1_0(CLISettings *settings, const char *const app)
...
+        parse_feature_config("cambi", app);
+}
+
 static void parse_aom_ctc(CLISettings *settings, const char *const optarg,
                           const char *const app)
 {
     if (!strcmp(optarg, "proposed"))
-        usage(app, "`--aom_ctc proposed` is deprecated. Use `--aom_ctc v1.0`");
-    else if (!strcmp(optarg, "v1.0") || !strcmp(optarg, "v2.0"))
+        usage(app, "`--aom_ctc proposed` is deprecated.");
```
1480c1, line 306: usage calls would be duplicated if `!strcmp(optarg, "proposed")` evals to true, which I guess could be desirable?

kylophone:
> usage calls would be duplicated if !strcmp(optarg, "proposed") evals to true, which I guess could be desirable?

`usage()` ends up calling `exit(1)` after it dumps it message, so not an issue.

1480c1:
> > usage calls would be duplicated if !strcmp(optarg, "proposed") evals to true, which I guess could be desirable?
> 
> `usage()` ends up calling `exit(1)` after it dumps it message, so not an issue.

Oh, didn't notice that, nvm, carry on

joelsole (approved):

PR #1011, opened 2022-04-05
libvmaf: add picture_pool api

kylophone:
The following PR adds a threadsafe picture pool API. This is experimental at this point but is open for comments, questions, etc. 

A `VmafPicturePool` is a place to store a queue of preallocated pictures for continual recycling. By registering a pair of custom alloc/free callbacks, the data buffers will be allocated just once and returned to the pool to be reused when fully unref'd. This is in contrast to the current workflow where each picture is newly allocated and then freed when fully unref'd. This API allows for more user control of `VmafPicture`data buffers. Hopefully this API should be generic enough to work for several usecases: this might mean sharing a decoded picture buffer, GPU memory, reusing an AVFrame data buffer instead of allocating and copying memory.

See `test_picture_pool()` for an example of API usage.

### DIFF
--- a/libvmaf/src/picture.h
+++ b/libvmaf/src/picture.h
@@ -21,6 +21,16 @@
 
 #include "libvmaf/picture.h"
 
+typedef struct VmafPicturePrivate {
+    void *cookie;
+    int (*release_picture)(VmafPicture *pic, void *cookie);
+} VmafPicturePrivate;
+
+int vmaf_picture_priv_init(VmafPicture *pic);
+
 int vmaf_picture_ref(VmafPicture *dst, VmafPicture *src);
+
+int vmaf_picture_set_release_callback(VmafPicture *pic, void *cookie,
+                        int (*release_picture)(VmafPicture *pic, void *cookie));
 
 #endif /* __VMAF_SRC_PICTURE_H__ */
--- a/libvmaf/test/test_picture.c
+++ b/libvmaf/test/test_picture.c
@@ -17,11 +17,15 @@
  */
 
 #include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
 
 #include "test.h"
 #include "picture.h"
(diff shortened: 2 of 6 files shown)


libvmaf/src/picture.h
```diff
@@ -21,6 +21,16 @@
 
 #include "libvmaf/picture.h"
 
+typedef struct VmafPicturePrivate {
+    void *cookie;
```
gedoensmax, line 25: The naming of this variable seems weird to me. Naming it cookie makes me assume it will be the same as the cookie i pass to the picture pool inside the config. But actually this cookie is used as picture pool reference. I would like to be able to have a cookie with every private instead of interpreting this cookie as a pool to then get to its config and at the end to its cookie. Furthermore this means the cookie is the same for all pictures in the pool instead of being able to change.

libvmaf/test/test_picture.c
```diff
@@ -65,9 +69,219 @@ static char *test_picture_data_alignment()
...
+            .pic_pool = pic_pool,
+            .i = i,
+            .timeout = 100000 - (100000 * ((float)i / n)),
+        };
+        err = vmaf_thread_pool_enqueue(thread_pool, request_picture_and_unref,
+                                       &my_thread_pool_data,
+                                       sizeof(my_thread_pool_data));
+        mu_assert("problem during vmaf_thread_pool_enqueue", !err);
+    }
+
```
gedoensmax, line 268: Pasting this snippet here makes my point above a bit clearer. Is it the desired functionality that this cookie changes ? 
```
    //ensure cookie is still MyCookie
    VmafPicture pic;
    vmaf_picture_pool_request_picture(pic_pool, &pic);
    VmafPicturePrivate* priv = pic.priv;
    err = memcmp(&my_cookie, priv->cookie, sizeof(my_cookie));
    mu_assert("my cookie is gone!", !err);

    vmaf_picture_unref(&pic);
```

PR #1014, opened 2022-04-11
libvmaf: remove deprecated api function compute_vmaf()

kylophone:
`compute_vmaf()` has been marked as deprecated since `v2.0.0` almost 1.5 years. This PR fully removes `compute_vmaf()` from libvmaf. Also the executable `vmafossexec` is finally removed from the build. FFmpeg was the last major consumer of this deprecated API, and now that they have switched over it seems like it a good time to fully remove it. Since this is a non-backwards compatible change, after this is merged the next `libvmaf` release will need to have a major version bump, _i.e._ `v3.0.0`.

There are still a number of failing Python tests which rely on `compute_vmaf()`/`vmafossexec`, so we'll need to figure out what to do with those.


kylophone:
Closed, Duplicate of #1258.

PR #944, opened 2021-11-19
Pipelined vif feature computation: memory requirements reduced and pe…

foosoftsrl:
…rformance improvement

Here I am.

See https://github.com/Netflix/vmaf/pull/943 for context

Have fun

@Kyle: Had a look at the differences as seen from github. Yes the space only modifications are annoying.

PR #1017, opened 2022-04-18
libvmaf: pipelined vif feature computation

kylophone:
This is a cleaned up, rebased and squashed version of #944. This performance bottleneck fix was identified and implemented by @foosoftsrl. Instead of a two-pass operation (filter, stats) now both operations have been pipelined and a costly second pass is not required. This is a significant contribution and the performance results speak for themselves. With AVX2 enabled, there is a **~16.6%** speedup over the master branch. The overall memory use is down by **~45%**. These gains come with no loss of numerical precision or inaccuracies. The following command was used to benchmark these changes. This is a 1920x1080 source with 130 frames. Thank you to @nilfm for the initial clean-up as well as very thoroughly testing the changes.

```
./build/tools/vmaf \
    --reference Sniper_1920x1080P_30fps_8bit.y4m \
    --distorted Sniper_1920x1080P_30fps_8bit.y4m.264.y4m
```
## CPU
| Command | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| `./master.sh` | 13.014 ± 0.058 | 12.936 | 13.138 |
| `./vif_pipeline.sh` | 10.865 ± 0.032 | 10.823 | 10.925 |

## Memory
```
./master.sh: 133234688  maximum resident set size 
./vif_pipeline.sh: 87543808  maximum resident set size
```

## AVX-512
In addition to the AVX2 optimizations benchmarked above, there are also AVX-512 optimizations included in this PR. At the moment there have been some rare mismatches observed. Until those can mismatches can be addressed these optimizations have been disabled by default.

### DIFF
--- a/libvmaf/src/feature/integer_vif.c
+++ b/libvmaf/src/feature/integer_vif.c
@@ -40,16 +40,12 @@
 #endif
 
 typedef struct VifState {
-    VifBuffer buf;
-    uint16_t log2_table[65537];
+    VifPublicState public;
     bool debug;
-    double vif_enhn_gain_limit;
-    void (*filter1d_8)(VifBuffer buf, unsigned w, unsigned h);
-    void (*filter1d_16)(VifBuffer buf, unsigned w, unsigned h, int scale,
-                        int bpc);
-    void (*filter1d_rd_8)(VifBuffer buf, unsigned w, unsigned h);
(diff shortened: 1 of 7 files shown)


kylophone:
Replaces #994 and #996

nilfm:
> Replaces #994 and #996

The #994 should be #944 :)

PR #1024, opened 2022-05-04
rename metrics to features

slhck:
"Metrics" is ambiguous. The term "features" should be consistently used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,13 @@ Also included in `libvmaf` are implementations of several other metrics: PSNR, P
 
 ## Documentation
 
-There is an [overview of the documentation](resource/doc/index.md) with links to specific pages, covering FAQs, available models and metrics, software usage guides, and a list of resources.
+There is an [overview of the documentation](resource/doc/index.md) with links to specific pages, covering FAQs, available models and features, software usage guides, and a list of resources.
 
 ## Usage
 
 The software package offers a number of ways to interact with the VMAF implementation.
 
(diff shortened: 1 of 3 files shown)


li-zhi:
@slhck In my view, a feature is an *element* that can guide the prediction of a *metric*. To put it more concretely, the inputs to VMAF's SVR model are considered as *features*. A metric can be used as a feature for another metric, but not the other way around. In other words: *features* is a superset of *metrics*. Would you agree with this interpretation? If yes, I think it is best to keep the current wording. Happy to discuss.

slhck:
I agree with you in principle.

My change was motivated by somewhat inconsistent usage of that terminology. Consider users coming from the ffmpeg libvmaf filter documentation, which talks about features that can be enabled but does not list them. (The improvement of that documentation itself is a more general issue we have discussed in another GitHub thread.) The documentation for VMAF itself should, in my opinion, make these features more discoverable. Otherwise there is no link between what is written in the ffmpeg documentation and what the user can find here.

How would you suggest to address this issue?

li-zhi:
Thanks for the explanation. With this context, I think I am more convinced that the changes are needed. We just need to have a clarification on the "features is a superset of metrics" aspect. For example, the statement "Features are individual metrics that VMAF models may use to provide a quality score. Features can be enabled individually (e.g., when using VMAF via FFmpeg)." could be updated to "Features are individual inputs that (VMAF) models may use to provide a quality metric score. Note that an existing quality metric can serve as a feature of a new quality metric. Features can be enabled individually (e.g., when using VMAF via FFmpeg)."

slhck:
Thanks, that makes perfect sense. Pushed with that change!

PR #1026, opened 2022-05-16
fix crash when vmaf_picture_unref

yunxuanwang:
some crash happened when use vmaf. seems a double free problem caused by multi-thread. This will ensure VmafPicture only free once.

### DIFF
--- a/libvmaf/test/test_cli_parse.c
+++ b/libvmaf/test/test_cli_parse.c
@@ -23,8 +23,11 @@
 #include "cli_parse.h"
 
 static int cli_free_dicts(CLISettings *settings) {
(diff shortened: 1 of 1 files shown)


kylophone:
Was your crash using the `vmaf` tool? Or are you using `libvmaf` directly? More information about your set up and the crash would be helpful.

yunxuanwang:
> Was your crash using the `vmaf` tool? Or are you using `libvmaf` directly? More information about your set up and the crash would be helpful.

Using libvamf directly. I tried just compute_vmaf or use vmaf_init and other api directly,  both get same crash when vmaf_picture_unref

nilfm:
I think this fix is on the right track, need to rebase it and adapt it to master changes though.

PR #1048, opened 2022-06-25
Add command line option --skip_ref_frame

dr-xiangli:
This command line option allows skipping the first N frame of the reference (source) when calculating VMAF, PSNR and other metrics. This is useful when only a clip of a source is coded and we need to calculate metrics only for the coded clip against the source. As --frame_cnt has been in the package, this PR is to address the case that the clip is not from the beginning of the source.

### DIFF
--- a/libvmaf/tools/cli_parse.c
+++ b/libvmaf/tools/cli_parse.c
@@ -25,8 +25,8 @@ enum {
     ARG_CPUMASK,
     ARG_AOM_CTC,
     ARG_FRAME_CNT,
-    ARG_SKIP_REF_FRAME,
-    ARG_SKIP_DIST_FRAME,
+    ARG_FRAME_SKIP_REF,
+    ARG_FRAME_SKIP_DIST,
 };
 
 static const struct option long_opts[] = {
@@ -48,8 +48,8 @@ static const struct option long_opts[] = {
     { "cpumask",          1, NULL, ARG_CPUMASK },
     { "aom_ctc",          1, NULL, ARG_AOM_CTC },
     { "frame_cnt",        1, NULL, ARG_FRAME_CNT },
-    { "skip_ref_frame",   1, NULL, ARG_SKIP_REF_FRAME },
-    { "skip_dist_frame",  1, NULL, ARG_SKIP_DIST_FRAME },
+    { "frame_skip_ref",   1, NULL, ARG_FRAME_SKIP_REF },
+    { "frame_skip_dist",  1, NULL, ARG_FRAME_SKIP_DIST },
     { "no_prediction",    0, NULL, 'n' },
     { "version",          0, NULL, 'v' },
     { "quiet",            0, NULL, 'q' },
@@ -66,31 +66,31 @@ static void usage(const char *const app, const char *const reason, ...) {
     }
     fprintf(stderr, "Usage: %s [options]\n\n", app);
     fprintf(stderr, "Supported options:\n"
-            " --reference/-r $path:       path to reference .y4m or .yuv\n"
-            " --distorted/-d $path:       path to distorted .y4m or .yuv\n"
-            " --width/-w $unsigned:       width\n"
-            " --height/-h $unsigned:      height\n"
-            " --pixel_format/-p: $string  pixel format (420/422/444)\n"
-            " --bitdepth/-b $unsigned:    bitdepth (8/10/12/16)\n"
-            " --model/-m $params:         model parameters, colon \":\" delimited\n"
-            "                             `path=` path to model file\n"
-            "                             `version=` built-in model version\n"
-            "                             `name=` name used in log (optional)\n"
-            " --output/-o $path:          output file\n"
-            " --xml:                      write output file as XML (default)\n"
-            " --json:                     write output file as JSON\n"
-            " --csv:                      write output file as CSV\n"
-            " --sub:                      write output file as subtitle\n"
-            " --threads $unsigned:        number of threads to use\n"
-            " --feature $string:          additional feature\n"
-            " --cpumask: $bitmask         restrict permitted CPU instruction sets\n"
-            " --frame_cnt $unsigned:      maximum number of frames to process\n"
-            " --skip_ref_frame $unsigned: skip the first N frames in reference\n"
-            " --skip_dist_frame $unsigned:skip the first N frames in distorted\n"
-            " --subsample: $unsigned      compute scores only every N frames\n"
-            " --quiet/-q:                 disable FPS meter when run in a TTY\n"
-            " --no_prediction/-n:         no prediction, extract features only\n"
-            " --version/-v:               print version and exit\n"
+            " --reference/-r $path:        path to reference .y4m or .yuv\n"
+            " --distorted/-d $path:        path to distorted .y4m or .yuv\n"
+            " --width/-w $unsigned:        width\n"
+            " --height/-h $unsigned:       height\n"
+            " --pixel_format/-p: $string   pixel format (420/422/444)\n"
+            " --bitdepth/-b $unsigned:     bitdepth (8/10/12/16)\n"
+            " --model/-m $params:          model parameters, colon \":\" delimited\n"
+            "                              `path=` path to model file\n"
+            "                              `version=` built-in model version\n"
+            "                              `name=` name used in log (optional)\n"
+            " --output/-o $path:           output file\n"
+            " --xml:                       write output file as XML (default)\n"
+            " --json:                      write output file as JSON\n"
+            " --csv:                       write output file as CSV\n"
+            " --sub:                       write output file as subtitle\n"
+            " --threads $unsigned:         number of threads to use\n"
+            " --feature $string:           additional feature\n"
+            " --cpumask: $bitmask          restrict permitted CPU instruction sets\n"
+            " --frame_cnt $unsigned:       maximum number of frames to process\n"
+            " --frame_skip_ref $unsigned:  skip the first N frames in reference\n"
+            " --frame_skip_dist $unsigned: skip the first N frames in distorted\n"
+            " --subsample: $unsigned       compute scores only every N frames\n"
+            " --quiet/-q:                  disable FPS meter when run in a TTY\n"
+            " --no_prediction/-n:          no prediction, extract features only\n"
+            " --version/-v:                print version and exit\n"
            );
     exit(1);
 }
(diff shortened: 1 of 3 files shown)


libvmaf/tools/cli_parse.c
```diff
@@ -46,6 +47,7 @@ static const struct option long_opts[] = {
     { "cpumask",          1, NULL, ARG_CPUMASK },
     { "aom_ctc",          1, NULL, ARG_AOM_CTC },
     { "frame_cnt",        1, NULL, ARG_FRAME_CNT },
+    { "skip_ref_frame",   1, NULL, ARG_SKIP_REF_FRAME },
```
li-zhi, line 50: It would be useful to slightly generalize the logic to allow negative values. For example, if the skip_ref_frame number is -2, it means to skip 2 dis frames instead. 

Once you have the logic ready, I can add two tests to cover both cases.

dr-xiangli: It is a good question. I thought about this (actually to introduce another offset instead of being negative for distorted y4m). But if we assume that the distorted y4m is a decoded or reconstructed version of the reference, we cannot have frames in distorted y4m before the reference since we don't have reference to encode. Or do we have a use case to compare two arbitrary y4ms? If yes, then we need the negative offset.

li-zhi: There were indeed use cases where you may need to skip some distorted frames (think of VMAF is computed on a shot-by-shot basis and for some reason the reference and distorted are a few frames misaligned). It would be useful to cover this case.

By the way, there is another way to achieve the frame skipping using [ffmpeg+libvmaf](https://github.com/Netflix/vmaf/blob/master/resource/doc/ffmpeg.md) combined with the select filter (example: `-vf select='gte(n\,0)*gte(6\,n)'`). I assume here you want to avoid calling ffmpeg.

dr-xiangli: OK. I will add the support of skipping frames in a distorted sequence. I will use negative offset for this case, as you suggested unless you prefer a new parameter.

Regarding ffmpeg, yes, I'd like to avoid the call to ffmpeg. For our tests, we need a temporary y4m file for the conversion (cut a longer sequence to a shorter one and then calculate VMAF/PSNR) if source and distorted have different numbers of frames as the decoder we needed is not embedded in ffmpeg (like the case of AV2). If we are doing a lot of 4K tests at the same time, the peak disk space requirement would be quite huge.

dr-xiangli: Maybe adding another parameter skip_dist_frame is better? If we use negative value to skip distorted frames, we can only have one skip offset, either for reference or for distorted, both not for both. If we add skip_dist_frame in addition to skip_ref_frame, we may have two different non-zero skip offsets at the same time. From the use case you mentioned, two parameters seem better. Your opinion? @li-zhi

li-zhi: This is a good point. It's more general and less confusing. Let's have two parameters.

dr-xiangli: --skip_dist_frame has been added. Please check.

kylophone:
OK, here's my take on this. Options should be called `frame_skip_ref` and `frame_skip_dist`, and there should be a third convenience option `frame_skip` which affects both ref/dist. These options combined with the already existing `frame_cnt` should be enough to grab any set of frames you want.

libvmaf/tools/cli_parse.c
```diff
@@ -62,29 +66,31 @@ static void usage(const char *const app, const char *const reason, ...) {
...
+            " --xml:                      write output file as XML (default)\n"
+            " --json:                     write output file as JSON\n"
+            " --csv:                      write output file as CSV\n"
+            " --sub:                      write output file as subtitle\n"
+            " --threads $unsigned:        number of threads to use\n"
+            " --feature $string:          additional feature\n"
+            " --cpumask: $bitmask         restrict permitted CPU instruction sets\n"
+            " --frame_cnt $unsigned:      maximum number of frames to process\n"
+            " --skip_ref_frame $unsigned: skip the first N frames in reference\n"
+            " --skip_dist_frame $unsigned:skip the first N frames in distorted\n"
```
kylophone, line 89: can you add one more space here

dr-xiangli:
> OK, here's my take on this. Options should be called `frame_skip_ref` and `frame_skip_dist`, and there should be a third convenience option `frame_skip` which affects both ref/dist. These options combined with the already existing `frame_cnt` should be enough to grab any set of frames you want.

Before updating the code, let me fully understand the intended behavior. Will frame_skip override frame_skip_ref and frame_skip_dist if frame_skip is present? If frame_skip, frame_skip_ref, frame_skip_dist have different values, how to skip?

kylophone:
> > OK, here's my take on this. Options should be called `frame_skip_ref` and `frame_skip_dist`, and there should be a third convenience option `frame_skip` which affects both ref/dist. These options combined with the already existing `frame_cnt` should be enough to grab any set of frames you want.
> 
> Before updating the code, let me fully understand the intended behavior. Will frame_skip override frame_skip_ref and frame_skip_dist if frame_skip is present? If frame_skip, frame_skip_ref, frame_skip_dist have different values, how to skip?

The command-line arguments are parsed in order, and they allow overriding by design. You can give the same arg more than once and the last one always wins. The `--frame_skip` arg doesn't need to do anything special, just set both ref and dist. This is mainly for the convenience of not having to set both, as your case is probably the special one.

Renaming the arguments to `--frame_skip_*` is just my preference to match `frame_count` a little bit better.

dr-xiangli:
> > > OK, here's my take on this. Options should be called `frame_skip_ref` and `frame_skip_dist`, and there should be a third convenience option `frame_skip` which affects both ref/dist. These options combined with the already existing `frame_cnt` should be enough to grab any set of frames you want.
> > 
> > 
> > Before updating the code, let me fully understand the intended behavior. Will frame_skip override frame_skip_ref and frame_skip_dist if frame_skip is present? If frame_skip, frame_skip_ref, frame_skip_dist have different values, how to skip?
> 
> The command-line arguments are parsed in order, and they allow overriding by design. You can give the same arg more than once and the last one always wins. The `--frame_skip` arg doesn't need to do anything special, just set both ref and dist. This is mainly for the convenience of not having to set both, as your case is probably the special one.
> 
> Renaming the arguments to `--frame_skip_*` is just my preference to match `--frame_count` a little bit better.

To better understand, let's say we have following command line options, --frame_skip 50 --frame_skip_ref 30 --frame_skip_dist 40 --frame_cnt 100. What is the expected skipping behavior for reference and distorted?

kylophone:
> > > > OK, here's my take on this. Options should be called `frame_skip_ref` and `frame_skip_dist`, and there should be a third convenience option `frame_skip` which affects both ref/dist. These options combined with the already existing `frame_cnt` should be enough to grab any set of frames you want.
> > > 
> > > 
> > > Before updating the code, let me fully understand the intended behavior. Will frame_skip override frame_skip_ref and frame_skip_dist if frame_skip is present? If frame_skip, frame_skip_ref, frame_skip_dist have different values, how to skip?
> > 
> > 
> > The command-line arguments are parsed in order, and they allow overriding by design. You can give the same arg more than once and the last one always wins. The `--frame_skip` arg doesn't need to do anything special, just set both ref and dist. This is mainly for the convenience of not having to set both, as your case is probably the special one.
> > Renaming the arguments to `--frame_skip_*` is just my preference to match `--frame_count` a little bit better.
> 
> To better understand, let's say we have following command line options, --frame_skip 50 --frame_skip_ref 30 --frame_skip_dist 40 --frame_cnt 100. What is the expected skipping behavior for reference and distorted?

reference: 30-130, distorted: 40-140.
 
`--frame_skip` sets both `settings->frame_skip_ref` and `settings->frame_skip_dist`, those values are then overwritten by `--frame_skip_ref ` and `--frame_skip_dist`.

li-zhi:
> The command-line arguments are parsed in order, and they allow overriding by design. You can give the same arg more than once and the last one always wins. The `--frame_skip` arg doesn't need to do anything special, just set both ref and dist. This is mainly for the convenience of not having to set both, as your case is probably the special one.

@kylophone @dr-xiangli  I am debating if this adds too much cognitive load for the user unnecessarily. Since the same functionalities can be accommondated by ffmpeg, here we only need to address Google's special case. I am leaning toward going back to the original proposal of adding a single parameter `--frame_skip_ref` only.

kylophone:
> > The command-line arguments are parsed in order, and they allow overriding by design. You can give the same arg more than once and the last one always wins. The `--frame_skip` arg doesn't need to do anything special, just set both ref and dist. This is mainly for the convenience of not having to set both, as your case is probably the special one.
> 
> @kylophone @dr-xiangli I am debating if this adds too much cognitive load for the user unnecessarily. Since the same functionalities can be accommondated by ffmpeg, here we only need to address Google's special case. I am leaning toward going back to the original proposal of adding a single parameter `--frame_skip_ref` only.

Too arbitrary to add one and not the other, IMO. Either we accommodate general purpose flexibility, or we just avoid it all together.

dr-xiangli:
> > > The command-line arguments are parsed in order, and they allow overriding by design. You can give the same arg more than once and the last one always wins. The `--frame_skip` arg doesn't need to do anything special, just set both ref and dist. This is mainly for the convenience of not having to set both, as your case is probably the special one.
> > 
> > 
> > @kylophone @dr-xiangli I am debating if this adds too much cognitive load for the user unnecessarily. Since the same functionalities can be accommondated by ffmpeg, here we only need to address Google's special case. I am leaning toward going back to the original proposal of adding a single parameter `--frame_skip_ref` only.
> 
> Too arbitrary to add one and not the other, IMO. Either we accommodate general purpose flexibility, or we just avoid it all together. Note that this is not a `libvmaf` thing, it only applies to the `vmaf` tool.

Then we keep two, --frame_skip_ref and --frame_skip_dist?

li-zhi:
> Then we keep two, --frame_skip_ref and --frame_skip_dist?

I think this is a better tradeoff. We lose some convenience of `frame_skip`, but that should be ok.

libvmaf/tools/cli_parse.c
dr-xiangli: Added.

dr-xiangli:
> > Then we keep two, --frame_skip_ref and --frame_skip_dist?
> 
> I think this is a better tradeoff. We lose some convenience of `frame_skip`, but that should be ok.

Variable names have been updated to frame_skip_ref and frame_skip_dist

li-zhi:
#1052 adds tests for frame_skip_ref and frame_skip_dist arguments.

PR #1055, opened 2022-07-06
Address a couple valgrind issues, use after free in predict.c::vmaf_p…

nvanderzwet-ssimwave:
Address a couple valgrind issues
- use after free in predict.c::vmaf_predict_score_at_index
- uninitialized memory in integer_vif.c::init

I am confident there are quit a few more (seems to be a log of allocs without memsets), these are just the ones exercised during my test

### DIFF
--- a/libvmaf/src/predict.c
+++ b/libvmaf/src/predict.c
@@ -284,14 +284,15 @@ int vmaf_predict_score_at_index(VmafModel *model,
         err = vmaf_feature_collector_get_score(feature_collector,
                                                feature_name, &feature_score,
                                                index);
-        free(feature_name);
 
         if (err) {
             vmaf_log(VMAF_LOG_LEVEL_ERROR,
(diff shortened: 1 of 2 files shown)


libvmaf/src/predict.c
```diff
@@ -284,14 +284,15 @@ int vmaf_predict_score_at_index(VmafModel *model,
...
-        free(feature_name);
 
         if (err) {
             vmaf_log(VMAF_LOG_LEVEL_ERROR,
                      "vmaf_predict_score_at_index(): no feature '%s' "
                      "at index %d\n", feature_name, index);
+            free(feature_name);
             goto free_node;
         }
+        free(feature_name);
```
kylophone, line 295: Would be better if this was under `free_node` instead.

nvanderzwet-ssimwave: This one's in a for-loop of all model->n_features, so it needs to be inside the loop

PR #1072, opened 2022-07-28
test_model: fix out of tree testing

gizahNL:
Fixes testing out of tree (i.e.: when builddir is not subdir of libvmaf dir) by defining the path to the model dir contained in vmaf project relative to meson.project_source_root.


kylophone:
Yes, I think this is good. One other alternative would be to move all of our models directly into the libvmaf tree, but I will need to check with a few folks on that first.

gizahNL:
> Yes, I think this is good. One other alternative would be to move all of our models directly into the libvmaf tree, but I will need to check with a few folks on that first.

Yeah the path containing "../" feels a bit unpleasant ;) 

Moving the model folder might break other peoples processes though. Keeping a symlink in git with a comment that that path is deprecated might be needed then. Not sure though how that would behave on Windows

PR #1086, opened 2022-08-23
Dynamic choice of GPU/CPU processing 

gedoensmax:
This PR solves:
- When building with CUDA support one can still use CPU feature extractors if `vmaf_cuda_init` was not called. Note that `vmaf_cuda_init` has to be called before querying feature extractor !
- Renaming `cuda.h`
- Adding more invalid value checks inside `ring_buffer.c` and cuda `common.h`

### DIFF
--- a/libvmaf/src/libvmaf.c
+++ b/libvmaf/src/libvmaf.c
@@ -48,6 +48,7 @@
 
 #ifdef HAVE_CUDA
 #include "libvmaf/vmaf_cuda.h"
+#include "libvmaf/vmaf_cuda_state.h"
 #include "cuda/cuda_helper.cuh"
 #include "cuda/picture_cuda.h"
 #include "cuda/common.h"
--- a/libvmaf/src/libvmaf.c
+++ b/libvmaf/src/libvmaf.c
@@ -306,11 +306,8 @@ int vmaf_use_features_from_model(VmafContext *vmaf, VmafModel *model)
 
     for (unsigned i = 0; i < model->n_features; i++) {
         VmafFeatureExtractor *fex =
-#ifdef HAVE_CUDA
-            vmaf_get_feature_extractor_by_feature_name(model->feature[i].name, vmaf->cuda.state.ctx);
-#else
-            vmaf_get_feature_extractor_by_feature_name(model->feature[i].name, 0);
-#endif
+
+        vmaf_get_feature_extractor_by_feature_name(model->feature[i].name, vmaf->cfg.gpumask);
         if (!fex) {
             vmaf_log(VMAF_LOG_LEVEL_ERROR,
                      "could not initialize feature extractor \"%s\"\n",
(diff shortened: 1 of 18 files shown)


kylophone:
> * When building with CUDA support one can still use CPU feature extractors if `vmaf_cuda_init` was not called. Note that `vmaf_cuda_init` has to be called before querying feature extractor !

I think it can actually be more implicit than that. `vmaf_read_pictures()` can take of any `enum VmafPictureBufferType` and auto conversion could take place based on which `enum VmafFeatureExtractorFlags` are set in the feature extractor. I'd prefer a gpumask to disable GPU at runtime if so desired. 

> * Renaming `cuda.h`

Didn't see this in time, so I took care of it in [8ab1119](https://github.com/Netflix/vmaf/pull/1080/commits/8ab111986e3ff50620a3d449e4a5a2e04c158344)

> * Adding more invalid value checks inside `ring_buffer.c` and cuda `common.h`

OK

gedoensmax:
I rebased the branch to remove duplicated renaming.

libvmaf/src/libvmaf.c
```diff
@@ -302,7 +306,11 @@ int vmaf_use_features_from_model(VmafContext *vmaf, VmafModel *model)
 
     for (unsigned i = 0; i < model->n_features; i++) {
         VmafFeatureExtractor *fex =
-            vmaf_get_feature_extractor_by_feature_name(model->feature[i].name, 1);
+#ifdef HAVE_CUDA
+            vmaf_get_feature_extractor_by_feature_name(model->feature[i].name, vmaf->cuda.state.ctx);
+#else
+            vmaf_get_feature_extractor_by_feature_name(model->feature[i].name, 0);
+#endif
```
gedoensmax, line 313: This does prevent a lazy cuda initialization. Cuda has to be initialized before selecting feature extractors.
@kylophone do you have an idea how to do this more elegant ? I would propose to add something like the `cpumask` to vmaf configuration struct which we could check here and could even throw as error when `vmaf_read_picture` is called with this config flag set, but `vmaf_cuda_init` not being called up to this point.

kylophone: > This does prevent a lazy cuda initialization. Cuda has to be initialized before selecting feature extractors.
> @kylophone do you have an idea how to do this more elegant ? I would propose to add something like the `cpumask` to vmaf configuration struct which we could check here and could even throw as error when `vmaf_read_picture` is called with this config flag set, but `vmaf_cuda_init` not being called up to this point.

Yes, I think something like that makes sense. Instead of matching on feature extractor name exactly, we should be matching on some sort of combination of hardware capabilities and gpumask.

gedoensmax: Roman proposed to make `vmaf_init` and `vmaf_cuda_init` mutually exclusive and set an internal state depending on this. Would you be ok with this too ?

PR #1088, opened 2022-08-26
fix memory leak by freeing frame when one stream ends before the other

cosmin:
During the process of developing and testing another feature extractor we encountered a memory leak in valgrind in cases where two inputs have unequal number of frames.

Co-authored-by: RakshithGB <luckydudefeelinggreat@gmail.com>


kylophone:
Looks good, thank you.

PR #1116, opened 2022-11-14
CUDA Driver API usage & dropping CUDART

MorkTheOrk:
This PR removes of the CUDA runtime dependency and usage when using libvmaf but still requires CUDA Toolkit to compile .cu files.

Each .cu file is compiled using nvcc to a .ptx file which is then being converted into a unsigned byte array using xxd. The resulting byte array is then linked into libvmaf. The needed kernels are loaded during the init function of each feature extractor. 
Due to C++ name mangling we needed to use defines to generate the templated kernels with a fixed name so we can load them manually.


kylophone:
I noticed a build failure if `-Denable_nvtx=false` is set.

PR #1120, opened 2022-11-22
libvmaf/cuda: gpumask logic, enable parallel cpu/gpu feature extraction

kylophone:
This PR makes it so `VmafConfiguration.gpumask` mask operates in a similar fashion to `VmafConfiguration.cpumask`. If `libvmaf` is compiled with CUDA support, then GPU feature extraction will be enabled by default. If you would like to fall back to CPU-only feature extraction, then you may set `VmafConfiguration.gpumask` to something non-zero. The command-line flag `--cuda` has been removed, and `--gpumask` has been added.

Additionally, this PR enables parallel feature extraction, it is now possible to compute vmaf on the GPU and other metrics (`psnr`, `cambi`, ...) on the CPU at the same time. The library can take either host or device pictures as input and will make copies as necessary. Implementation is still not 100% where it should be (more to come), but this small change to the API behavior should hopefully be final.

For a usage example demonstrating these changes, see: `libvmaf/tools/test/test_vmaf_cuda_gpumask.sh`.

### DIFF
--- a/libvmaf/src/feature/feature_extractor.c
+++ b/libvmaf/src/feature/feature_extractor.c
@@ -26,6 +26,10 @@
 #include "feature_name.h"
 #include "log.h"
 
+#ifdef HAVE_NVTX
+#include "nvtx3/nvToolsExt.h"
+#endif
+
 #if VMAF_FLOAT_FEATURES
 extern VmafFeatureExtractor vmaf_fex_float_psnr;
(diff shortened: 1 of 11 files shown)


kylophone:
I will need to fix the cpu-only build.

kylophone:
fyi @gedoensmax

gedoensmax:
Sorry for the delay ! I tested and it looks good to me!

PR #1130, opened 2022-12-01
fix vif_avx512 bug; enable avx512 for vif

nilfm:
Context: there was a numerical discrepancy in the AVX-512 implementation of VIF that led us to temporarily disable it. This PR fixes it and re-enables AVX-512 for VIF.


kylophone:
Merged!

PR #1152, opened 2023-01-07
CUDA feature extraction

kylophone:
Replaces #1080.


HunterAP23:
I know this was merged, but is there any guide on how to build libvmaf for Windows under MINGW64 MYSYS2 with CUDA? I'm not able to find any method of putting CUDA on MINGW64 MYSYS2, so building this with CUDA enabled for Windows doesn't seem possible at the moment, unless there's some alternative method of doing so that isn't documented.

For reference, I've also tried Ubuntu through WSL2 but that seems to have a lot of errors, and it's not possible to cross-compile from there anyways.

kylophone:
I have not personally tried this with Windows yet, I have been developing on Linux only. Could you please make an issue, we should be able to support it.

ehmkhay:
Does this PR implement CUDA pipeline to ffmpeg -lavfi for in-GPU nvdec>vmaf (in linux)?
(or is this still in development?)
Thanks!

kylophone:
> Does this PR implement CUDA pipeline to ffmpeg -lavfi for in-GPU nvdec>vmaf (in linux)? (or is this still in development?) Thanks!

That work is in progress, you should see it on `ffmpeg-devel` soon.

ehmkhay:
Brilliant, thanks!

maxpain:
How to use cuda acceleration in vmaf today?

Issue #1155, opened 2023-01-12
adm_cuda: high bitdepth crash


kylophone:
`adm_cuda` feature extractor is crashing with high bitdepth input, looks like a memory issue.

The issue is reproducible with the following command:
```sh
./build/tools/vmaf \
    --reference /dev/zero \
    --distorted /dev/zero \
    -w 1920 -h 1080 -b 10 -p 420 \
    --frame_cnt 10 \
    --feature adm_cuda -n
```

Output:
```
VMAF version ad468e0e
1 frame  ⢀⠀ 0.00 FPScode: 700; description: CUDA_ERROR_ILLEGAL_ADDRESS
vmaf: ../src/feature/cuda/integer_adm_cuda.c:914: integer_compute_adm_cuda: Assertion `0' failed.
./run.sh: line 49: 1023085 Aborted                 ./build_cuda/tools/vmaf --reference /dev/zero --distorted /dev/zero -w 1920 -h 1080 -b 10 -p 420 --frame_cnt 10 --feature adm_cuda -n
```

Backtrace from `cuda-gdb`:
```sh
Starting program: /root/vmaf_private/libvmaf/build/tools/vmaf --reference /dev/zero --distorted /dev/zero -w 1920 -h 1080 -b 10 -p 420 --frame_cnt 10 --feature adm_cuda -n
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
VMAF version ad468e0e
[Detaching after fork from child process 1023077]
[New Thread 0x7fffef451700 (LWP 1023081)]
[New Thread 0x7fffeec50700 (LWP 1023082)]
1 frame  ⢀⠀ 0.00 FPS
CUDA Exception: Warp Out-of-range Address
The exception was triggered at PC 0x555555c70f10

Thread 1 "vmaf" received signal CUDA_EXCEPTION_5, Warp Out-of-range Address.
[Switching focus to CUDA kernel 0, grid 1, block (0,0,0), thread (64,1,0), device 0, sm 0, warp 6, lane 0]
0x0000555555c70f50 in adm_dwt2_8_vert_hori_kernel_16_32768_4_16_32768_128_8_uint16_t
   <<<(16,68,1),(128,2,1)>>> ()
```

kylophone:
@gedoensmax

MorkTheOrk:
@kylophone ill take a look tomorrow.

kylophone:
[...]

PR #1160, opened 2023-01-16
fixed device pointer argument

MorkTheOrk:
This fixes issue #1155 
Wrong pointer was passed to the GPU causing a crash in a ADM_DWT kernel for 16 bits.


kylophone:
Thank you!

PR #1163, opened 2023-01-24
Fix: Missing alias on cuda options

gedoensmax:
I found that when running with the "neg" model the feature names are not correctly propagated due to having different options on CUDA and normal feature extractors. 
```bash
$ vmaf -r data/distorted_1080p_yuv420p.yuv -d data/reference_1080p_yuv420p.yuv -w 1080 -h 1920 -p 420 -b 8 -m path=./model/vmaf_v0.6.1neg.json --gpumask 0 
VMAF version b9150a23
128 frames ⠋⠉ 273.88 FPS
libvmaf ERROR vmaf_predict_score_at_index(): no feature 'integer_adm2_egl_1' at index 0
problem generating pooled VMAF score
```
This PR fixes that but i am not sure if it would be more optimal to move the options for adm into `integer_adm.h` and "template" them with `AdmState` and `AdmStateCuda` due to different offsets. Currently it looks like a dangerous code duplication of the option definition.
@kylophone What do you think of that?

### DIFF
--- a/libvmaf/src/feature/cuda/integer_adm_cuda.c
+++ b/libvmaf/src/feature/cuda/integer_adm_cuda.c
@@ -596,6 +596,7 @@ static const VmafOption options_cuda[] = {
     },
     {
         .name = "adm_enhn_gain_limit",
+        .alias = "egl",
         .help = "enhancement gain imposed on adm, must be >= 1.0, "
             "where 1.0 means the gain is completely disabled",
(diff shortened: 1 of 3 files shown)


kylophone:
Thanks, yes I think we should try to consolidate as much as possible between the CUDA and CPU feature extractors. For CPU optimizations we are able to set function pointers during init, something similar should be possible for GPU optimizations.

PR #1171, opened 2023-02-27
add dis_enc_bitdepth to CambiFeatureExtractor

krasuluk:
This PR fixes a problem in ``CambiFeatureExtractor`` for cases where the encode's bitdepth is lower than the workfile's bitdepth. The previous version did not propagate ``enc_bitdepth`` to the executable which caused CAMBI not to compensate for the conversions from the original format to the workfile format (similarly as it compensates for scaling via ``enc_width`` and ``enc_height``). This led to lowering the CAMBI scores for 8bit videos with 10bit references when calculated via VMAF Python Library.

The PR consists of:

- adding ``dis_encode_bitdepth`` as a property of ``Asset``. The logic is to automatically assign this property for supported yuv files and return None for notyuv when not specified.
- adding the requirement for ``dis_encode_bitdepth`` to be specified (similarly to ``dis_encode_width_height``) to ``CambiFeatureExtractor``
- covering the changes with test cases
- bumping up the VERSION of ``CambiFeatureExtractor`` to 0.5

### DIFF
--- a/python/vmaf/core/asset.py
+++ b/python/vmaf/core/asset.py
@@ -12,7 +12,8 @@ import os
 
 from vmaf.core.mixin import WorkdirEnabled
 from vmaf.tools.misc import get_file_name_without_extension, \
-    get_file_name_with_extension, get_unique_str_from_recursive_dict
+    get_file_name_with_extension, get_unique_str_from_recursive_dict, \
+    map_yuv_type_to_bitdepth
 from vmaf.config import VmafConfig
 from vmaf.core.proc_func import proc_func_dict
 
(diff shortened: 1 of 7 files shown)


python/vmaf/core/asset.py
```diff
@@ -466,6 +489,10 @@ def dis_str(self):
             w, h = self.dis_encode_width_height
             s += "_e_{w}x{h}".format(w=w, h=h)
 
+        if self.dis_encode_bitdepth:
```
li-zhi, line 492: I would prefer `if self.dis_encode_bitdepth is not None`, because `if self.dis_encode_bitdepth` has ambiguity: it could be that `if self.dis_encode_bitdepth is not None` or `if self.dis_encode_bitdepth != 0`.

li-zhi (approved):

kylophone:
lgtm, please squash before merge.

PR #1174, opened 2023-03-14
Fix: CUDA+multithreading

gedoensmax:
I noticed some problems when running a CUDA build with multithreading enabled. Even if not CUDA extractors are enabled multithreading is disabled. 
I tested with below command and appended a matrix of options:
`--feature psnr --feature float_ssim` - CPU only extractors 
`--threads $THREADS` - THREADS [0, 1, 16]
`--gpumask $CU_DISABLED` - CU_DISABLED [0, 1]
```
./libvmaf/build/tools/vmaf --reference $file_ref --distorted $file_dis \
    --width $var_width --height $var_height --pixel_format $var_pfmt --bitdepth $var_bpc \
    -o $file2 --json 
```

Following errors occured:
- CUDA build did not use any mutlithreading even with CUDA disabled
- mixing CUDA+CPU extractors forces to run in a single thread

### DIFF
--- a/libvmaf/src/libvmaf.c
+++ b/libvmaf/src/libvmaf.c
@@ -385,6 +385,8 @@ static int threaded_read_pictures(VmafContext *vmaf, VmafPicture *ref,
     for (unsigned i = 0; i < vmaf->registered_feature_extractors.cnt; i++) {
         VmafFeatureExtractor *fex =
             vmaf->registered_feature_extractors.fex_ctx[i]->fex;
+        if (fex->flags & VMAF_FEATURE_EXTRACTOR_CUDA)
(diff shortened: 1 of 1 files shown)


kylophone:
Nice, thank you. I will have a closer look at this later today. Just FYI, the threadpool implementation in libvmaf is likely to be replaced very soon with something that scales better across many threads. I think you have seen some benchmarks which show the problem.

gedoensmax:
Friendly reminder as this actually fixes hard crashes.

kylophone:
> Friendly reminder as this actually fixes hard crashes.

Thanks for the ping. Will review tomorrow.

PR #1177, opened 2023-03-29
fix integer_adm_cuda for bit depth > 8

JamesHutch:
I found incorrect VMAF scores for 10-bit content when enabling CUDA. Tracked it down to incorrect calculation of the ADM feature.

Shift in scale 0 used to assume bit depth of 16, this is now variable.
A typo in the call to adm_dwt2_16_device caused i4_ref_dwt2 not to be written.

### DIFF
--- a/libvmaf/src/feature/cuda/integer_adm/adm_dwt2.cu
+++ b/libvmaf/src/feature/cuda/integer_adm/adm_dwt2.cu
@@ -142,12 +142,13 @@ __device__ __forceinline__ void dwt_s123_combined_hori_kernel(cuda_i4_adm_dwt_ba
     i4_dwt2.band_d[i * dst_stride + idx] = (int32_t)((accum + add_shift) >> shift);
 }
 
-template<int16_t v_shift, int32_t v_add_shift, int v_rows_per_thread, int32_t h_shift, int32_t h_add_shift, int32_t tile_width, int tile_height, typename T>
+template<int v_rows_per_thread, int32_t h_shift, int32_t h_add_shift, int32_t tile_width, int tile_height, typename T>
 __device__ __forceinline__ void adm_dwt2_8_vert_hori_kernel(const T * d_picture,
         cuda_adm_dwt_band_t dst,
         cuda_i4_adm_dwt_band_t i4_dwt2,
         int w, int h,
         int src_stride, int dst_stride,
+        int16_t v_shift, int32_t v_add_shift,
         AdmFixedParametersCuda params) {
 
     // The fused kernel writes the result of the vertical dwt2 to shared memory for consumption by the horizontal dwt2.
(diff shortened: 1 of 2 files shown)


libvmaf/src/feature/cuda/integer_adm/adm_dwt2.cu
```diff
@@ -316,25 +317,26 @@ __device__ __forceinline__ void adm_dwt2_8_vert_hori_kernel(const T * d_picture,
...
+            int16_t v_shift, int32_t v_add_shift,                                                                                                                      \
             AdmFixedParametersCuda params) {                                                                                                                           \
-        adm_dwt2_8_vert_hori_kernel<v_shift, v_add_shift, v_rows_per_thread, h_shift, h_add_shift, tile_width, tile_height, type>(                                     \
+        adm_dwt2_8_vert_hori_kernel<v_rows_per_thread, h_shift, h_add_shift, tile_width, tile_height, type>(                                                           \
                 d_picture, dst, i4_dwt2,                                                                                                                               \
-                w, h, src_stride, dst_stride, params);                                                                                                                 \
+                w, h, src_stride, dst_stride, v_shift, v_add_shift, params);                                                                                           \
     }
```
gedoensmax, line 332: Is it necessary to move those values to an argument instead of a template argument ? It should not make a major impact though.

JamesHutch: v_shift should be set to the bit depth of the input and v_add_shift is 1 << (bit_depth - 1). We could have a function per common bit depth (10,12,16) but it seemed simpler to pass as an argument. There might be a trade off I'm missing though, would be very interested if there is!

JamesHutch: That makes sense. I'll run some experiments to see if there's any performance impact.

JamesHutch: I've run a benchmark of 240 frames of 3840x2160 10 bit content. On a 3060Ti I get a mean of 6.623s with the current pull request and 6.617s with the new arguments removed and hard coded.

I think the current pull request is a little simpler than the optimisation, so I'd suggest we keep the pull request as is. What do you think?

vibhoothi:
I can confirm that with this patch, 10bit video VMAF computation is virtually the same as the CPU variant. 

CPU variant. 
```
VMAF version 5ee0051c
130 frames ⠉⠙ 3.87 FPS
vmaf_v0.6.1: 22.051488
```

Current Master with CUDA:
```
VMAF version 5ee0051c
130 frames ⠉⠙ 0.23 FPS
vmaf_v0.6.1: 62.470298
vmaf_v0.6.1neg: 62.261970
```

With this path (after rebasing a manual rebase to upstream/master)
```
VMAF version 13d115c8
130 frames ⠉⠙ 42.98 FPS
vmaf_v0.6.1: 22.051486
```

Test stream I used: https://media.xiph.org/video/aomctc/test_set/a1_4k/NocturneDance_3840x2160p_10bit_60fps.y4m

gedoensmax:
This is superset by #1195. I cherry picked your changes @JamesHutch and rebased onto the newer CUDA api. Are you fine with closing this ?

JamesHutch:
Sure, thanks for doing that. I haven't had enough time to look at this recently.

Issue #788, opened 2020-12-27
libvmaf won't statically build into ffmpeg


Provissy:
## Description
If build ffmpeg with libvmaf, libvmaf won't be compiled into ffmpeg staticlly. Instead, ffmpeg still needs libvmaf.so.1 as runtime dependency.

## How to reproduce
Build ffmpeg statically in a container.
```
# build libvmaf
meson build --default-library static --buildtype release
ninja -C build
ninja -C build install

# build ffmpeg statically 
PKG_CONFIG_PATH="/usr/local/lib64/pkgconfig:/usr/local/lib/pkgconfig" ./configure \
  --pkg-config-flags="--static" \
  --extra-cflags="-I/usr/local/include" \
  --extra-ldflags="-L/usr/local/lib" \
  --extra-ldflags="-L/usr/local/lib64" \
  --extra-libs=-lpthread \
  --extra-libs=-lm \
  --enable-gpl \
  --enable-libvmaf
make -j$(nproc)
make install
```
Now get the ffmpeg out from the container, run it, the result is 
```./ffmpeg: error while loading shared libraries: libvmaf.so.1: cannot open shared object file: No such file or directory```

## Note
Might be related to **https://github.com/Netflix/vmaf/issues/753#issuecomment-742002566**  https://github.com/Netflix/vmaf/issues/557

**Comment by 1480c1** [comment]
Can you try patching vmaf with https://github.com/Netflix/vmaf/pull/768? you can wget  `https://github.com/Netflix/vmaf/pull/768.patch` and using git apply with it

**Comment by Provissy** [comment]
> Can you try patching vmaf with #768? you can wget `https://github.com/Netflix/vmaf/pull/768.patch` and using git apply with it

Hmm, after patched, `./configure` of ffmpeg gives this

```ERROR: libvmaf >= 1.5.2 not found using pkg-config```

No error if not patched.

More info
```
[root@3863aa09c895 libvmaf]# meson build --default-library static --buildtype release -Denable_tests=false -Denable_docs=
false
The Meson build system
Version: 0.49.2
Source dir: /root/vmaf/libvmaf
Build dir: /root/vmaf/libvmaf/build
Build type: native build
Project name: libvmaf
Project version: 2.0.0
Appending CFLAGS from environment: '-march=native'
Native C compiler: /usr/bin/clang (clang 10.0.1 "clang version 10.0.1 (Red Hat 10.0.1-1.module+el8.3.0+7459+90c24896)")
[...]

PR #1178, opened 2023-04-02
Fix FFmpeg configuration issue for adding -lstdc++ in meson build file

MartinEesmaa:
This will add libraries_private in libvmaf/src/meson.build to fix FFmpeg configuration issue.
I checked with FFmpeg configuration after -lstdc++ is added to libs.private in pkgconfig with ninja install and works well.

- Martin Eesmaa


kylophone:
More context is needed. Can you please describe the "FFmpeg configuration issue"?

MartinEesmaa:
Hi, @kylophone! I can describe about FFmpeg configuration issue.

When I finished compiling libvmaf and installed my prefix folder, then I head to FFmpeg to type command example like `./configure --enable-libvmaf --enable-static --extra-ldexeflags=-static --pkg-config-flags=-static --disable-w32threads` and then FFmpeg doesn't recognize libvmaf means not found using pkg-config, so here's the error message and ffbuild config log of five last lines:
```
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../lib\libvmaf.a(svm.cpp.obj):svm.cpp:(.rdata$.refptr._ZNSt13runtime_errorD1Ev[.refptr._ZNSt13runtime_errorD1Ev]+0x0): undefined reference to `std::runtime_error::~runtime_error()'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../lib\libvmaf.a(svm.cpp.obj):svm.cpp:(.rdata$.refptr._ZTVSt15basic_streambufIcSt11char_traitsIcEE[.refptr._ZTVSt15basic_streambufIcSt11char_traitsIcEE]+0x0): undefined reference to `vtable for std::basic_streambuf<char, std::char_traits<char> >'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/12.2.0/../../../../lib\libvmaf.a(svm.cpp.obj):svm.cpp:(.rdata$.refptr._ZTVNSt7__cxx1115basic_stringbufIcSt11char_traitsIcESaIcEEE[.refptr._ZTVNSt7__cxx1115basic_stringbufIcSt11char_traitsIcESaIcEEE]+0x0): undefined reference to `vtable for std::__cxx11::basic_stringbuf<char, std::char_traits<char>, std::allocator<char> >'
collect2.exe: error: ld returned 1 exit status
ERROR: libvmaf >= 2.0.0 not found using pkg-config
```

The solution is I added -lstc++ in libvmaf/src/meson.build to fix FFmpeg configuration problem, so the FFmpeg can detect libvmaf operation well using additional command of `-lstdc++`, otherwise without -lstdc++ will fail.

Some peoples were having issue to compile FFmpeg with libvmaf static library, the issue is there:
#788

If you have any question or issues, please feel reply to me.
Thank you! :)

- Martin Eesmaa

1480c1:
A note of two things:

1. Another possible solution, as mentioned in the issue, is to use the c++ compiler as the linker so that the appropriate standard c++ library gets linked,  so --ld=g++ or clang++, but that's impossible to express from pkg config as far as I know.
2. Adding -lstdc++ to the pkg config file could be wrong in the case that the user of the library compiled it against llvm's libc++ rather than libstdc++ <https://libcxx.llvm.org/UsingLibcxx.html>. I am not sure of the top of my head what would be the best way to detect which libc to use other than perhaps detecting `-lc++` or `-stdlib=libc++` from cflags.

MartinEesmaa:
@1480c1, yeah you're probably right. I just tried with libc++ command in my GCC compiler and won't work. libc++ is the only can work for clang compiler and the gcc is libstdc++. I was wondering, if we can detect two compilers of GCC and Clang in meson.build, let's just say...

If compiler is GCC, it should go -lstdc++ with pre-pkgconfig. If compiler is clang, it should go -lc++ with pre-pkgconfig.

- Martin Eesmaa

1480c1:
> @1480c1, yeah you're probably right. I just tried with libc++ command in my GCC compiler and won't work. libc++ is the only can work for clang compiler and the gcc is libstdc++. I was wondering, if we can detect two compilers of GCC and Clang in meson.build, let's just say...
> 
> If compiler is GCC, it should go -lstdc++ with pre-pkgconfig. If compiler is clang, it should go -lc++ with pre-pkgconfig.
> 
> - Martin Eesmaa

It is technically possible to use libc++ and g++ as noted in llvm's page
```bash
g++ -nostdinc++ -nodefaultlibs \
    -isystem <install>/include/c++/v1 \
    -L <install>/lib \
    -Wl,-rpath,<install>/lib \
    -lc++ -lc++abi -lm -lc -lgcc_s -lgcc \
    test.cpp
```
But it looks impossibly hard, so I think at least g++ == libstdc++ is a good assumption to start with.
However, I don't think clang++ == libc++ is a good one as on linux systems and on windows with mingw-w64, it uses libstdc++ by default. 

You could try checking if the cxx_link_args variable contains the strings I mentioned. That's the best I got for now until I wake up in the morning.

MartinEesmaa:
Please check my new commit.

Thank you. :)
- Martin Eesmaa

PR #1189, opened 2023-06-06
Fix memory leak: Free data when cleaning up feature extractor

freqmod:
Hi

While doing release testing for  Bridgetech probes we found a memory leak when deinitializing the vmaf feature extractor. This can be verified in valgrind by initializing a vmaf scorer, scoring an image and cleaning up. Valgrind will then report that some memory is leaked. 

This PR contains a fix that fixes the leakage for us.  It seems like a free function call was lost during an earlier merge, so this PR adds the free back, so the code looks similar to the other feature extractors.


kylophone:
Thank you, merged.

PR #1194, opened 2023-07-06
Feature/cambi qrunner version

krasuluk:


nilfm:
@krasuluk If you don't mind rebasing this on master, we can go ahead and merge it when the build passes. The test that you fixed is already fixed in master, so you can leave that out.

PR #1200, opened 2023-07-25
Fix edge cases with AVX implementations of vif

nilfm:
In cases where the input width is not a multiple of 16, there are issues with the AVX-2 and AVX-512 implementations of VIF, leading to small numerical discrepancies or even a segmentation fault. This PR addresses these issues.


kylophone:
lgtm

christosbampis (approved):

PR #1204, opened 2023-07-27
Update build envs to ubuntu-latest and mac-latest.

christosbampis:

PR #1208, opened 2023-07-28
Update os from no-longer-supported ubuntu-18.04 to ubuntu-19.04.

li-zhi:


nilfm:
FYI this PR is doing something very similar: https://github.com/Netflix/vmaf/pull/1204

PR #1232, opened 2023-09-05
Docker + CUDA documentation

gedoensmax:
Actually the base image is a bit overkill. CUDA Toolkit is only required for compilation but not for execution. It should be sufficient to compile in a CUDA docker but copy everything over to the respective Ubuntu base image. Any help with reducing the image into a build image and a final image would be much appreciated ! If image size is not a concern this is working as is.

### DIFF
--- a/Dockerfile.cuda
+++ b/Dockerfile.cuda
@@ -7,12 +7,7 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update && DEBIAN_FRONTEND=noninteract
 
 RUN git clone https://github.com/Netflix/vmaf.git
 
-# ffmpeg patch from http://ffmpeg.org/pipermail/ffmpeg-devel/2023-August/313073.html
-RUN curl -LJ -o vf_libvmaf_cuda.diff https://ffmpeg.org/pipermail/ffmpeg-devel/attachments/20230807/29cf249a/attachment.obj
-
 RUN git clone https://github.com/FFmpeg/FFmpeg.git
-RUN cd FFmpeg && \
-    git apply --ignore-whitespace ../vf_libvmaf_cuda.diff
 
 RUN git clone https://github.com/FFmpeg/nv-codec-headers.git && cd nv-codec-headers && make && make install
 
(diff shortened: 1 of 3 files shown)


kylophone:
I've been thinking that it might be easier to use ffnvcodec inside libvmaf rather than maintain a special Dockerfile for building this. I would like to continue being cross-platform as possible, meaning in addition to Linux, we still have support for Windows, FreeBSD, etc. Thoughts?

gedoensmax:
While I agree with you this is much more work that simply doing a separate Dockerfile. I am not sure when or if I can get to do this. 
Let's leave this PR as is to serve as a public documentation until then.
When using the ffnvcodec I would also recommend doing a 2 stage build - compiling inside a CUDA devel container and then copying the binaries over. CUDA container are not small by any means ...

Dockerfile.cuda
```diff
@@ -0,0 +1,43 @@
...
+    --enable-nvdec \
+    --enable-nvenc \
+    --enable-cuvid \
+    --enable-cuda \
+    --enable-cuda-nvcc \
+    --enable-libvmaf \
+    --enable-ffnvcodec \
+    --disable-stripping \
+    --extra-cflags="-I/usr/local/cuda/include" \
+    --extra-ldflags="-L/usr/local/cuda/lib64 -L/usr/local/cuda/lib64/stubs/"
```
gedoensmax, line 33: Linking against stubs as during build the driver is not available.

kylophone:
@gedoensmax I will merge this now, and we can look at non-Docker alternatives in the future. Anything else you want to add before I merge?

gedoensmax:
@MorkTheOrk please take a look if you think something is missing otherwise I am happy with it.

libvmaf/README.md
```diff
@@ -30,7 +30,11 @@ Run:
 meson build --buildtype release
 ```
 
-(add `-Denable_float=true` flag in the rare case if you want to use the floating-point feature extractors.)
+Special cases:
+- add `-Denable_float=true` flag in the rare case if you want to use the floating-point feature extractors.
+- add `-Denable_avx512=true` to support wider SIMD instructions to achieve the fastest processing on supported CPUs
```
nilfm, line 35: nit: this is now [enabled by default](https://github.com/Netflix/vmaf/blob/master/libvmaf/meson_options.txt#L18) so this wording may be misleading. We can fix this later though.

MorkTheOrk:
@gedoensmax For the Docker and Documentation part: Looks fine to me.
There are a few bugs I would like to fix soonish, but I dont know if it needs to be part of this PR, we might want to open a new one for more fixes.

Issue #1181, opened 2023-04-19
Segmentation Fault on arm64 but not x86_64.


brycechesternewman:
Using the latest code from vmaf compiled with ffmpeg master branch we get a segmentation fault in vmaf running on arm64.
We do not get  the same fault running vmaf/ffmpeg on x86_64. I have tried compiling with other ffmpeg branches like 5.1 or 6.0 the the segmentation fault still occurs. The backtrace after the reproduce steps shows the location of the fault.

Linux transcoder-job-7b6794585c-56knp 5.10.173-154.642.amzn2.aarch64 #1 SMP Wed Mar 15 00:26:38 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux

Steps to reproduce. 

```
uname -a
Linux transcoder-job-7b6794585c-56knp 5.10.173-154.642.amzn2.aarch64 #1 SMP Wed Mar 15 00:26:38 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux
```

```
"/usr/local/bin/ffmpeg" "-loglevel" "info" "-y" "-hide_banner" "-i" \
"https://user-images.githubusercontent.com/4998643/232534850-53ba56ee-e7e3-43bc-8eab-20d9798b99c8.mp4" \
"-sws_flags" "bicubic" "-filter_complex" "[0:v:0]scale='1080:1920'[hd16]" "-map" "[hd16]" "-map_metadata" "-1" "-f" "mov" "-vcodec" "libx264" "-pix_fmt" "yuv420p" "-profile:v" "high" "-level" "4.0" "-c:a" "aac" "-b:a" "128k" "-map" "0:a?" "-crf" "9" "-maxrate" "60M" "-bufsize" "60M" "-movflags" "faststart" "-refs" "4" "-color_primaries" "bt709" "-color_trc" "bt709" "-colorspace" "bt709" "-preset" "medium" "/tmp/hd16.mov"
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from 'https://user-images.githubusercontent.com/4998643/232534850-53ba56ee-e7e3-43bc-8eab-20d9798b99c8.mp4':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomiso2mp41
    creation_time   : 2021-03-02T18:53:26.000000Z
    encoder         : Lavf58.45.100
  Duration: 00:00:03.04, start: 0.000000, bitrate: 7184 kb/s
[...]

PR #1245, opened 2023-11-01
Fix vif_neon segmentation fault

nilfm:
The variable `i_dst_stride` is being updated inside the main loop. See lines 258 and 403. The way it's being used in the "scalar code for horizontal leftover" loop is incorrect - it should not be multiplied by `i`.

This bug affects neon code on widths not divisible by 32.

Addresses #1181


kylophone:
lgtm

Issue #1243, opened 2023-10-31
Y4M parsing is incorrect on certain inputs


nilfm:
We have observed inconsistent results on Y4M vs YUV parsing on certain types of inputs: high bit depth (>8b) and width not multiple of 16.

PR #1246, opened 2023-11-01
Add support to v2.3.1 tag to support y4m input files with different bit depths

dthoang-apple:
This pull request implements code to support AOM CWG-D088 "Coding 8-bit video using 10 bits" which was adopted for CTC v6.0. The code changes are made on top of tag v2.3.1, which is used for CTC v5.0.

I am requesting a new tag to be created for this pull request so that AOM can use it for CTC v6.0.

### DIFF
--- a/libvmaf/tools/cli_parse.c
+++ b/libvmaf/tools/cli_parse.c
@@ -351,7 +351,7 @@ void cli_parse(const int argc, char *const *const argv,
             break;
         case 'b':
             settings->bitdepth = parse_bitdepth(optarg, 'b', argv[0]);
-            settings->use_yuv = true;
+            // allow bitdepth to be set for y4m, so do not set use_yuv = true
             break;
         case 'o':
             settings->output_path = optarg;
--- a/libvmaf/tools/vidinput.c
+++ b/libvmaf/tools/vidinput.c
@@ -46,9 +46,9 @@ int raw_input_open(video_input *_vid,FILE *_fin,
   return -1;
 }
 
-int video_input_open(video_input *_vid,FILE *_fin) {
+int video_input_open(video_input *_vid,FILE *_fin,int _out_bd) {
   void *ctx;
-  if ((ctx = Y4M_INPUT_VTBL.open(_fin))!=NULL){
+  if ((ctx = Y4M_INPUT_VTBL.open(_fin,_out_bd))!=NULL){
     _vid->vtbl=&Y4M_INPUT_VTBL;
     _vid->ctx=ctx;
     _vid->fin=_fin;
--- a/libvmaf/tools/vidinput.h
+++ b/libvmaf/tools/vidinput.h
@@ -51,7 +51,7 @@ struct video_input_plane {
 };
 typedef struct video_input_plane video_input_ycbcr[3];
 
-typedef void* (*video_input_open_func)(FILE *_fin);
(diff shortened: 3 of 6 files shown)


nilfm:
Not commenting on the changes on this PR yet, but FYI: there were two bugs fixed the last few days (#1243 and #1245) and there will probably be discussion on cutting a new release from current master to include these fixes, to be used in CTC v6. In that case, we would need to coordinate to get these changes rebased on current master and into the new release.

cc @kylophone for your opinion on this PR's changes.

kylophone:
I think I would prefer a simple function in `vmaf.c` to copy and scale data from one `VmafPicture` to another `VmafPicture`. Shouldn't be a reason to change anything in the yuv/y4m parsing. What do you think?

dthoang-apple:
> I think I would prefer a simple function in `vmaf.c` to copy and scale data from one `VmafPicture` to another `VmafPicture`. Shouldn't be a reason to change anything in the yuv/y4m parsing. What do you think?

I am not so familiar with VmafPicture so confined my code changes to y4m_input.c. The change to y4m is limited to converting the bit-depth and uses the existing hook function convert(). In my opinion, this change is clean if the output bit-depth is known. The complication comes when the output bit-depth is determined as the larger of the bit-depths from the reference and distorted input files. In this case, your suggestion would be preferred.

I can port y4m_convert_scale_bitdepth() function to apply to VmafPicture is you can provide a skeleton of a function to do so and show how to insert that function call into vmaf.c

kylophone:
> I can port y4m_convert_scale_bitdepth() function to apply to VmafPicture is you can provide a skeleton of a function to do so and show how to insert that function call into vmaf.c

Alright, I had a quick look at this. Here is what I would suggest: Update [`fetch_picture()`](https://github.com/Netflix/vmaf/blob/master/libvmaf/tools/vmaf.c#L71-L121) in vmaf.c. This function allocates a `VmafPicture` and copies the already parsed y4m/yuv data into it. If you can flag this, all you need to do is make sure the call to `vmaf_picture_alloc()` is allocating a new picture of the appropriate output bitdepth and then you can copy/scale the pixel data into the new `VmafPicture`. Let me know if this makes sense to you.

dthoang-apple:
> Alright, I had a quick look at this. Here is what I would suggest: The only thing we should need to do is update [`fetch_picture()`](https://github.com/Netflix/vmaf/blob/master/libvmaf/tools/vmaf.c#L71-L121) in vmaf.c. This function allocates a `VmafPicture` and copies the already parsed y4m/yuv data into it. If you can flag this, all you need to do is make sure the call to `vmaf_picture_alloc()` is allocating a new picture of the appropriate output bitdepth and then you can copy/scale the pixel data into this new `VmafPicture`. Let me know if this makes sense to you.

This sounds good. Let me study it in more detail. What commit should I make the change to?

kylophone:
> What commit should I make the change to?

Please make your changes on top of `master`. We'll cut a new release (v3.0) for the upcoming AOM CTCv6.

dthoang-apple:
It seems VmafPicture uses little-endian byte order for frame buffer. Can you confirm?

kylophone:
Depends on your system, I think. Input and output buffers should match. In any case, you should be able to just `<< 2` to turn 8-bit into 10-bit.

nilfm:
I think we can close this one and focus on #1260.

PR #1260, opened 2023-11-15
tools/vmaf: support y4m input files with different bit depths

dthoang-apple:
This pull request implements code to support AOM CWG-D088 "Coding 8-bit video using 10 bits" which was adopted for CTC v6.0.

### DIFF
--- a/libvmaf/tools/vmaf.c
+++ b/libvmaf/tools/vmaf.c
@@ -115,7 +115,6 @@ static int fetch_picture(video_input *vid, VmafPicture *pic, int depth)
     } else if (depth > 8) {
         // unequal bit-depth
         // therefore depth must be > 8 since we do not support depth < 8
-        uint16_t value;
         int left_shift = depth - info.depth;
         if (info.depth == 8) {
             for (unsigned i = 0; i < 3; i++) {
@@ -124,15 +123,13 @@ static int fetch_picture(video_input *vid, VmafPicture *pic, int depth)
                 uint8_t *ycbcr_data = ycbcr[i].data +
                     (info.pic_y >> ydec) * ycbcr[i].stride +
                     (info.pic_x >> xdec);
-                uint8_t *pic_data = (uint8_t *)pic->data[i];
+                uint16_t *pic_data = (uint16_t*)pic->data[i];
 
                 for (unsigned j = 0; j < pic->h[i]; j++) {
                     for (unsigned k = 0; k < pic->w[i]; k++) {
-                        value=ycbcr_data[k]<<left_shift;
-                        pic_data[2*k]=(value)&0xff;
-                        pic_data[2*k+1]=(value>>8);
+                        pic_data[k] = ycbcr_data[k] << left_shift;
                     }
-                    pic_data += pic->stride[i];
+                    pic_data += pic->stride[i] / 2;
                     ycbcr_data += ycbcr[i].stride;
                 }
             }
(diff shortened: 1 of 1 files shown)


nilfm:
Thanks for the contribution! I have tried it out on a few 8b and 10b y4m files and it looks correct, i.e. the VMAF(8b, 10b) == VMAF(10b, 8b) == VMAF(10b, 10b) for 8b inputs upscaled to 10b with ffmpeg.

I would suggest the following simplification:
```
From dc9076f93bb757c06cd95d4bc5cbefa9a2d195b0 Mon Sep 17 00:00:00 2001
From: nilfm <nilf@netflix.com>
Date: Wed, 15 Nov 2023 14:54:28 +0000
Subject: [PATCH] tools/vmaf: simplify bitdepth conversion

---
 libvmaf/tools/vmaf.c | 23 +++++++++--------------
 1 file changed, 9 insertions(+), 14 deletions(-)

diff --git a/libvmaf/tools/vmaf.c b/libvmaf/tools/vmaf.c
index d93f46a0..dcf18e8b 100644
--- a/libvmaf/tools/vmaf.c
+++ b/libvmaf/tools/vmaf.c
@@ -115,7 +115,6 @@ static int fetch_picture(video_input *vid, VmafPicture *pic, int depth)
     } else if (depth > 8) {
         // unequal bit-depth
         // therefore depth must be > 8 since we do not support depth < 8
-        uint16_t value;
         int left_shift = depth - info.depth;
         if (info.depth == 8) {
             for (unsigned i = 0; i < 3; i++) {
@@ -124,15 +123,13 @@ static int fetch_picture(video_input *vid, VmafPicture *pic, int depth)
                 uint8_t *ycbcr_data = ycbcr[i].data +
                     (info.pic_y >> ydec) * ycbcr[i].stride +
                     (info.pic_x >> xdec);
-                uint8_t *pic_data = (uint8_t *)pic->data[i];
+                uint16_t *pic_data = (uint16_t*)pic->data[i];
 
                 for (unsigned j = 0; j < pic->h[i]; j++) {
                     for (unsigned k = 0; k < pic->w[i]; k++) {
-                        value=ycbcr_data[k]<<left_shift;
-                        pic_data[2*k]=(value)&0xff;
-                        pic_data[2*k+1]=(value>>8);
+                        pic_data[k] = ycbcr_data[k] << left_shift;
                     }
-                    pic_data += pic->stride[i];
+                    pic_data += pic->stride[i] / 2;
                     ycbcr_data += ycbcr[i].stride;
                 }
             }
@@ -140,19 +137,17 @@ static int fetch_picture(video_input *vid, VmafPicture *pic, int depth)
             for (unsigned i = 0; i < 3; i++) {
                 int xdec = i&&!(info.pixel_fmt&1);
                 int ydec = i&&!(info.pixel_fmt&2);
-                uint8_t *ycbcr_data = ycbcr[i].data +
-                    (info.pic_y >> ydec) * ycbcr[i].stride +
+                uint16_t *ycbcr_data = (uint16_t*) ycbcr[i].data +
+                    (info.pic_y >> ydec) * (ycbcr[i].stride / 2) +
                     (info.pic_x >> xdec);
-                uint8_t *pic_data = pic->data[i];
+                uint16_t *pic_data = (uint16_t*)pic->data[i];
 
                 for (unsigned j = 0; j < pic->h[i]; j++) {
                     for (unsigned k = 0; k < pic->w[i]; k++) {
-                        value=((ycbcr_data[2*k+1]<<8)+ycbcr_data[2*k])<<left_shift;
-                        pic_data[2*k]=(value)&0xff;
-                        pic_data[2*k+1]=(value>>8);
+                        pic_data[k] = ycbcr_data[k] << left_shift;
                     }
-                    pic_data += pic->stride[i];
-                    ycbcr_data += ycbcr[i].stride;
+                    pic_data += pic->stride[i] / 2;
+                    ycbcr_data += ycbcr[i].stride / 2;
                 }
             }
         }
-- 
2.42.0
```

By using `uint16_t*` to read and write the data, we can avoid the bit masking and complications of converting one 16b value to two 8b values. Would you agree?

joelsole:
y4m and yuv would have different support with this PR. This might lead to some confusion. You could either extend the support for input files with different bit-depths to yuv or make this difference very clear on the documentation. 
@kylophone any preference?

nilfm:
IMO it's acceptable to allow different bit depths for Y4M but not YUV, as long as we document it properly. Making this change for Y4M is transparent to users who don't use it, but changing this for YUV would imply changing the CLI to add another `bitdepth` parameter, and may set a precedent for later adding another `pix_fmt` parameter, etc.

dthoang-apple:
> Thanks for the contribution! I have tried it out on a few 8b and 10b y4m files and it looks correct, i.e. VMAF(8b, 10b) == VMAF(10b, 8b) == VMAF(10b, 10b) for 8b inputs upscaled to 10b with ffmpeg. (Actually, VMAF(8b, 8b) was also identical in this case, not sure if that is always true, but CAMBI scores were different which is expected).
> 
> I would suggest the following simplification:
> 
> ```

I tried something like that the first time and got wrong results, which I suspect to be related to wrong endianness. Did you test your code?

dthoang-apple:
> Yes, I would suggest a few things:
> 
> * Implicit bitdepth conversion cannot become the new default, this needs to be **enabled** by a command-line flag.
> * If you don't set this command-line flag and input bitdepths do not match, fail (maintain current behavior).
> * It is ok if this only works for y4m only, fail if this command-line flag is used with yuv input.

I suggest using "--aom_ctc v6.0" to enable this behavior in addition to any other command-line flag. If "--aom_ctc v6.0" is used then the other command-line flag is not needed.

kylophone:
> > Yes, I would suggest a few things:
> > 
> > * Implicit bitdepth conversion cannot become the new default, this needs to be **enabled** by a command-line flag.
> > * If you don't set this command-line flag and input bitdepths do not match, fail (maintain current behavior).
> > * It is ok if this only works for y4m only, fail if this command-line flag is used with yuv input.
> 
> I suggest using "--aom_ctc v6.0" to enable this behavior in addition to any other command-line flag. If "--aom_ctc v6.0" is used then the other command-line flag is not needed.

Sure, that works.

nilfm:
Closing this, follow up on #1264

PR #1262, opened 2023-11-17
refactor: format conversion to reduce code repetition

arjun-234:
This commit introduces a new function convert_format that handles the conversion of y, u, and v based on the bit depth. This change helps to reduce code repetition in the reader.py file. The bit depth is determined based on the conditions and the convert_format function is used to convert y, u, and v. This makes the code more maintainable and easier to understand.

### DIFF
--- a/python/vmaf/tools/reader.py
+++ b/python/vmaf/tools/reader.py
@@ -204,8 +204,8 @@ class YuvReader(object):
                 assert False
 
             y = self.convert_format(y, bit_depth)
-            u = self.convert_format(u, bit_depth)
-            v = self.convert_format(v, bit_depth)
+            u = self.convert_format(u, bit_depth) if u is not None else None
+            v = self.convert_format(v, bit_depth) if v is not None else None
             return y, u, v
 
         else:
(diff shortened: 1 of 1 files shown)


nilfm:
Thanks for the refactor! The only thing that seems a bit concerning is the treatment of `u` and `v` when they are `None`. I'll further comment inline.

python/vmaf/tools/reader.py
```diff
@@ -190,27 +193,20 @@ def next(self, format='uint'):
...
-                u = u.astype(np.double) / (2.0**16 - 1.0) if u is not None else None
-                v = v.astype(np.double) / (2.0**16 - 1.0) if v is not None else None
-                return y, u, v
+                bit_depth = 16
             else:
                 assert False
 
+            y = self.convert_format(y, bit_depth)
+            u = self.convert_format(u, bit_depth)
+            v = self.convert_format(v, bit_depth)
```
nilfm, line 208: I think just changing this to
```
u = self.convert_format(u, bit_depth) if u is not None else None
v = self.convert_format(v, bit_depth) if v is not None else None
```
would maintain the existing behavior. No concerns other than this.

arjun-234: Hello @nilfm, Thank you for your review! I have revised the code according to your instructions.

nilfm: Looks good to me! Tagging @li-zhi since he wrote most of this code, but I'll be merging soon if no objections.

PR #1264, opened 2023-11-20
tools/vmaf: support y4m input files with different bit depths; add command-line flag to gate this behavior

nilfm:
Continuation of #1260, with two extra commits to gate the behavior behind the new `--convert_bitdepth` CLI flag, and to add `--aom_ctc v6.0` enabling this flag.

### DIFF
--- a/libvmaf/tools/cli_parse.c
+++ b/libvmaf/tools/cli_parse.c
@@ -320,6 +320,12 @@ static void aom_ctc_v5_0(CLISettings *settings, const char *app)
     aom_ctc_v4_0(settings, app);
 }
 
+static void aom_ctc_v6_0(CLISettings *settings, const char *app)
+{
+    aom_ctc_v5_0(settings, app);
+    settings->common_bitdepth = true;
+}
+
 static void parse_aom_ctc(CLISettings *settings, const char *const optarg,
                           const char *const app)
 {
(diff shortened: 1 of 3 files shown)


nilfm:
@dthoang-apple FYI, this is the continuation of your PR. I took it to my own branch for convenience but you retained authorship of your commits.

@kylophone Let me know if this looks good to you. I have tested y4m inputs with (8, 8), (8, 10), (10, 8), (10, 10) bitdepths and they work as expected, and yuv inputs always throw an error when this flag is set.

joelsole:
lgtm

nilfm:
Just thought of one problem, we can't disallow `convert_bitdepth` on YUV inputs and at the same time enable the flag on `--aom_ctc v6.0`, or the AOM setting won't be available to use with YUV inputs. Two options:
1. Inside `aom_ctc_v6_0`, only set `convert_bitdepth` to true if `use_yuv` is true.
2. Do not enforce `convert_bitdepth` to be false on YUV inputs.

I'll make the change for (1) since it seems simple enough... please let me know if you would prefer (2)

nilfm:
On second thought, we will not add the `common_bitdepth` parameter to the CLI options but rather set it only via the AOM CTC flag.

nilfm:
I will squash my 5 commits into one and we can merge. @kylophone FYI

PR #1267, opened 2023-11-20
Fix remaining warnings

nilfm:
Need to run some tests to make sure this doesn't change anything.


nilfm:
This PR is currently broken, please do not merge. Checking.

nilfm:
Appears to be fixed but I wouldn't risk including this in the release, let's hold off on it.

nilfm:
After this PR, I don't get any warnings compiling with `-Denable_float=true -Denable_avx512=true`.

nilfm:
I think this is ready. I'll merge on Monday if no objections.

PR #1268, opened 2023-11-22
add test_cli_parse.c

nilfm:
To start with, I only tested some basic behavior of the `--aom_ctc` and `--nflx_ctc` flags but this can easily be extended to cover other code paths.


kylophone:
lgtm

PR #1257, opened 2023-11-08
Test Github Actions MacOS M1 Runner

kylophone:
https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

If we can get arm64 builds on Github Actions CI, we should be able to get rid of Travis completely.

PR #1269, opened 2023-11-23
libvmaf.yml: test C, AVX-2 and AVX-512 code paths

nilfm:


nilfm:
@kylophone This is what I had in mind when I mentioned testing all the code paths, by setting the flags while building the tests should use the right code path for each case. In this case we had 4 cases already so I used them to test the different code paths, but if we want to, we can have 4 * 3 cases and test all possible combinations. This can also be extended with the arm64 machines when we have them available.

nilfm:
I tried to break some code paths to see if CI detects it. It seems that this CI can run C and AVX-2 but not AVX-512.

* #1272 - Breaking C code - all tests are failing, but the AVX versions are only failing on tests specifically meant to disable AVX. Working as expected.
* #1270 - Breaking AVX-2 code - failing tests for AVX-512 and AVX-2, passing tests for C. This probably means that AVX-512 was not available and it fell back to AVX-2 on that case.
* #1271 - Breaking AVX-512 code - passing tests, so this code is never executed, even though it is compiled (see ninja output in the CI step).

We should look into getting instances that can run AVX-512 code. It would be good to address this and #1257 together.

nilfm:
A few advantages of this vs `cpumask`: 
1. It is run in parallel so no extra CI time to test these additional code paths.
2. It is clear from the CI tab which code path is broken (not entirely - there are a couple of tests that disable optimizations).
3. It runs all tests on all instruction sets - better coverage than just adding a few tests with various `cpumask` settings, also less duplication in the test code.

nilfm99:
Closing this for now - without AVX-512 hardware, it doesn't provide much gain.

PR #1277, opened 2023-12-07
libvmaf: v3.0.0

kylophone:


libvmaf/meson.build
```diff
@@ -8,7 +8,7 @@ project('libvmaf', ['c', 'cpp'],
                       ],
     meson_version: '>= 0.56.1')
 
-vmaf_soname_version       = '1.1.3'
```
nilfm, line 11: curious why this was not 2.3.1, where does this make a difference?

kylophone: libvmaf v2.0 actually had no backwards-incompatible changes, so bumping the major soversion technically was not required. Now that we've deprecated `compute_vmaf()` and need to bump major version I took the opportunity to jump straight to v3.0.0.

PR #1279, opened 2023-12-11
Update README.md badges

nilfm:
The badge urls in README.md were not correct and were not refreshing when a build passed or failed.


nilfm:
This ffmpeg build failure is unrelated to libvmaf and clearly unrelated to this change, and we'll have to deal with it separately.

PR #1280, opened 2023-12-11
Update README.md - v3.0.0

nilfm:


ltrudeau-twoorioles:
Do you also plan on updating `CHANGELOG.MD`?

nilfm:
Thanks for the reminder. @kylophone FYI, do you have something like a list of important changes since v2.3.1? Otherwise I can go through them.

kylophone:
> @kylophone FYI, do you have something like a list of important changes since v2.3.1? Otherwise I can go through them.

You can check API changes with `git diff v2.3.1 libvmaf/include`. Two things here:

- Removal of `compute_vmaf()` necessitating lib version and soversion bump.
- Addition of new `vmaf_cuda_*` APIs.

I think we had some speed related changes, might be a good idea to benchmark those against v2.3.1.

PR #1281, opened 2023-12-11
ffmpeg.yml: use macos-13

nilfm:


nilfm:
FYI @kylophone

kylophone:
Interesting. Once `macos-latest` moves, we'll want to switch back, I guess.

kylophone:
@nilfm Oh, even better. Let's just disable this in the FFmpeg configure so we can just stick to the latest. I think `--disable-indevs` should work for all platforms.

nilfm:
Opened #1282 to try this out

nilfm:
I think this is related too: https://ffmpeg.org/pipermail/ffmpeg-devel/2023-December/318200.html

PR #1290, opened 2023-12-18
Update meson command

ltrudeau-twoorioles:
Removes the following warning

```
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```


nilfm (approved):
Thanks!

Issue #1228, opened 2023-08-19
-Wstringop-truncation build warning


AdelKS:
Hello,

When compiling on Linux with GCC 13.2, I get these warnings:

```
 * ../vmaf-2.3.1/libvmaf/src/model.c:129:9: warning: ‘__builtin_strncpy’ specified bound depends on the length of the source argument [-Wstringop-truncation]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1108:33: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1110:33: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1112:33: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1114:17: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1121:33: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1123:33: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1125:33: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
 * ../vmaf-2.3.1/libvmaf/src/feature/x86/vif_avx2.c:1127:33: warning: passing argument 1 of ‘_mm_loadu_si128’ from incompatible pointer type [-Wincompatible-pointer-types]
[...]

PR #1304, opened 2024-01-10
model.c: change strncpy to strcpy, fix warning -Wstringop-truncation

nilfm99:
I don't think this changes anything, since the size parameter `name_sz` was already computed from the lengths of the respective sources using `strlen`. This fixes the following warning:

```
warning: ‘__builtin_strncpy’ specified bound depends on the length of the source argument [-Wstringop-truncation]
```

Addresses #1228.

### DIFF
--- a/libvmaf/src/model.c
+++ b/libvmaf/src/model.c
@@ -131,9 +131,9 @@ char *vmaf_model_generate_name(VmafModelConfig *cfg)
     memset(name, 0, name_sz);
 
     if (!cfg->name)
-        strncpy(name, default_name, name_sz);
+        strcpy(name, default_name);
     else
-        strncpy(name, cfg->name, name_sz);
+        strcpy(name, cfg->name);
 
     return name;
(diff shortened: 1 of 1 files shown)


kylophone:
Probably OK. I always prefer strncpy, is this a legit warning or just an overzealous compiler?

nilfm99:
I think its concern is that `strncpy` does not necessarily add the null terminator if the destination is full. In this case there's always enough space for it due to how `name_sz` is computed.

```
Warning: If there is no null byte among the first n bytes of src, the string placed in dest will not be null-terminated.
```

Some discussion here: https://stackoverflow.com/questions/50198319/gcc-8-wstringop-truncation-what-is-the-good-practice

I also tried setting the null terminator explicitly after the `strncpy` calls but it doesn't make the warning go away. Some people in that thread suggest making a custom `strncpy` function that uses `memcpy` and null-terminates it...

PR #1310, opened 2024-01-12
model.c: fix -Wlto-type-mismatch

nilfm99:
Partly addresses #1228. The mismatch is between these `extern` variables and the output of `xxd` which hex-dumps the model JSON files into unsigned char arrays in C files.


nilfm99:
@kylophone is this okay to merge? I don't think it changes anything but it fixes some warnings due to a type mismatch between the output of `xxd` and the type we declare here

kylophone:
Yes, I think this is good to go. Thanks!

PR #1312, opened 2024-01-17
Added implementation for framesync api and integration into vmaf

niranjankumar-ittiam:
All features (extract API) have been modified to pass framesync context.

### DIFF
--- a/libvmaf/include/libvmaf/libvmaf.h
+++ b/libvmaf/include/libvmaf/libvmaf.h
@@ -30,8 +30,6 @@
 extern "C" {
 #endif
 
-#define MAX_VMAF_THREADS  8 // default maximum threads
-
 enum VmafLogLevel {
     VMAF_LOG_LEVEL_NONE = 0,
     VMAF_LOG_LEVEL_ERROR,
--- a/libvmaf/include/libvmaf/libvmaf.h
+++ b/libvmaf/include/libvmaf/libvmaf.h
@@ -30,6 +30,8 @@
 extern "C" {
 #endif
 
+#define MAX_VMAF_THREADS  8 // default maximum threads
+
 enum VmafLogLevel {
     VMAF_LOG_LEVEL_NONE = 0,
     VMAF_LOG_LEVEL_ERROR,
--- a/libvmaf/src/feature/cambi.c
+++ b/libvmaf/src/feature/cambi.c
@@ -23,6 +23,7 @@
 #include "common/macros.h"
 #include "cpu.h"
 #include "feature_collector.h"
+#include "framesync.h"
 #include "feature_extractor.h"
 #include "log.h"
 #include "luminance_tools.h"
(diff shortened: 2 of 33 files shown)


libvmaf/include/libvmaf/libvmaf.h
```diff
@@ -30,6 +30,8 @@
 extern "C" {
 #endif
 
+#define MAX_VMAF_THREADS  8 // default maximum threads
```
kylophone, line 33: I don't think we want this is the public header.

libvmaf/src/feature/cambi.c
```diff
@@ -1080,7 +1081,8 @@ static double combine_dist_src_scores(double dist_score, double src_score) {
 static int extract(VmafFeatureExtractor *fex,
                    VmafPicture *ref_pic, VmafPicture *ref_pic_90,
                    VmafPicture *dist_pic, VmafPicture *dist_pic_90,
-                   unsigned index, VmafFeatureCollector *feature_collector) {
+                   unsigned index, VmafFeatureCollector *feature_collector,
+				   VmafFrameSyncContext *framesync) {
```
kylophone, line 1085: Please add the `VmafFrameSyncContext` inside the `VmafFeatureExtractor` struct. That way we do not need to break the `extract` API and touch all these other files.

kylophone:
Closing #1287, we'll track this here instead.

libvmaf/test/test_framesync.c
```diff
@@ -1,90 +1,139 @@
...
-    err = vmaf_framesync_submit_data(fs_ctx, my_data, 2);
-    mu_assert("problem during vmaf_framesync_submit_data", !err);
-
-    mu_assert("callback should have been executed for index == 1, "
-              "callback dependencies are met: (0, 1, 2)",
-              most_recent_callback_index == 1);
-
-    err = vmaf_framesync_close(fs_ctx);
-    mu_assert("problem during vmaf_framesync_close", !err);
+		err = vmaf_thread_pool_enqueue(pool, framesync_proc_func, &data, sizeof(ThreadData));
```
nilfm99, line 113: Note: There is inconsistent use of spaces and tabs in a few places

libvmaf/include/libvmaf/libvmaf.h
niranjankumar-ittiam: MAX_VMAF_THREADS  has been deleted and framesync has been modified to allocate memory dynamically.

libvmaf/src/feature/cambi.c
niranjankumar-ittiam: `.extract()` API has been clean up and `framesync` context has been passed inside the `VmafFeatureExtractor` struct.

libvmaf/test/test_framesync.c
niranjankumar-ittiam: Tabs have been converted to spaces.

nilfm99:
Thank you for the PR. A few thoughts:
* I tried to run the tests on my M1 mac but they don't compile due to a `pthread` include issue. The CI runs using mac OS also had this problem.
* The sleep function is not portable across windows and linux, windows uses `Sleep` which takes milliseconds as input. I'm also not 100% sure about adding sleep to the tests being good practice, in this case it's just a few seconds though.
* I took a stab at fixing those issues and came up with [this commit](https://github.com/Netflix/vmaf/commit/1c393dd323da401e0996d9aead16c31f9d0cba99), which also fixes a few typos and style inconsistencies. Feel free to cherry-pick it into your PR if you want.

MallikarjunKamble:
All the suggested changes (pthread, sleep, typos and style inconsistencies) are pushed in the new commit @nilfm99

Issue #1313, opened 2024-01-17
Errors installing Python dependencies on ARM based macOS


cosmin:
I'm trying to setup a development environment on an ARM based macOS (M1) and it's been much more exciting than I expected. It might be worth updating either the dependencies or the build instructions or both.

I started by following the usual Python requirements from `resource/doc/python.md` until I got to

```
pip install -r ./python/requirements.txt
```

fails with the default `clang` on macOS with an error building `libsvm-official`

```
clang: error: unsupported option '-fopenmp'
```

which makes sense as macOS llvm does not support OpenMP. Common alternatives are to use GCC instead (`brew install gcc`). However

```
 CC=gcc-13 CXX=g++-13 pip install -r ./python/requirements.txt
```

fails due to `gcc-13: error: unrecognized command-line option '-iwithsysroot/System/Library/Frameworks/System.framework/PrivateHeaders'`.

Another alternative is using llvm from homebrew (`brew install llvm`) which would support OpenMP, however

```
LLVM_CONFIG=/opt/homebrew/opt/llvm/bin/llvm-config pip install -r ./python/requirements.txt
```

fails because it's trying to compile a universal binary (`-arch arm64 -arch x86_64`) and the Homebrew llvm does not support universal builds, so it complains about trying to build x86_64 against libraries only compiled for arm64.

Tracked this down to using system python which is a universal binary, so I tried again with homebrew python3 which would be arm64 only. 

```
brew install python
[...]
LLVM_CONFIG=/opt/homebrew/opt/llvm/bin/llvm-config pip install -r ./python/requirements.txt
```

this then fails to build PyWavelets due to 

```
pywt/_extensions/_pywt.c:253:12: fatal error: 'longintrepr.h' file not found
```

It looks like this this may have changed around Python 3.11, presumably a new version of PyWavelets might fix this but it is pinned to 1.1.1, so I tried downgrading Python to 3.10 to see if that would fix it

```
brew install python@3.10
[...]
LLVM_CONFIG=/opt/homebrew/opt/llvm/bin/llvm-config pip install -r ./python/requirements.txt
```

but then ran into another error in PyWavelets

```
 pywt/_extensions/_pywt.c:32259:5: error: expression is not assignable
          ++Py_REFCNT(o);
```

[...]

PR #1315, opened 2024-01-18
materialize map object before calling np.vstack

cosmin:
Newer numpy versions no longer allow passing in iterable objects to array stack. 
```
arrays to stack must be passed as a "sequence" type such as list or tuple
```

This has been a future warning for a while and I guess at some point it was actually removed. The easy fix is to wrap the map object in a list() call.


cosmin:
Fixes one of the test failures mentioned in #1313

nilfm99 (approved):

PR #1317, opened 2024-01-19
Fix Python unit tests on arm64

cosmin:
As mentioned in #1313 running unit tests on arm64 (on an ARM based Mac) causes about 20 failures due to floating point differences. This PR resolved all cases where the difference was ~1 decimal place. There are still two outstanding failures where the relative difference is much larger

```
======================================================================
FAIL: test_adm_dwt2_cy_dc (cy_test.AdmDwt2CyTest.test_adm_dwt2_cy_dc)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/cstejerean/workspace/vmaf/python/test/cy_test.py", line 110, in test_adm_dwt2_cy_dc
    self.assertAlmostEqual(float(np.max(v)), 8.526512829121202e-14, places=16)
AssertionError: 7.574895579319176e-14 != 8.526512829121202e-14 within 16 places (9.51617249802026e-15 difference)

======================================================================
FAIL: test_adm_dwt2_cy_xsmallP_dc (cy_test.AdmDwt2CyTest.test_adm_dwt2_cy_xsmallP_dc)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/cstejerean/workspace/vmaf/python/test/cy_test.py", line 91, in test_adm_dwt2_cy_xsmallP_dc
    self.assertAlmostEqual(float(np.max(v)), 8.526512829121202e-14, places=16)
AssertionError: 7.574895579319176e-14 != 8.526512829121202e-14 within 16 places (9.51617249802026e-15 difference)
```

For that one I'm not sure if the right solution is to update the test to allow for a 9e-15 difference or whether there are some legitimate numerical stability problems that need to be addressed in the underlying code. Thoughts?


nilfm99:
Hi @cosmin, thanks for the PR! I am trying to reproduce these failing tests on my M1 Macbook Pro, and so far I can't reproduce: with any of python 3.9, 3.10, and 3.11, and Cython==3.0.8, I get the results currently on master, and changing them to the ones you mention causes the tests to fail. Is there anything about your environment that may be different?

Looping in @li-zhi who originally worked on these tests for any ideas.

cosmin:
The changes here strictly reduce the precision so it shouldn't cause any tests to fail. What kind of failures are you seeing? 

I'll try to repro this on a second machine and report back.

nilfm99:
@cosmin Sorry I wasn't clear - the changes in the PR don't cause it to fail, but what I meant is that the major changes you are seeing (`7.574895579319176e-14 != 8.526512829121202e-14 within 16 places`) would make it fail if we changed the expected score to the one you are getting. This PR seems good to me, I'm trying to understand what causes the other two changes.

cosmin:
Ah ok, that one I was concerned about as well given the relative magnitude. Let me see if I can track that down.

nilfm99:
Worth noting that even for the tests changed in this PR, on my machine they pass on the master branch, which probably points to a difference in our setups as well.

nilfm99:
I checked what the code is supposed to be doing, and the expected results for these tests should really be {110, 0, 0, 0}, anything else is the product of floating point inaccuracies due to the filter coefficients not adding up to sqrt(2) and 0 respectively.

I think we can set those expected values and 12 decimal places, and that should pass in all environments. I don't think there is value in having a test to check that the result is inaccurate in this specific way.

I talked to @li-zhi and he shares this view as well.

nilfm99:
@cosmin please feel free to update this PR for those tests or let me know and I will start a new PR for it

cosmin:
@nilfm99 I'll update the pull request shortly

nilfm99:
thank you very much for the contribution! LGTM

nilfm99 (approved):

PR #1322, opened 2024-01-30
github/workflows/libvmaf.yml: upload to a specific name for each matrix.oss and matrix.cc

### LINKED ISSUE #1278: CI fails on every release
https://github.com/Netflix/vmaf/actions/runs/7133167143/job/19425266007

The upload_url seems wrongly templated:
```
upload_url: [https://uploads.github.com/repos/Netflix/vmaf/releases/133078881/assets{?name,label}](https://uploads.github.com/repos/Netflix/vmaf/releases/133078881/assets%7B?name,label})
```

cc @kylophone if you know what is causing this, otherwise I'll investigate with low priority. Binaries are still being uploaded correctly, see [here](https://github.com/Netflix/vmaf/releases/tag/v3.0.0). Not sure if there's any impact at all.

nilfm99:
Should fix #1278 and unblock #1292

### DIFF
--- a/.github/workflows/libvmaf.yml
+++ b/.github/workflows/libvmaf.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Upload vmaf
         uses: actions/upload-artifact@v3
         with:
-          name: ${{ matrix.os }}-vmaf
+          name: ${{ matrix.os }}-${{ matrix.CC }}-vmaf
           path: ${{ steps.get_info.outputs.path }}
       - name: Upload vmaf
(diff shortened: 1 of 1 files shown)


nilfm99:
on second thought, to fix #1278 we also need to fix the `actions/upload-release-asset@v1` step

nilfm99:
I think this still makes sense to merge to unblock #1292, and then we can think how to solve #1278. The [action](https://github.com/actions/upload-release-asset) we use to upload release assets has been archived and the README suggests using a newer one, we can look into that.

Issue #1214, opened 2023-08-02
Common encoder bug/feature leading to inaccuracy for even n_subsample values


veikk0:
Thought I'd bring this up here, even though it's not an issue with VMAF itself, since it's likely to affect VMAF users who make use of the subsampling feature.

An issue [was reported](https://github.com/xiph/rav1e/issues/2468) in rav1e where the encoder retains a significantly differing amount of grain/detail every other frame. [shssoichiro](https://github.com/shssoichiro) did some digging and [came to the conclusion](https://github.com/xiph/rav1e/issues/2468#issuecomment-1086533406) that this is a natural effect of using frame pyramids and that all encoders are prone to it. So far I've also reproduced this with x264, x265, libaom, and SVT-AV1 using the ducks_take_off_1080p50_60f.y4m test clip from the objective-2-slow dataset. I was unable to reproduce this using h264_vaapi and hevc_vaapi on a Radeon RX 6600, so hardware encoders may or may not be affected.

How this relates to VMAF is that libvmaf offers the `n_subsample` option to speed up calculation by only computing scores every N frames. When an even value of `n_subsample` is used with an input file exhibiting the aforementioned issue, only the frames with a higher amount of detail are evaluated, leading to a variance from the ground truth VMAF score. This variance can be pretty significant and varies by encoder and often with the encoder preset. With rav1e's speed 9, `n_subsample=2` lead to a VMAF score of 96.716700, when the ground truth was 93.087603. That's an error of +3.62. With SVT-AV1's preset 12 + tune 0, the error was +2.71.

Now, this might not be as much of an issue when calculating the average VMAF of longer videos. Due to a variance in content, the error might be averaged out. But in some of my tests with a 60-second video, there was still an error of +0.39 with `n_subsample=2`, whereas using a value of 3 or 5 had an error of about +0.02 at most.

Where I can see this being an issue is VMAF-guided shot-based encoding and other applications where the content being evaluated is shorter clips. In such cases, this may lead to quality dips from one scene to the next.

I haven't done thorough testing, since the issue being present can vary a lot based on content, the encoder and settings used, and possibly the kind of VMAF quality being targeted (since more/less detail will be present). But perhaps this is a common enough of an encoder quirk that recommending against the use of even subsampling values would be a good idea?

Attached are some sample encodes of ducks_take_off_1080p50_60f done with various encoders for easier reproducibility, followed by the error introduced by using `n_subsample=2`.

[...]

PR #1324, opened 2024-02-01
Add link to n_subsample observation on usage page

nilfm99:
Addresses the comment at #1214

### DIFF
--- a/libvmaf/include/libvmaf/libvmaf.h
+++ b/libvmaf/include/libvmaf/libvmaf.h
@@ -55,6 +55,31 @@ enum VmafPoolingMethod {
     VMAF_POOL_METHOD_NB
 };
 
+/**
+ * @struct VmafConfiguration
+ * @brief  Configuration needed to initialize a `VmafContext`
+ *
+ * @param log_level   How verbose the logger is.
+ *
+ * @param n_threads   How many threads can be used to run
+ *                    feature extractors concurrently.
+ * 
+ * @param n_subsample Compute scores only every N frames.
+ *                    Note that setting an even value for N can lead to
+ *                    inaccurate results. For more detail, see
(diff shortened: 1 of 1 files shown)


kylophone:
I think maybe the cli tool help is not the right place for this, since there is actually come nuance involved. Would it be better if we put this into some of our usage docs or README?

nilfm99:
got it, I'll update this PR tomorrow

nilfm99:
I thought that documenting the entire `VmafConfiguration` struct was useful and a good place for this information, given that this is a very specific detail, and that our `README.md` already says:

> For complete API documentation, see [libvmaf.h](https://github.com/Netflix/vmaf/blob/master/libvmaf/include/libvmaf/libvmaf.h).

libvmaf/include/libvmaf/libvmaf.h
```diff
@@ -55,6 +55,31 @@ enum VmafPoolingMethod {
...
+ *
+ * @param n_threads   How many threads can be used to run
+ *                    feature extractors concurrently.
+ * 
+ * @param n_subsample Compute scores only every N frames.
+ *                    Note that setting an even value for N can lead to
+ *                    inaccurate results. For more detail, see
+ *                    https://github.com/Netflix/vmaf/issues/1214
+ * 
+ * @param cpumask     Restrict permitted CPU instruction sets.
```
kylophone, line 72: `cpumask` is actually platform independent, we also have a mask for arm64 neon.

nilfm99: sorry, do you mean platform dependent? I can add the NEON comment on the "if cpumask & 1" line

kylophone:
lgtm

PR #1325, opened 2024-02-01
Framesync API

kylophone:
Re-opening of PR #1287.

### DIFF
--- a/libvmaf/src/framesync.c
+++ b/libvmaf/src/framesync.c
@@ -1,0 +1,227 @@
+/**
+ *
+ *  Copyright 2016-2023 Netflix, Inc.
+ *
+ *     Licensed under the BSD+Patent License (the "License");
+ *     you may not use this file except in compliance with the License.
+ *     You may obtain a copy of the License at
+ *
+ *         https://opensource.org/licenses/BSDplusPatent
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ *
+ */
+
+#include <errno.h>
+#include <string.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include "framesync.h"
+
+enum {
+    BUF_FREE = 0,
+    BUF_ACQUIRED,
+    BUF_FILLED,
+    BUF_RETRIEVED,
+};
+
+typedef struct VmafFrameSyncBuf {
+    void *frame_data;
+    int buf_status;
+    signed long index;
+    struct VmafFrameSyncBuf *next;
+} VmafFrameSyncBuf;
+
+typedef struct VmafFrameSyncContext {
+    VmafFrameSyncBuf *buf_que;
+    pthread_mutex_t acquire_lock;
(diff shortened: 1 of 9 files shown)


kylophone:
Squashed and rebased, ready for merge.

libvmaf/src/framesync.c
```diff
@@ -0,0 +1,227 @@
...
+    // create a new node if all nodes are occupied in the list and append to the tail
+    if (*data == NULL) {
+        VmafFrameSyncBuf *new_buf_node = malloc(sizeof(VmafFrameSyncBuf));
+        buf_que->next = new_buf_node;
+        new_buf_node->buf_status = BUF_FREE;
+        new_buf_node->index = -1;
+        new_buf_node->next = NULL;
+        fs_ctx->buf_cnt++;
+
+        new_buf_node->frame_data = *data = malloc(data_sz);
```
nilfm99, line 101: nit: There is some duplicated logic here, maybe it is possible to extract out the "create new node" and "allocate node's data" into two functions to be reused. But we can refactor this later.

libvmaf/src/framesync.c
```diff
@@ -0,0 +1,227 @@
...
+{
+    VmafFrameSyncBuf *buf_que = fs_ctx->buf_que;
+
+    pthread_mutex_lock(&(fs_ctx->retrieve_lock));
+
+    // loop until a matchng buffer is found
+    for (unsigned i = 0; i < fs_ctx->buf_cnt; i++) {
+        if ((buf_que->index == index) && (buf_que->buf_status == BUF_ACQUIRED)) {
+            buf_que->buf_status = BUF_FILLED;
+            if (data != buf_que->frame_data)
```
nilfm99, line 124: what is the purpose of this check? 

It seems that to succeed, the pair `(index, data)` needs to match the pair `(buf_que->index, buf_que->frame_data)`. Curious about:
1. Why that condition is needed
2. Why the treatment is different if the index doesn't match (do nothing) vs if the index matches but the data pointer doesn't (mark as filled and return -1)
3. What if the entire loop finishes without finding a match?

kylophone: Forwarded to Ittiam.

IttiamVijayakumarGR: Q1:
Index matching is done to find the node which holds the data buffer of picture "index"

Q2:
Index matching is the primary check to change the buffer status to FILLED, 
Data pointer checking is for error checking purpose, which can be cleaned up if not necessary. if the caller has not passed the matching data buffer, then there can be OOB or unallocated buffer access, this additional check is added to catch such cases.
Q3:
This condition is not applicable since the thread which has acquired new buffer for any particular picture index is responsible for submitting it after filling, so index should match with a node in the queue.

libvmaf/src/framesync.c
IttiamVijayakumarGR: Yes, agreed with 2 functions the code can made reusable.

libvmaf/src/framesync.c
nilfm99: Thanks @IttiamVijayakumarGR for clarifying, I was discussing this with @kylophone and it seemed like the extra check can be cleaned up, so I did that in the PR below. I will leave the last call to Kyle though.

kylophone: Discussed this with @nilfm99, we'll merge this one as-is for now to unblock the funque upstreaming work. However, I think it's likely that we'll have some small changes to the api in a later iteration. Since this is an internal api with just one use so far, we do have some flexibility when it comes to refactoring.

PR #1327, opened 2024-02-06
Simplify framesync API

nilfm99:
The `data` parameter to `vmaf_framesync_submit_filled_data` and `vmaf_framesync_release_buf` was only used to check that it was the same pointer that was present in the framesync internal data structure. I believe this can be simplified.

As an example usage, the `integer_funque` implementation [here](https://github.com/MallikarjunKamble/vmaf/blob/FLOAT_FUNQUE_PLUS_MULTITHREADING/libvmaf/src/feature/third_party/funque/integer_funque.c) creates two dummy variables, `shared_buf_temp` and `dependent_buf_temp`, just to pass into these functions. With this simplification, we could avoid those.

### DIFF
--- a/libvmaf/src/framesync.c
+++ b/libvmaf/src/framesync.c
@@ -110,8 +110,7 @@ int vmaf_framesync_acquire_new_buf(VmafFrameSyncContext *fs_ctx, void **data,
     return 0;
 }
 
-int vmaf_framesync_submit_filled_data(VmafFrameSyncContext *fs_ctx, void *data,
-                                      unsigned index)
+int vmaf_framesync_submit_filled_data(VmafFrameSyncContext *fs_ctx, unsigned index)
 {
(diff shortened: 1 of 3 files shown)


nilfm99:
This PR is stacked on top of #1325

kylophone:
I feel like it might be more intuitive to pass the data pointer back and forth rather than just index. Thoughts on that idea?

nilfm99:
Sorry, could you clarify what you mean by

> pass the data pointer back and forth rather than just index

is that the current status at #1325 or is there any difference in what you suggest?

nilfm99:
@kylophone hi, can we go ahead with this one?

kylophone:
> @kylophone hi, can we go ahead with this one?

I think we still need to wait on the pr using this. I was expecting something sooner, so I'll ping them and see what's up.

PR #1339, opened 2024-03-12
Install python build dependencies in Makefile; simplify Dockerfile

nilfm99:
This seems like a more streamlined user experience to make sure they have the pip dependencies installed when they run `make` or `make install`.


kylophone (approved):

kylophone:
lgtm, please squash

Issue #1351, opened 2024-03-21
Segfault on some invalid inputs


nilfm99:
If the user passes in a YUV or Y4M that specifies `bitdepth=10` but actually has pixel values greater than 1023 (which is possible to represent in the YUV format, just invalid), we don't have any validation against that. As a result, we can either get nonsensical results from most features, or even get a segfault from CAMBI since it uses the pixel values to index an array of histograms.

Ideally we should validate inputs, but it is computationally expensive to check every pixel. In a quick benchmark, I saw an ~8% performance degradation by checking all the pixels inside `validate_pic_params` in `libvmaf.c`. Alternatively, we could sample a few pixels as a heuristic.

nilfm99:
Note that it is impossible to validate this with 100% confidence, since a 12-bit YUV could have low values throughout and be a valid 10-bit YUV, with no way of telling them apart.

PR #1352, opened 2024-03-21
libvmaf: add validation for pixel values above expected max value

nilfm99:
Addresses #1351. Validates only one row from ref and dist, with no observable performance overhead, since validating the entire buffer is very expensive.


kylophone:
I'm not sure this sort of incomplete check is worth anything since it both overhead and does not guarantee we avoid the problem. It might be better to just make the check inside cambi before updating the histogram.

PR #1355, opened 2024-03-26
cambi.c: validate input value ranges

nilfm99:
Replaces #1352. Benchmarks on my laptop show no observable overhead wrt master but need to check further.

(my benchmarks were on a 10-bit input so it's not just because of the early return in 8-bit and 16-bit cases)


kylophone:
LGTM for now. Might need to revisit if we need to go after CAMBI speedups.

kylophone (approved):

PR #1359, opened 2024-04-22
Selectively allow more leading zeros when outputting feature values

cosmin:
During the implementation of FUNQUE+ feature extractors we ran into problems with ST-RRED due to the amount of leading zeros required. Even after applying some normalization factors it would still routinely exceed the 0.6f formatting in VMAF. However changing the default feature format to allow more zeros seemed invasive and broke a bunch of tests.

As a result we selectively enabled printing more digits only when required by the actual feature value, while maintaining compatibility with existing feature extractors which don't need this. 

This can wait until the FUNQUE+ feature extractor that requires it.

Co-authored-by: Priyanka-885 <priyanka.mane@ittiam.com>

### DIFF
--- a/libvmaf/src/output.c
+++ b/libvmaf/src/output.c
@@ -44,6 +44,26 @@ static const char *pool_method_name[] = {
     [VMAF_POOL_METHOD_HARMONIC_MEAN] = "harmonic_mean",
 };
 
+int countLeadingZeroesFloat(double x)
+{   
+    if(x < 0)
+        x = fabs(x);
+
+    int intPart = (int)x;
+    double fractionalPart = x - intPart;
+
+    // Count leading zeroes in the fractional part
+    int leadingZeroesCount = 0;
+
+    while (fractionalPart < 1.0 && fractionalPart != 0)
+    {
+        fractionalPart *= 10; // Shift decimal point to the right
+        leadingZeroesCount++;
+    }
+
+    return leadingZeroesCount;
+}
+
 int vmaf_write_output_xml(VmafContext *vmaf, VmafFeatureCollector *fc,
                           FILE *outfile, unsigned subsample, unsigned width,
                           unsigned height, double fps, unsigned pic_cnt)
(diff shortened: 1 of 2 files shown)


nilfm99:
Does the dynamic range of the feature span more than 6 orders of magnitude? If so, then I guess this is the way to go, cc @kylophone for another pair of eyes.

In any case, could this be updated to use `snake_case` in both function name and variable names throughout to match the rest of the code?

libvmaf/src/output.h
```diff
@@ -19,6 +19,8 @@
 #ifndef __VMAF_OUTPUT_H__
 #define __VMAF_OUTPUT_H__
 
+int countLeadingZeroesFloat(float x);
```
nilfm99, line 22: I think this function could be `static` and then the header doesn't need to declare it

libvmaf/src/output.c
```diff
@@ -44,6 +44,26 @@ static const char *pool_method_name[] = {
     [VMAF_POOL_METHOD_HARMONIC_MEAN] = "harmonic_mean",
 };
 
+static int count_leading_zeros_flat(double x)
```
nilfm99, line 47: nit: I think this used to be `float` and got typoed to `flat`? in either case, it is a bit weird to call it `float` and have it operate on doubles. There are some uses of the "`_s` for single, `_d` for double" convention in this repo, such as [here](https://github.com/Netflix/vmaf/blob/262db8dbdb6210d2f5aaec36b77a1fb76e9c9823/libvmaf/src/feature/adm_tools.h#L278-L280) - maybe `count_leading_zeros_d`?

libvmaf/src/output.c
```diff
@@ -44,6 +44,26 @@ static const char *pool_method_name[] = {
...
+{
+    if(x < 0)
+        x = fabs(x);
+
+    int intPart = (int)x;
```
nilfm99, line 52: nit: a couple more snake_case candidates

libvmaf/src/output.c
cosmin: welp, that was definitely a typo. Renamed to `count_leading_zeros_d` as suggested.

nilfm99 (approved):
LGTM, let's give it a few days in case @kylophone has any comments and merge

PR #1360, opened 2024-04-22
increase max feature count to 64

cosmin:
During the implementation of FUNQUE+ feature extractors we ran into the 32 max limit for JSON features and hand to bump this up to 64. In practice this may not be needed as not all exposed features are likely to be used, but we needed this to exhaustively exercise the feature extractor.

Co-authored-by: Mallikarjun Kamble <mallikarjun.kamble@ittiam.com>


nilfm99 (approved):
LGTM

PR #1361, opened 2024-04-23
Implement SAST resizer from FUNQUE

cosmin:
The FUNQUE implementation at https://github.com/cosmin/vmaf/tree/funque contains an optional SAST resizer which downscales the input image dimensions by a factor of 2.

This brings in that implementation of the resizer with an optimized code path for 0.5x downscaling with AVX2, AVX512 and NEON SIMD implementation.

In theory this could be used by more than just the FUNQUE extractor, for example SSIM can benefit from the same SAST transform[1] and we could use this for an optimized standalone eSSIM feature extractor.

[1] A. K. Venkataramanan, C. Wu, A. C. Bovik, I. Katsavounidis and Z. Shahid, "A Hitchhiker’s Guide to Structural Similarity," in IEEE Access, vol. 9, pp. 28872-28896, 2021, doi: 10.1109/ACCESS.2021.3056504 https://ieeexplore.ieee.org/document/9344648


nilfm99:
Hi, sorry for the delay here, I have looked at this a few times and haven't found any significant issues. Do you have any tests on your branches that could give us extra confidence in the changes?

Issue #1362, opened 2024-04-23
tests: linking with ldd causes issues with missing pthread symbols


1480c1:
Using clang64 with msys2 (thus using clang+ldd by default) shows the following error

```c
[127/172] Linking target test/test_feature_extractor.exe
FAILED: test/test_feature_extractor.exe
"cc" @test/test_feature_extractor.exe.rsp
ld.lld: error: undefined symbol: pthread_mutex_init
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_create)
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_collector.c.obj:(vmaf_feature_collector_init)

ld.lld: error: undefined symbol: pthread_mutex_lock
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_aquire)
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_release)
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_flush)
>>> referenced 6 more times

ld.lld: error: undefined symbol: pthread_cond_init
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_aquire)

ld.lld: error: undefined symbol: pthread_cond_wait
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_aquire)

ld.lld: error: undefined symbol: pthread_mutex_unlock
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_aquire)
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_release)
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_flush)
>>> referenced 5 more times

ld.lld: error: undefined symbol: pthread_cond_signal
>>> referenced by src/liblibvmaf_feature.a.p/feature_feature_extractor.c.obj:(vmaf_fex_ctx_pool_release)

[...]

PR #1363, opened 2024-04-23
test_feature_extractor: add thread_lib dependency

1480c1:
msys2's clang64 environment errors out due to undefined symbols with pthread.
The non-clang environments are fine as gcc implicitly links pthread.

Fixes https://github.com/Netflix/vmaf/issues/1362


1480c1:
image

Compilation and test works locally. Also checked with ucrt64 and mingw64 and they all pass.

nilfm99:
I opened #1364 to fix the ffmpeg CI issue, do you know why the other tests may be failing?

nilfm99:
alright, I see that the other failure also has nothing to do with this PR, no idea how it got introduced 😞 

This one LGTM, we will get it in after fixing that error

nilfm99 (approved):

PR #1369, opened 2024-06-11
Faster process pool

101arrowz:
Improves the performance of the process pool by replacing the polling-based approach with Python's built-in process pool executor, which reduces synchronization overhead and reuses the Python interpreter when possible instead of spinning up a new instance for every single task.

### DIFF
--- a/python/vmaf/core/executor.py
+++ b/python/vmaf/core/executor.py
@@ -182,7 +182,7 @@ class Executor(TypeVersionEnabled):
                 lock.release()
                 return result
 
-            self.results = parallel_map(_run, list_args, processes=processes, sleep_sec=0.1)
+            self.results = parallel_map(_run, list_args, processes=processes)
         else:
             self.results = list(map(self._run_on_asset, self.assets))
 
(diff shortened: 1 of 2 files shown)


nilfm99:
I think the build failures are on Github's side, not ours. See https://github.com/actions/runner-images/issues/6131 and https://www.githubstatus.com/

101arrowz:
The build failures seem to be due to an internal error with GitHub Actions, but I noticed there's a test for Windows support. As mentioned in the diff, neither the old implementation nor the new one works with kernels that do not support POSIX fork(), NT included. However, the old implementation would silently use a spawn-based method (and might thereby fail while pickling the arguments), while the new one fails fast due to an assertion. In the specific situation that all arguments are pickleable and the code is running on Windows, however, the old code would work fine while the new version does not.

nilfm99:
I think this should be okay, the python wrapper is secondary to the C library IMO, and Linux/Mac is our main focus. Tagging @christosbampis, @krasuluk and @li-zhi who are the main users of the python library in case they have objections.

nilfm99 (approved):

PR #1373, opened 2024-06-30
[bugfix] Print required options hint for vmaf tool correctly

Cleanbit25:
This seems to be a copy and paste bug.

Calling "vmaf" with only -r option prints "could not open file: (null)". Instead it should print the correct usage hint.


nilfm99 (approved):
Great catch, thanks for the contribution!

PR #1374, opened 2024-07-02
GSoC 2024: libvmaf: Propagate metadata

yigithanyigit:
Part of the [GSoC 2024](https://summerofcode.withgoogle.com/programs/2024/projects/yma73MGc) project.

Supports propagating metadata with modified `feature_vector_append`.

### DIFF
--- a/libvmaf/include/libvmaf/libvmaf.h
+++ b/libvmaf/include/libvmaf/libvmaf.h
@@ -254,6 +254,51 @@ int vmaf_feature_score_at_index(VmafContext *vmaf, const char *feature_name,
                                 double *score, unsigned index);
 
 /**
+ * Metadata structure.
+ *
+ * @param feature_name   Name of the feature to fetch.
+ *
+ * @param picture_index  Picture index.
+ *
+ * @param score          Score.
+ *
+ * @note This structure is used to pass metadata to a callback function.
+ */
+typedef struct VmafMetadata {
+    char *feature_name;
+    unsigned picture_index;
+    double score;
+} VmafMetadata;
+
+/**
+ * Metadata configuration.
+ *
+ * @param feature_name Name of the feature to fetch.
+ *
+ * @param callback     Callback to receive metadata.
+ *
+ * @param data         User data to pass to the callback.
+ */
+typedef struct VmafMetadataConfiguration {
+    char *feature_name;
+    void (*callback)(void *data, VmafMetadata *metadata);
+    void *data;
+} VmafMetadataConfiguration;
+
+/**
+ * Register a callback to receive VMAF metadata.
+ *
+ * @param vmaf The VMAF context allocated with `vmaf_init()`.
+ *
+ * @param cfg  Metadata configuration.
+ *
+ *
+ * @return 0 on success, or < 0 (a negative errno code) on error.
+ */
+
+int vmaf_register_metadata_handler(VmafContext *vmaf, VmafMetadataConfiguration cfg);
+
+/**
  * Pooled VMAF score for a specific interval.
  *
  * @param vmaf         The VMAF context allocated with `vmaf_init()`.
--- a/libvmaf/src/feature/feature_collector.h
+++ b/libvmaf/src/feature/feature_collector.h
@@ -24,6 +24,8 @@
 #include <time.h>
 
 #include "dict.h"
+#include "model.h"
+#include "metadata_handler.h"
 
 typedef struct {
     char *name;
@@ -42,9 +44,16 @@ typedef struct {
     unsigned cnt, capacity;
 } AggregateVector;
 
+typedef struct VmafPredictModel {
+    VmafModel *model;
+    struct VmafPredictModel *next;
+} VmafPredictModel;
+
 typedef struct VmafFeatureCollector {
     FeatureVector **feature_vector;
     AggregateVector aggregate_vector;
+    VmafCallbackList *metadata;
+    VmafPredictModel *models;
     unsigned cnt, capacity;
     struct { clock_t begin, end; } timer;
     pthread_mutex_t lock;
@@ -52,9 +61,14 @@ typedef struct VmafFeatureCollector {
 
 int vmaf_feature_collector_init(VmafFeatureCollector **const feature_collector);
 
+int vmaf_feature_collector_mount_model(VmafFeatureCollector *feature_collector, VmafModel *model);
+
 int vmaf_feature_collector_append(VmafFeatureCollector *feature_collector,
                                   const char *feature_name, double score,
                                   unsigned index);
+
+int vmaf_feature_collector_register_metadata(VmafFeatureCollector *feature_collector,
+                                             VmafMetadataConfiguration metadata_cfg);
 
 int vmaf_feature_collector_append_with_dict(VmafFeatureCollector *fc,
         VmafDictionary *dict, const char *feature_name, double score,
--- a/libvmaf/src/predict.c
+++ b/libvmaf/src/predict.c
@@ -228,6 +228,7 @@ int vmaf_predict_score_at_index(VmafModel *model,
                                 VmafFeatureCollector *feature_collector,
                                 unsigned index, double *vmaf_score,
                                 bool write_prediction,
+                                bool propagate_metadata,
                                 enum VmafModelFlags flags)
 {
     if (!model) return -EINVAL;
@@ -287,9 +288,11 @@ int vmaf_predict_score_at_index(VmafModel *model,
                                                index);
 
         if (err) {
-            vmaf_log(VMAF_LOG_LEVEL_ERROR,
-                     "vmaf_predict_score_at_index(): no feature '%s' "
-                     "at index %d\n", feature_name, index);
+            if (!propagate_metadata) {
+              vmaf_log(VMAF_LOG_LEVEL_ERROR,
+                       "vmaf_predict_score_at_index(): no feature '%s' "
+                       "at index %d\n", feature_name, index);
+            }
             free(feature_name);
             goto free_node;
         }
@@ -366,7 +369,7 @@ static int vmaf_bootstrap_predict_score_at_index(
         err = vmaf_predict_score_at_index(model_collection->model[i],
                                           feature_collector, index,
                                           &scores[i], false,
-                                          flags);
+                                          false, flags);
         if (err) return err;
 
         // do not override the model's transform/clip behavior
@@ -374,7 +377,7 @@ static int vmaf_bootstrap_predict_score_at_index(
         double score;
         err = vmaf_predict_score_at_index(model_collection->model[i],
                                           feature_collector, index,
-                                          &score, true, 0);
+                                          &score, true, false, 0);
         if (err) return err;
     }
 
--- a/libvmaf/test/meson.build
+++ b/libvmaf/test/meson.build
@@ -17,8 +17,13 @@ test_picture = executable('test_picture',
     dependencies:[stdatomic_dependency, thread_lib, cuda_dependency],
 )
 
+test_propagate_metadata = executable('test_propagate_metadata',
+    ['test.c', 'test_propagate_metadata.c', '../src/metadata_handler.c'],
+    include_directories : [libvmaf_inc, test_inc, include_directories('../src/')],
+)
+
 test_feature_collector = executable('test_feature_collector',
-    ['test.c', 'test_feature_collector.c', '../src/log.c'],
+    ['test.c', 'test_feature_collector.c', '../src/log.c', '../src/predict.c', '../src/svm.cpp', '../src/metadata_handler.c'],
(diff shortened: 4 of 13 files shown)


kylophone:
Thanks @yigithanyigit, will add my review tomorrow morning.

libvmaf/include/libvmaf/libvmaf.h
```diff
@@ -63,25 +63,28 @@ enum VmafPoolingMethod {
...
- * 
+ *
  * @param gpumask     Restrict permitted GPU operations.
  *                    if gpumask: disable CUDA
  */
 typedef struct VmafConfiguration {
     enum VmafLogLevel log_level;
+    void (*callback)(void*, const char *, double);
+    void* meta_data;
+    size_t meta_data_sz;
```
kylophone, line 87: Rather than configure this with  `VmafConfiguration` could we please introduce another api function to register these callbacks? The function should take an already allocated VmafContext and it's own configuration struct. If there is a use case for registering more than just a single callback, we should be able to support that by calling this new api more than once.

yigithanyigit: Should a new function automatically unregister existing callback when attempting to register itself, or should we define a new function for unregistering callbacks?

kylophone: When we register the callback, the `VmafContext` should take ownership of the callback/associated memory, when the `VmafContext` is destroyed, anything related to the callback(s) can be cleaned up. The ownership of data passed into the callback I think still belongs with the user. Thoughts?

yigithanyigit: Thank you for the explanation! This approach looks logical and helps maintain responsibility boundaries.

I am planning a to use a linked list structure something like this;
```c
typedef struct VmafPropagateMetadataNode {
       void (*callback)(void *, const char *, double);
       void* meta_data;
       size_t meta_data_sz;
       VmafPropagateMetadataNode *next;
}
```

With that we could support multiple callbacks as you told, What do you think?

And could you say what do you think about the parameters of callback?

```c
void (*callback)(void*, const char *, double);
```

I can take an offset value instead of `const char *` and `double` and I can change with `offset` value. However current one looks more sensible to me.

libvmaf/src/feature/feature_collector.h
```diff
@@ -42,15 +43,32 @@ typedef struct {
     unsigned cnt, capacity;
 } AggregateVector;
 
+typedef struct {
+        void (*callback)(void *, const char * ,double);
+        void *data;
+        size_t data_sz;
+} VmafMetadata;
+
+typedef struct VmafPredictModel VmafPredictModel;
```
nilfm99, line 52: Nit: you can avoid the duplicated typedef by using `struct VmafPredictModel *next` inside the `VmafPredictModel` definition (usages outside of this definition can remain the same)

libvmaf/src/predict.c
```diff
@@ -287,9 +288,11 @@ int vmaf_predict_score_at_index(VmafModel *model,
                                                index);
 
         if (err) {
+            if (!propagate_metadata) {
```
nilfm99, line 291: I'd like to understand this better - are there plans to make more changes to this function based on this flag, or will it just be used to suppress logs in case of error?

libvmaf/src/predict.c
```diff
@@ -287,9 +288,11 @@ int vmaf_predict_score_at_index(VmafModel *model,
...
             vmaf_log(VMAF_LOG_LEVEL_ERROR,
```
nilfm99, line 292: nit: this block should be indented one more level

libvmaf/src/predict.c
yigithanyigit: In my case I used to suppress the logs because when calculating the `vmaf` score, each feature appends to the collector and attempts to access non-initialized features and is causing numerous errors. However, I am aware that this flag prevents visibility of 'real' errors, I am open to discuss that.

libvmaf/src/feature/feature_collector.h
yigithanyigit: Thanks for the feedback! You are right, fixed in local.

libvmaf/test/meson.build
```diff
@@ -17,8 +17,13 @@ test_picture = executable('test_picture',
     dependencies:[stdatomic_dependency, thread_lib, cuda_dependency],
 )
 
+test_propagate_metadata = executable('test_propagate_metadata',
+    ['test.c', 'test_propagate_metadata.c', '../src/propagate_metadata.c'],
```
nilfm99, line 21: Heads up: the build is failing because `test_propagate_metadata` doesn't exist

libvmaf/include/libvmaf/propagate_metadata.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef __VMAF_PROPAGATE_METADATA_H__
+#define __VMAF_PROPAGATE_METADATA_H__
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stddef.h>
+
+typedef struct VmafMetadataConfig {
```
kylophone, line 28: Let's keep this all inside `libvmaf.h`, no need for this separate header.

libvmaf/test/test_psnr.c
```diff
@@ -84,7 +85,7 @@ static char *test_16b_large_diff()
     mu_assert("wrong mse_y", almost_equal(mse_y, 4294836225.0));
     mu_assert("wrong mse_cb", almost_equal(mse_cb, 4294836225.0));
     mu_assert("wrong mse_cr", almost_equal(mse_cr, 4294836225.0));
-
```
kylophone, line 87: In general we should avoid introducing unrelated changes into the diff, makes it much harder to read.

libvmaf/test/test_propagate_metadata.c
```diff
@@ -0,0 +1,95 @@
...
+                                                         propagate_metadata ||
+                                                         !propagate_metadata->head);
+    return NULL;
+}
+
+static char *test_propagate_metadata_append()
+{
+    VmafMetadata *propagate_metadata;
+    int err = vmaf_metadata_init(&propagate_metadata);
+    mu_assert("problem during vmaf_propagate_metadata_init", !err ||
```
kylophone, line 36: For debugging the tests, 3 separate asserts here is much easier to follow. It lets you know exactly which condition is failing in a very clear way. Ditto for the similar cases below.

kylophone:
I still need to spend some time with your tests to see how this all works.

Also, in general, I think the `propagate_metadata` naming of this api is not the greatest since it is flexible enough to have other uses outside of ffmpeg AVFrame metadata propagation. Don't worry about changing this now, we can save that for the end.

libvmaf/test/test_predict.c
```diff
@@ -49,14 +57,120 @@ static char *test_predict_score_at_index()
...
+    VmafDictionary *dict = NULL;
+    MetaStruct meta_data = {
+        .metadata = &dict,
+        .flags    = 0,
+    };
+
+    VmafMetadataConfig m = {
+        .callback = set_meta,
+        .data     = &meta_data,
+        .data_sz  = sizeof(meta_data),
```
kylophone, line 91: Doesn't seem like `data_sz` is used anywhere. Since this API should not be in charge of allocating `data`, I don't think we need this.

kylophone:
@yigithanyigit Looks like there is a use after free in `test_model_mount`.

You can enable address sanitizer via meson like this: `LDFLAGS=-fsanitize=address meson build --buildtype debug -Db_sanitize=address`

```
test_feature_vector_init_append_and_destroy: libvmaf WARNING feature "psnr_y" cannot be overwritten at index 8
[32mpass[0m
test_feature_collector_init_append_get_and_destroy: [32mpass[0m
test_aggregate_vector_init_append_and_destroy: [32mpass[0m
test_model_mount: =================================================================
==7613==ERROR: AddressSanitizer: heap-use-after-free on address 0x60c000000ba0 at pc 0x00010d57ed3a bp 0x7ff7b298e1f0 sp 0x7ff7b298e1e8
READ of size 8 at 0x60c000000ba0 thread T0
    #0 0x10d57ed39 in test_model_mount test_feature_collector.c:76
    #1 0x10d57c322 in run_tests test_feature_collector.c:251
    #2 0x10d572753 in main test.c:26
    #3 0x7ff8144f1365 in start+0x795 (dyld:x86_64+0xfffffffffff5c365)

0x60c000000ba0 is located 32 bytes inside of 128-byte region [0x60c000000b80,0x60c000000c00)
freed by thread T0 here:
    #0 0x10e218b69 in wrap_free+0xa9 (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0xdcb69)
    #1 0x10d57620e in vmaf_feature_collector_destroy feature_collector.c:443
    #2 0x10d57ecde in test_model_mount test_feature_collector.c:75
    #3 0x10d57c322 in run_tests test_feature_collector.c:251
    #4 0x10d572753 in main test.c:26
    #5 0x7ff8144f1365 in start+0x795 (dyld:x86_64+0xfffffffffff5c365)

previously allocated by thread T0 here:
    #0 0x10e218a20 in wrap_malloc+0xa0 (libclang_rt.asan_osx_dynamic.dylib:x86_64h+0xdca20)
    #1 0x10d5733a6 in vmaf_feature_collector_init feature_collector.c:210
    #2 0x10d57e968 in test_model_mount test_feature_collector.c:57
    #3 0x10d57c322 in run_tests test_feature_collector.c:251
    #4 0x10d572753 in main test.c:26
    #5 0x7ff8144f1365 in start+0x795 (dyld:x86_64+0xfffffffffff5c365)

SUMMARY: AddressSanitizer: heap-use-after-free test_feature_collector.c:76 in test_model_mount
Shadow bytes around the buggy address:
  0x60c000000900: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x60c000000980: fd fd fd fd fd fd fd fd fa fa fa fa fa fa fa fa
  0x60c000000a00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x60c000000a80: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x60c000000b00: fd fd fd fd fd fd fd fd fa fa fa fa fa fa fa fa
=>0x60c000000b80: fd fd fd fd[fd]fd fd fd fd fd fd fd fd fd fd fd
  0x60c000000c00: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x60c000000c80: fd fd fd fd fd fd fd fd fa fa fa fa fa fa fa fa
  0x60c000000d00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x60c000000d80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x60c000000e00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==7613==ABORTING
```

libvmaf/include/libvmaf/libvmaf.h
```diff
@@ -253,6 +253,25 @@ int vmaf_score_at_index_model_collection(VmafContext *vmaf,
 int vmaf_feature_score_at_index(VmafContext *vmaf, const char *feature_name,
                                 double *score, unsigned index);
 
+typedef struct VmafMetadataConfig {
+    void (*callback)(void *, const char *, double);
```
kylophone, line 257: Rather than passing two extra parameters the callback `(char*, double)` let's pass another struct. For now, the struct can contain these two `(char*, double)` but it is much more flexible if we need to add anything in the future, since we can do so without breaking API.

libvmaf/include/libvmaf/libvmaf.h
```diff
@@ -253,6 +253,31 @@ int vmaf_score_at_index_model_collection(VmafContext *vmaf,
...
+    void *data;
+} VmafMetadataConfiguration;
+
+/**
+ * Register a callback to receive VMAF metadata.
+ *
+ * @param vmaf The VMAF context allocated with `vmaf_init()`.
+ *
+ * @param cfg  Metadata configuration.
+ *
```
kylophone, line 274: update api docs here to include `data`

nilfm99:
This new failure seems legitimate.

yigithanyigit:
> This new failure seems legitimate.

Issue looks like related our temporary `vmaf.c` changes, I am working on it.

Thanks

Issue #1309, opened 2024-01-12
version mismatch between build version for libvmaf and the python code


SingingBush:
as mentioned in #1307, the version for the meson build is defined in `libvmaf/meson.build` and the version for the python library comes from `python/vmaf/__init__.py`. This means it's likely for them to get out of sync, which is currently the case in master.

jacoblubecki:
I'm open to taking a stab at this, but I wanted to also double check that it makes sense -- does the release process guarantee that the python code will reflect support for the co-bundled libvmaf code version? I could imagine a scenario where the libvmaf version gets released first and then the python bindings are updated later to match.

nilfm99:
I think we are never going to intentionally release in such a way that the python wrapper is incompatible with the bundled libvmaf version. If we deprecate an API, we won't remove it immediately and there will be a transition period in which the python wrapper can be updated before it is removed. So if the python version number is made to automatically track the libvmaf one, that seems correct to me.

pinging @kylophone to verify this

nilfm99:
We have discussed this and we think decoupling the C and python versions makes the most sense so they can progress separately without needing to bump one of them for changes to the other. Currently they are still the same but they may/will diverge. Closing this thread since there is no need for the synchronization anymore. Thanks for raising this issue.

PR #1376, opened 2024-07-05
Fix fifo hangs

nilfm99:


101arrowz:
Are there any public consumers of the Python interface that will be affected by the function signature change, and/or do we need some sort of minor revision bump to merge this?

101arrowz (approved):

nilfm99:
I'm not sure @101arrowz, I think we have kept proper versions of the C code but not the python code (see #1309 which is asking about a similar thing). We have a version [here](https://github.com/Netflix/vmaf/blob/5a48363d5ea0948f30b80907c3e4c9146286e81f/python/vmaf/__init__.py#L7) that we could bump to 3.1. Tagging @kylophone and @li-zhi in case they know more about this versioning history.

Issue #1379, opened 2024-07-11
VMAF CUDA build fails in Ubuntu 22.04LTS Docker image


Brainiarc7:
Hello there,

I've encountered a build error with libvmaf in an Ubuntu 22.04LTS Docker image where the NVIDIA CUDA SDK is installed via apt. Related to #1176 .

Specifically, the build dies on:

```
FAILED: tools/vmaf 
c++  -o tools/vmaf tools/vmaf.p/vmaf.c.o tools/vmaf.p/cli_parse.c.o tools/vmaf.p/y4m_input.c.o tools/vmaf.p/vidinput.c.o tools/vmaf.p/yuv_input.c.o -Wl,--as-needed -Wl,--no-undefined -Wl,-O1 -Wl,--start-group src/libvmaf.a -lcuda /usr/local/cuda/lib64/libcudart_static.a -lrt -lpthread -ldl -pthread -lm -lcuda -Wl,--end-group
/usr/bin/ld: cannot find -lcuda: No such file or directory
/usr/bin/ld: cannot find -lcuda: No such file or directory
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
```

The build environment avails CUDA in the system path, as shown below:

```
printenv PATH
/usr/local/cuda/bin:/root/.cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```
```
printenv CUDA_HOME
/usr/local/cuda
```

And from `ldconfig -p | grep cuda`, we get:

```
	libpcsamplingutil.so (libc6,x86-64) => /usr/local/cuda/targets/x86_64-linux/lib/libpcsamplingutil.so
	libnvrtc.so.12 (libc6,x86-64) => /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so.12
	libnvrtc.so (libc6,x86-64) => /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc.so
	libnvrtc-builtins.so.12.3 (libc6,x86-64) => /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc-builtins.so.12.3
	libnvrtc-builtins.so (libc6,x86-64) => /usr/local/cuda/targets/x86_64-linux/lib/libnvrtc-builtins.so
	libnvperf_target.so (libc6,x86-64) => /usr/local/cuda/targets/x86_64-linux/lib/libnvperf_target.so
[...]

PR #1380, opened 2024-07-12
Update Dockerfile.cuda

Brainiarc7:
This patch is a fix to #1379, wherein:

1. The CUDA library stubs are added to `ldconfig`'s cache.
2. The build type for `libvmaf` is set to `shared`.
3. The included models are also copied to `/usr/local/share/model` for re-usability with custom FFmpeg pipelines as needed.
4. `ldconfig` is called up immediately after `libvmaf` build so that `FFmpeg` can detect it.

### DIFF
--- a/Dockerfile.cuda
+++ b/Dockerfile.cuda
@@ -13,11 +13,30 @@ RUN git clone https://github.com/FFmpeg/FFmpeg.git && cd FFmpeg && git checkout 
 
 RUN git clone https://github.com/FFmpeg/nv-codec-headers.git && cd nv-codec-headers && make && make install
 
+#Add CUDA Stubs to ldconfig:
+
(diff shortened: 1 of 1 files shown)


guyoun:
I think the FFmpeg build failure is due to the missing libcuda.so.1 library file. 

The vmaf is linked with libcuda.so.1 because of SONAME of libcuda.so is `libcuda.so.1`
When enable libvmaf of FFmpeg, it requires libcuda.so.1 but this file does not exist when building the Docker image. 
I am unsure why the SONAME of /usr/local/cuda/lib64/stubs/libcuda.so is libcuda.so.1. 
According to [this issue](https://github.com/NVIDIA/nvidia-container-toolkit/issues/185) on the NVIDIA GitHub repository, there are no plans to fix this.

liuxunchenglxc (approved):

PR #1383, opened 2024-08-07
Fix ARG placement in Dockerfile.cuda

guyoun:
```
An ARG declared before a FROM is outside of a build stage, so it can't be used in any instruction after a FROM
```
From https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact


nilfm99:
This looks correct, thank you. Sorry about the build failures, I am looking into them separately. We will also fix the typo `FFFMPEG` -> `FFMPEG`, feel free to make that change yourself or I will make a follow up PR for it.

nilfm99 (approved):

PR #1384, opened 2024-08-07
Fix build by pinning libsvm-official to <=3.32

nilfm99:
The build has started failing with this error: 
```
TypeError: <lambda>() got an unexpected keyword argument 'nopython'
```

This is coming from the libsvm module. Past builds have succeeded with 3.32. Trying to pin to see if the upgrade to 3.33 has caused the failures.


nilfm99:
Raised an issue in libsvm repo: https://github.com/cjlin1/libsvm/issues/218

PR #1387, opened 2024-09-11
Modify metadata_register API for support individual metric propagation.

yigithanyigit:
In current state of this API it can only propagate explicitly given features, In other words giving an identifier doesn't work.

For example; for propagating `PSNR`:
 `psnr_y`, `psnr_cb`, `psnr_cr` should be given.

With these changes, propagating individual features and propagating with identifier possible.


kylophone:
I'm not sure if we need an API like this yet (?). I will hold off on the reviewing this PR until we figure that out. In the meantime, I will reply to your patch in ffmpeg-devel.

yigithanyigit:
Thanks for the comment,

I found some problems about this patch and I wanted to add for future reference.

After some tests I realized this patch doesn't work expected in some identifiers (`VIF`, `ADM`).
The main reason of the problem those identifiers has some conditional features.

For example:

https://github.com/Netflix/vmaf/blob/d95b69e03d9401ad9f3532eaab14066427761c35/libvmaf/src/feature/integer_adm.c#L2707-L2741

https://github.com/Netflix/vmaf/blob/d95b69e03d9401ad9f3532eaab14066427761c35/libvmaf/src/feature/integer_vif.c#L692-L746

As you see those features going to append if `debug` enabled. However my implementation going to add those features anyway:
 
https://github.com/Netflix/vmaf/blob/d95b69e03d9401ad9f3532eaab14066427761c35/libvmaf/src/feature/integer_adm.c#L2757-L2767

We should handle this somehow. Maybe we can add an option to `VmafMetadataConfiguration`.

Thanks

yigithanyigit:
I significantly refactored my patch to escape thread synchronization issue

- Changed from per-feature propagation to propagating metadata when VMAF
scores are complete
- Added frame index tracking in metadata handler to ensure ordered
processing
- Added test coverage for non-monotonic frame processing scenarios

This change eliminates the need for complex reordering logic at the caller level and reduces potential thread synchronization issues by only propagating metadata when all necessary metrics are calculated.

yigithanyigit:
@nilfm99 are those failures related to me? It seems like the tox tests are stuck.

nilfm99:
@yigithanyigit I don't see why CI is failing, my [test PR](https://github.com/Netflix/vmaf/pull/1404) is passing - would it be okay if we try rebasing your branch on master?

yigithanyigit:
There was a bug that triggers recursive callback calls due to `vmaf_predict_score_at_index` calling internal `vmaf_feature_vector_append`. I reverted until new version of the patch.

PR #1389, opened 2024-10-10
libvmaf: fix case of windows.h

alyssais:
mingw provides a windows.h header, not Windows.h.  This causes a build failure when cross compiling from a Unix system with a case sensitive filesystem.


XuShiSheng:
你好！邮件已收到，我会尽快回复你--许思焱

nilfm99:
Thanks, this looks good to me.

PR #1407, opened 2025-02-02
Simply metadata propagation API.

yigithanyigit:
This PR continuation of #1387. Since somehow it's stuck with tests I decided to write changes in clean branch.

Main summary of PR:

- Changed from per-feature propagation to propagating metadata when VMAF
scores are complete
- Added frame index tracking in metadata handler to ensure ordered
processing
- Added test coverage for non-monotonic, non-monotonic frame processing scenarios
- Added test coverage for multiple callback configuration, multiple model scenarios.
- moved some tests inside `test_predict` to `test_propagate_metadata` for better test separation.

This change eliminates the need for complex reordering logic at the caller level and reduces potential thread synchronization issues by only propagating metadata when all necessary metrics are calculated.

Most of the additions coming from tests. I tried to implement the all possible edge cases that can occur carefully.

### DIFF
--- a/libvmaf/src/feature/feature_collector.c
+++ b/libvmaf/src/feature/feature_collector.c
@@ -31,6 +31,8 @@
 #include "log.h"
 #include "predict.h"
 
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+
 static int aggregate_vector_init(AggregateVector *aggregate_vector)
 {
     if (!aggregate_vector) return -EINVAL;
@@ -245,6 +247,8 @@ int vmaf_feature_collector_mount_model(VmafFeatureCollector *feature_collector,
 
     m->model = model;
     m->next = NULL;
+    m->last_highest_seen_index = 0;
+    m->last_lowest_seen_index = 0;
 
     VmafPredictModel *head = feature_collector->models;
     if (!head) {
@@ -354,30 +358,12 @@ int vmaf_feature_collector_append(VmafFeatureCollector *feature_collector,
 
     int res = 0;
 
-    VmafCallbackItem *metadata_iter = feature_collector->metadata ?
-                                      feature_collector->metadata->head : NULL;
-    while (metadata_iter) {
-        // Check current feature name is the same as the metadata feature name
-        if (!strcmp(metadata_iter->metadata_cfg.feature_name, feature_name)) {
+    VmafPredictModel *model_iter = feature_collector->models;
 
-            // Call the callback function with the metadata feature name
-            VmafMetadata data = {
-                .feature_name = metadata_iter->metadata_cfg.feature_name,
-                .picture_index = picture_index,
-                .score = score,
-            };
-            metadata_iter->metadata_cfg.callback(metadata_iter->metadata_cfg.data, &data);
-            // Move to the next metadata
-            goto next_metadata;
-        }
+    while (model_iter && feature_collector->metadata->cnt) {
+        VmafModel *model = model_iter->model;
 
-        VmafPredictModel *model_iter = feature_collector->models;
-
-        // If metadata feature name is not the same as the current feature feature_name
-        // Check if metadata feature name is the predicted feature
-        while (model_iter) {
-            VmafModel *model = model_iter->model;
-
+        if (strcmp(model->name, feature_name)) {
             pthread_mutex_unlock(&(feature_collector->lock));
             res = vmaf_feature_collector_get_score(feature_collector,
                     model->name, &score, picture_index);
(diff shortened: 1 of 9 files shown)


libvmaf/test/test_propagate_metadata.c
```diff
@@ -16,10 +16,51 @@
...
+    VmafDictionary **metadata;
+    int flags;
+} MetaStruct;
+
+static void set_meta(void *data, VmafMetadata *metadata)
+{
+    if (!data) return;
+    MetaStruct *meta = data;
+    char key[128], value[128];
+    snprintf(key, sizeof(value), "%s_%d", metadata->feature_name,
```
kylophone, line 40: `sizeof(key)`

libvmaf/src/feature/feature_collector.h
```diff
@@ -46,6 +46,8 @@ typedef struct {
 
 typedef struct VmafPredictModel {
     VmafModel *model;
+    unsigned last_highest_seen_index;
+    unsigned last_lowest_seen_index;
```
kylophone, line 50: Can you explain why these index tracking fields are needed?

libvmaf/test/test_predict.c
```diff
@@ -19,19 +19,13 @@
...
 
 #include <libvmaf/model.h>
 #include <math.h>
-
-typedef struct {
-    VmafDictionary **metadata;
-    int flags;
-} MetaStruct;
-
+#include <string.h>
```
kylophone, line 28: Did you mean to add this header?

libvmaf/src/feature/feature_collector.h
yigithanyigit: Of course, as you know if application is multithreaded model scores might be calculated in out of order. Those indexes basically helping them to make in order. I am using those indexes for the current calculated index is in order. For example;

Let's say we have frame order like this:

3, 0, 1, 2

Frame 3 arrives -> Calculate but not propagate, Update `last_highest_seen_index` to 3;
Propagated frame -> None

Frame 0 arrives -> Calculate and propagate since frame index (0) == last_lowest_seen_index. Increase `last_lowest_seen_index` by 1. It tries to propagate until `last_highest_seen_index (3)`  but since there is no Frame 1 it stops.
Propagated frame -> Frame 0

Frame 1 arrives -> Calculate and propagate since frame index (1) == last_lowest_seen_index. Increase `last_lowest_seen_index` by 1. It tries to propagate until `last_highest_seen_index` but since there is no Frame 2 it stops.
Propagated frame -> Frame 1

Frame 2 arrives -> Calculate and propagate since frame index (2) == last_lowest_seen_index. Increase `last_lowest_seen_index` by 1. It tries to propagate until `last_highest_seen_index` and we have ready Frame 3 so also it propagates that frame.
Propagated frame -> Frame 0, Frame 1

libvmaf/test/test_predict.c
yigithanyigit: Ups, no. Sorry. I'll fix asap.

libvmaf/src/feature/feature_collector.h
kylophone: Is it a problem if the callbacks come out of order? Using your method, callbacks could be withheld for an unlimited number of frames (1 - N) until frame 0 is ready?

yigithanyigit: > Is it a problem if the callbacks come out of order?

I don't think it's a problem for us. But for the sake of usability of API I think we need something like this. Otherwise users might need to reorder the frames. For example I am thinking FFmpeg, in that implementation we need to send frames in order to filtergraph (If we want to write features to original frame's metadata otherwise we don't need).

> Using your method, callbacks could be withheld for an unlimited number of frames (1 - N) until frame 0 is ready?

Unfortunately, yes. Is there a possibility that frame 0 somehow not calculated? I thought this situation while implementing this and I concluded if somehow one of the frames doesn't calculate then we can't calculate any other future frames due to metrics like `Motion` right? So I ended up VMAF can't drop frames.

libvmaf/src/feature/feature_collector.c
```diff
@@ -347,30 +357,12 @@ int vmaf_feature_collector_append(VmafFeatureCollector *feature_collector,
...
+    VmafPredictModel *model_iter = feature_collector->models;
 
-        // If metadata feature name is not the same as the current feature feature_name
-        // Check if metadata feature name is the predicted feature
-        while (model_iter) {
-            VmafModel *model = model_iter->model;
+    while (model_iter && feature_collector->metadata->cnt) {
+        VmafModel *model = model_iter->model;
 
+        if (strcmp(model->name, feature_name)) {
```
kylophone, line 365: Do these callbacks only work for model predicted metrics now? What about something like `psnr`?

PR #1408, opened 2025-02-11
Fix gcc14 build errors with CUDA enabled

michalkielan:
I observed a bunch of errors when compiling vmaf with CUDA enabled. Looks like those errors were exposed when using gcc-14. Running meson with gcc-13 works fine with no issues. This PR fixes errors that prevent me from compiling: explicit casting, missing headers, unused parameter.

Repro:
```
$ meson libvmaf/build libvmaf -Denable_cuda=true -Denable_avx512=true
$ ninja -vC libvmaf/build
```

System:
```
$ cat /etc/redhat-release
Fedora release 41 (Forty One)
```
```
$ gcc --version
gcc (GCC) 14.2.1 20250110 (Red Hat 14.2.1-7)
Copyright (C) 2024 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

### DIFF
--- a/libvmaf/src/cuda/picture_cuda.c
+++ b/libvmaf/src/cuda/picture_cuda.c
@@ -22,6 +22,7 @@
 #include "picture_cuda.h"
 #include "common.h"
 #include "log.h"
+#include "ref.h"
 
 #include <cuda.h>
 #include <errno.h>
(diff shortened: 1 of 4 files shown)


nilfm99 (approved):
LGTM, thanks!

PR #1409, opened 2025-02-12
Fix C++ build (svm.cpp).

sethk:


sethk:
Should have mentioned: this is coming up when compiling against MSYS2 with clang, x86_64-pc-windows-msys.

nilfm99 (approved):

PR #1410, opened 2025-02-12
Add ya_getopt, a 2-clause-BSD-licensed implementation of getopt_long()

sethk:
This satisfies the dependency for MSVC. Systems with Glibc, or a suitable getopt_long() available in libc (e.g. macOS/*BSD) will continue using that.


kylophone:
I guess #1412 is a dependency for this one, right?

PR #1411, opened 2025-02-12
Adjust a few macros for MSVC-compatibility

sethk:


sethk:
@nilfm99 Thank you for reviewing my changes. I have a few more that I need to test more broadly before I submit them, but the result is that I've been able to get it compiling under MSVC again.

PR #1418, opened 2025-03-13
In meson.build, change xxd calls for portability (fixes #1417)

### LINKED ISSUE #1417: Cannot build with xxd from busybox.
in src/meson.build
`command : [xxd, '--include', '@INPUT@', '@OUTPUT@']`

If one's xxd is from busybox, then those calls can fail, since busybox often lacks support for long-options:

```
xxd: unrecognized option '--include'
BusyBox v1.35.0 (Debian 1:1.35.0-4+b3) multi-call binary.

Usage: xxd [-pri] [-g N] [-c N] [-n LEN] [-s OFS] [-o OFS] [FILE]

Hex dump FILE (or stdin)

        -g N            Bytes per group
        -c N            Bytes per line
        -p              Show only hex bytes, assumes -c30
        -i              C include file style
        -l LENGTH       Show only first LENGTH bytes
        -s OFFSET       Skip OFFSET bytes
        -o OFFSET       Add OFFSET to displayed offset
        -r              Reverse (with -p, assumes no offsets in input)
```
And those calls result in a exit code of 1.

This could be fixed by using:

`command : [xxd, '-i', '@INPUT@', '@OUTPUT@']`
since the short -i option is supported by all xxd versions.

slycordinator:
Change xxd calls to use '-i' instead of '--include', since the latter is less supported/portable

fixes #1417

### DIFF
--- a/libvmaf/src/meson.build
+++ b/libvmaf/src/meson.build
@@ -151,7 +151,7 @@ if built_in_models_enabled
                     output : model_file,
                     copy: true
                   ),
-                  command : [xxd, '--include', '@INPUT@', '@OUTPUT@'],
+                  command : [xxd, '-i', '@INPUT@', '@OUTPUT@'],
             )
         endforeach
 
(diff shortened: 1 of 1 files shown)


nilfm99 (approved):
thank you!

PR #1419, opened 2025-03-19
Just #define _GNU_SOURCE and _USE_MATH_MACROS unconditionally

sethk:
Remove definitions of M_PI from source.
- `_USE_MATH_MACROS` is required by MSVC when compiling against UCRT.
- `_GNU_SOURCE` is required for other functions that are extensions to

The macros in math.h are not part of POSIX, but are visible by default so long as you don't declare you are trying to be strictly POSIX-conformant by defining `_POSIX_C_SOURCE` or `_XOPEN_SOURCE`.

Tested on:

- macOS 11.7.9 (it's old, sorry)
  Apple clang version 13.0.0
  x86_64-apple-darwin20.6.0

- FreeBSD 14.2-RELEASE
  FreeBSD clang version 18.1.6
  x86_64-unknown-freebsd14.2

- Amazon Linux release 2 (Karoo) 4.14.336-253.554.amzn2.x86_64
  cc (GCC) 7.3.1 20180712 (Red Hat 7.3.1-17)
  x86_64-redhat-linux

- Windows 11 24H2:
  - MSYS2 c++ (GCC) 13.3.0
    x86_64-pc-msys
    msys-2.0.dll runtime

  - CLANG64 clang version 19.1.4
    x86_64-w64-windows-gnu
    ucrtbase.dll runtime

  - MINGW64 cc.exe (Rev2, Built by MSYS2 project)
    14.2.0 x86_64-w64-mingw32
    msvcrt.dll runtime

  - UCRT64 cc.exe (Rev2, Built by MSYS2 project)
    14.2.0 x86_64-w64-mingw32
    ucrtbase.dll runtime

  - Microsoft (R) C/C++ Optimizing Compiler Version 19.42.34433 for x64
    (Requires a few additional PRs that I will submit)

### DIFF
--- a/libvmaf/meson.build
+++ b/libvmaf/meson.build
@@ -19,14 +19,8 @@ cc = meson.get_compiler('c')
 libvmaf_inc = include_directories(['include'])
 
 # Arguments in test_args will be used even on feature tests
-test_args = []
-if host_machine.system() == 'linux' or host_machine.system() == 'windows' or host_machine.system() == 'cygwin'
-    test_args += '-D_GNU_SOURCE'
-    add_project_arguments('-D_GNU_SOURCE', language: ['c', 'cpp'])
-elif host_machine.system() == 'darwin'
-    test_args += '-D_DARWIN_C_SOURCE'
-    add_project_arguments('-D_DARWIN_C_SOURCE', language: ['c', 'cpp'])
-endif
+test_args = ['-D_GNU_SOURCE', '-D_USE_MATH_DEFINES']
+add_project_arguments(test_args, language: ['c', 'cpp'])
 
 # Header checks
 stdatomic_dependency = []
--- a/libvmaf/src/feature/adm_tools.c
+++ b/libvmaf/src/feature/adm_tools.c
@@ -28,10 +28,6 @@
 #include "adm_options.h"
 #include "adm_tools.h"
 
-#ifndef M_PI
-#define M_PI 3.1415926535897932384626433832795028841971693993751
-#endif
-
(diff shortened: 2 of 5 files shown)


sethk:
@alyssais and @kylophone , can you please review? I tried to simplify the defines here to also support MSVC on Windows, FreeBSD, and everything else imaginable without having to redefine M_PI in multiple places.

sethk:
`_USE_MATH_MACROS` exposes `M_PI` when compiling against UCRT on Windows using MSVC or MinGW64 GCC. It has no effect anywhere else.
`_GNU_SOURCE` has no effect when using UCRT on Windows.
`_GNU_SOURCE` has no effect on FreeBSD.
`_GNU_SOURCE` has no effect on macOS.
`_GNU_SOURCE` exposes `M_PI` and `strnlen()` when compiling against glibc (on e.g. GNU/Linux, MSYS2).

`strsep()` is a BSD-ism and seems to be available everywhere except UCRT on Windows. It doesn't require `_BSD_SOURCE` on other systems. There is a local implementation of `strsep()` in `libvmaf/tools/cli_parse.c`, and a test for it in `libvmaf/tools/meson.build`, which may have been added for the sake of UCRT.

Per @alyssais ' last change to `libvmaf/meson.build`, as long as you don't declare strict POSIX or XOPEN conformance with the `_POSIX_C_SOURCE` and `_XOPEN_SOURCE` macros, then `strnlen()` and `strsep()` are available.

alyssais (approved):

PR #1421, opened 2025-03-28
Update Makefile

mariodagostin:


Makefile
```diff
@@ -71,4 +71,4 @@ $(NINJA): $(VENV_PIP)
 	$(VENV_PIP) install ninja || { echo "Failed to install ninja"; exit 1; }
 
 cythonize-deps: $(VENV_PIP)
-	$(VENV_PIP) install setuptools cython numpy || { echo "Failed to install dependencies"; exit 1; }
\ No newline at end of file
+	$(VENV_PIP) install setuptools cython numpy || { echo "Failed to install dependencies"; exit 1; }
```
sethk, line 74: Just a note: this is adding an EOL at the end of the file, which is a good thing and my last change was probably missing it because I'm running Windows.

PR #1422, opened 2025-04-01
MSVC: Miscellaneous uncontroversial fixes

sethk:
- Copy __builtin_clz wrapper into another source file.
- mode_t is not defined on _WIN32
- Avoid arithmetic on void *
- Try to thin out of use of <windows.h>

### DIFF
--- a/libvmaf/tools/vmaf.c
+++ b/libvmaf/tools/vmaf.c
@@ -1,7 +1,18 @@
 #include <stdlib.h>
 #include <string.h>
 #include <time.h>
+#ifdef _WIN32
+#include <corecrt_io.h> // isatty()
+#endif
+#ifdef HAVE_MALLOC_H
+#include <malloc.h> // alloca()
+#endif
+#ifdef HAVE_ALLOCA_H
+#include <alloca.h>
+#endif
+#ifdef HAVE_UNISTD_H
 #include <unistd.h>
+#endif
 
 #include "cli_parse.h"
 #include "spinner.h"
--- a/libvmaf/meson.build
+++ b/libvmaf/meson.build
@@ -29,6 +29,19 @@ elif host_machine.system() == 'darwin'
 endif
 
 # Header checks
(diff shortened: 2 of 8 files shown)


libvmaf/tools/vmaf.c
```diff
@@ -1,7 +1,18 @@
...
+#endif
+#ifdef HAVE_MALLOC_H
+#include <malloc.h> // alloca()
+#endif
+#ifdef HAVE_ALLOCA_H
+#include <alloca.h>
+#endif
+#ifdef HAVE_UNISTD_H
 #include <unistd.h>
+#endif
```
kylophone, line 15: You are guarding all of these headers, but if you don't have any of these the code will fail to compile. Is that right?

sethk: Sorry about the massive delay getting back to this, had to ship a demo.

Different combinations of headers are required for different runtimes. Most systems put many things into unistd.h, and then there's MSVC/UCRT, which doesn't have unistd.h, but instead isatty() and mkdir() are provided by corecrt_io.h and direct.h, respectively. MINGW64 defines _WIN32 but has unistd.h and lacks corecrt_io.h, meaning that we can't just use #ifdef _WIN32 to guard that. These header tests seem like the simplest way to address the full build matrix.

I have another change that checks for malloc.h/alloca.h so it can use alloca() in place of variable-length arrays. I separated it from this one because I think it could be more controversial.

sethk: Forgot to mention: I revised this patch with some comments to clarify why each header is used. It's true that I could have done:

#ifdef HAVE_OTHER_HEADER_H
#include <other_header.h> // foo()
#elif defined(HAVE_UNISTD_H)
#include <unistd.h>
#else
#error "Don't know which header defines foo() on this system"
#endif

But unistd.h also has a ton of other stuff, and the pattern I used is one I've seen in many autoconf-based projects.

PR #1453, opened 2025-12-04
meson: Add top level meson.build

xclaesse:
It is needed to use libvmaf as Meson subproject, like GStreamer does.

Also fix public header includes:
Applications, and thus our public headers, should be included as
`#include <libvmaf/libvmaf.h>`. Adding `include/libvmaf` into include
directories means applications could do e.g. `#include <feature.h>`
which is ambiguous.


xclaesse:
@kylophone what do you think about this PR?

PR #1460, opened 2026-01-02
Add 'enable_tools' option

StormBytePP:
Currently, when you only want the libvmaf library there is no way to disable tool building. This causes problems specially in MSVC environment where those tools do not even compile.

This PR adds an option 'enable_tools' with to allow for tools disabling. As it has a default true value it does not affect current builds.


XuShiSheng:
你好！邮件已收到，我会尽快回复你--许思焱

PR #1472, opened 2026-03-16
cuda: enable CUDA feature extraction on Windows (MSYS2/MinGW)

### LINKED ISSUE #1154: Request: CUDA feature extraction for Windows
Referencing #1152 which added the ability to use Nvidia CUDA acceleration for calculating VMAF and related scores, the existing pipeline does not seem possible to be able to compile a version of libvmaf with CUDA enabled for Windows.

Based on the [Windows build pipeline](https://github.com/Netflix/vmaf/blob/master/.github/workflows/windows.yml) used here, it uses MING64 with MYSYS2, which does not have a way to install the necessary CUDA dependencies (MYSYS2 uses the `pacman` package manager tool, which Nvidia does not have official instructions on installing CUDA for, and the MYSYS2 package repositories do not include CUDA).

Would it also be possible to include builds with CUDA enabled in the existing pipelines? I don't believe there should be any issues in terms of licensing, but someone more knowledgeable than me on that topic can confirm if this is correct.

birkdev:
## Summary

Enables building libvmaf with `-Denable_cuda=true` on Windows using MSYS2/MinGW, closing the gap described in #1154.

On Windows, nvcc requires MSVC's cl.exe as its host compiler for preprocessing `.cu` files, even when the rest of the project is built with MinGW GCC. This creates two categories of issues that this PR addresses:

**Source portability (commit 1):** Headers included by `.cu` files contain constructs that cl.exe cannot process — `<pthread.h>` (POSIX-only), `<ffnvcodec/dynlink_*.h>` (installed in MinGW paths), C99 designated initializers (not supported by nvcc in C++ mode with MSVC). These are resolved with `#ifdef DEVICE_CODE` / `#ifndef __CUDACC__` guards that are no-ops on Linux.

**Build system (commit 2):** The meson build auto-detects cl.exe via `vswhere` (without polluting PATH, which would cause meson to pick MSVC as the default compiler), discovers MSVC and Windows SDK include directories, and passes them to nvcc as `-I` flags since cl.exe runs outside a vcvars environment.

## Prerequisites on Windows

- MSYS2 with MinGW GCC toolchain
- NVIDIA CUDA Toolkit (provides `nvcc`, `bin2c`)
- nv-codec-headers built from git master (commit `876af32` or later). The latest release (`n13.0.19.0`) is missing several `CudaFunctions` members that vmaf uses (`cuMemFreeHost`, `cuStreamCreateWithPriority`, `cuLaunchHostFunc`, etc.). This is a pre-existing issue, not specific to this PR.
- MSVC Build Tools (provides `cl.exe`, needed by nvcc for preprocessing)
- Windows SDK (provides UCRT headers)

## Test results

Tested on Windows 11 with:
- CUDA Toolkit 13.2
- MSVC Build Tools (VS 18), Windows 11 SDK
- MinGW GCC 15.2.0 (MSYS2)
- RTX 5090

The full pipeline works end-to-end:
- `meson setup -Denable_cuda=true` configures successfully
- `ninja` compiles all 7 `.cu` files to fatbin, library links
- FFmpeg `libvmaf_cuda` filter scores at **1,135 fps** on 1080p60 content (37-minute video scored in 2 minutes)

## Transparency note

This patch was developed with assistance from Claude (Anthropic's AI), as reflected in the Co-Authored-By lines. All changes were manually tested on real hardware and real video content.

Closes #1154

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/libvmaf/src/cuda/common.h
+++ b/libvmaf/src/cuda/common.h
@@ -20,7 +20,6 @@
 #ifndef __VMAF_SRC_CUDA_COMMON_H__
 #define __VMAF_SRC_CUDA_COMMON_H__
 
-#include <pthread.h>
 #include <stdbool.h>
 
 #include "config.h"
--- a/libvmaf/src/cuda/cuda_helper.cuh
+++ b/libvmaf/src/cuda/cuda_helper.cuh
@@ -29,7 +29,11 @@
 
 #include "assert.h"
 #include "stdio.h"
+#ifdef DEVICE_CODE
+#include <cuda.h>
+#else
 #include <ffnvcodec/dynlink_loader.h>
+#endif
 
 #define DIV_ROUND_UP(x, y) (((x) + (y)-1) / (y))
 #define MAX(x, y) (((x) > (y)) ? (x) : (y))
--- a/libvmaf/src/feature/cuda/integer_adm/adm_cm.cu
+++ b/libvmaf/src/feature/cuda/integer_adm/adm_cm.cu
@@ -16,7 +16,9 @@
  *     limitations under the License.
  *
  */
+#ifndef DEVICE_CODE
 #include "feature_collector.h"
+#endif
 #include "cuda/integer_adm_cuda.h"
 #include "common.h"
 #include "cuda_helper.cuh"
--- a/libvmaf/src/feature/cuda/integer_adm/adm_csf.cu
+++ b/libvmaf/src/feature/cuda/integer_adm/adm_csf.cu
@@ -17,7 +17,9 @@
  *
  */
 
+#ifndef DEVICE_CODE
 #include "feature_collector.h"
+#endif
 #include "cuda/integer_adm_cuda.h"
 
 #include "common.h"
--- a/libvmaf/src/feature/cuda/integer_adm/adm_csf_den.cu
+++ b/libvmaf/src/feature/cuda/integer_adm/adm_csf_den.cu
@@ -17,7 +17,9 @@
  *
  */
 
+#ifndef DEVICE_CODE
 #include "feature_collector.h"
+#endif
 #include "cuda/integer_adm_cuda.h"
 
 #include "common.h"
--- a/libvmaf/src/feature/cuda/integer_adm/adm_decouple.cu
+++ b/libvmaf/src/feature/cuda/integer_adm/adm_decouple.cu
@@ -17,7 +17,9 @@
  *
  */
 
+#ifndef DEVICE_CODE
 #include "feature_collector.h"
+#endif
 #include "cuda/integer_adm_cuda.h"
 
 #include "common.h"
--- a/libvmaf/src/feature/cuda/integer_adm/adm_dwt2.cu
+++ b/libvmaf/src/feature/cuda/integer_adm/adm_dwt2.cu
@@ -17,7 +17,9 @@
  *
  */
 
+#ifndef DEVICE_CODE
 #include "feature_collector.h"
+#endif
 #include "cuda/integer_adm_cuda.h"
 
 #include "common.h"
--- a/libvmaf/src/feature/integer_adm.h
+++ b/libvmaf/src/feature/integer_adm.h
@@ -132,10 +132,12 @@ struct dwt_model_params {
 };
 
 // 0 -> Y, 1 -> Cb, 2 -> Cr
+#ifndef __CUDACC__
 static const struct dwt_model_params dwt_7_9_YCbCr_threshold[3] = {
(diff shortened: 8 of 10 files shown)


birkdev:
Any thoughts on whether this approach is worth pursuing?

1480c1:
I fail to see how the contents of the PR is linked to the title. I do not see how this would allow someone to use any mingw-w64 based toolchain to compile libvmaf with cuda components without requiring msvc, which would imply you are already going to use msvc with with meson, so this whole setup looks a bit convoluted.

birkdev:
nvcc on Windows can't use GCC as its host compiler. You hit this yourself in #1154 (size_t redeclaration, "cuda sdk's nvcc doesn't understand gcc/g++ on Windows at all"). No PR removes that.

What this enables is libvmaf building in an MSYS2/MinGW environment with CUDA on. .cu files still route through cl.exe via nvcc (unavoidable), host code stays MinGW (matching how FFmpeg is typically cross-built with CUDA on Windows).

"Just use MSVC end to end" isn't a smaller alternative (see #1477, which does exactly that: ~700 net lines, pthread-win32 submodule, four dependency PRs). This is the minimum bridge, #1477 is the full rewrite.

Happy to rename the title.

lusoris:
We have a concrete native Windows/MSVC + CUDA build sample from VMAFx that may help with the native-versus-MinGW discussion here. We carried the CUDA header separation and compiler-discovery work from this PR into our fork, with additional Windows compatibility code.

The [successful Windows MSVC+CUDA job from today](https://github.com/VMAFx/vmafx/actions/runs/34241384400/job/102112347235) built checkout `491d8d2c34e6a3c96c74db408be484a54a0ce26f` (the PR test merge of `2422c608962a7026fdec404def9f77ba827df2e1` into `7bafbb8cc9df61e372b075a934433db20515c23e`). Its transcript contains NVCC fatbin generation, the native `link /MACHINE:x64 /OUT:tools/vmaf.exe` command, and installation of `libvmaf.a`, `vmaf.exe` and `libvmaf_cuda.h`.

The recorded configuration is `windows-2025`, Meson 1.12.0, host `cl` 19.51.36256, Windows SDK 10.0.26100.0, and the CUDA 13.3.1 installer reporting NVCC 13.3.73. One detail matters when reproducing this: the discovery code selected **MSVC 14.29.30133 for NVCC's `-ccbin`**, while Meson used the newer host compiler. The NVCC command also includes `--allow-unsupported-compiler`; this is the actual build transcript, not a claim that every VS/CUDA combination is vendor-supported.

The [complete CI recipe at that checkout](https://github.com/VMAFx/vmafx/blob/491d8d2c34e6a3c96c74db408be484a54a0ce26f/.github/workflows/libvmaf-build-matrix.yml#L819) includes the SDK installation and environment setup. After those prerequisites, its `cmd.exe` build is:

```bat
set INCLUDE=%CD%\nvcodec-prefix\include;%INCLUDE%
set CFLAGS=/experimental:c11atomics
set CXXFLAGS=/experimental:c11atomics
meson setup core core\build --buildtype release ^
  --prefix %CD%\install -Denable_float=true ^
  --default-library=static -Denable_cuda=true ^
  -Denable_nvcc=true -Denable_sycl=false
ninja -v -C core\build install
```

The potentially reusable source pieces are our [native Win32 pthread subset](https://github.com/VMAFx/vmafx/blob/491d8d2c34e6a3c96c74db408be484a54a0ce26f/core/src/compat/win32/pthread.h) and [CUDA custom-target include/dependency wiring](https://github.com/VMAFx/vmafx/blob/491d8d2c34e6a3c96c74db408be484a54a0ce26f/core/src/meson.build#L1304). Our fork now has substantial C/C++ and dependency changes, so this recipe applies to that source checkout; it is not a drop-in build command for unmodified upstream v3.2.0 or independent validation of this PR's MinGW path. Also, that CI recipe clones nv-codec-headers HEAD without recording its SHA, so the transcript does not establish a fully pinned header-version matrix.

The scope of the evidence is **native compilation, linking and installation**. That Windows job has no GPU test step and does not upload its CUDA binaries; the run's `MINGW64-vmaf` artifact comes from a separate job. I cannot use it to confirm native Windows GPU scoring, numerical parity or the RTX 5090 throughput reported above.

One small issue in this PR's current `b7b65e64` build logic: when the `vswhere` search returns no path but `find_program('cl')` succeeds, the fallback fills `nvcc_ccbin_flags` without assigning `cl_path`. The following MSVC include-directory command still concatenates `cl_path`. Assigning `cl_path = cl_exe.full_path()` in that fallback before constructing the flags would make that branch consistent. Our imported copy has the same issue.

birkdev:
Superseded by #1596, which builds the kernels with clang and needs neither nvcc nor MSVC on Windows. Closing this in favor of that approach.

PR #1475, opened 2026-03-25
AVX2: Improve SIMD portability by replacing non‑portable indexing and casts

StormBytePP:
This PR improves the portability of the AVX2 (and AVX-512) implementation by removing two GCC/Clang-specific behaviors that are not supported by other compilers:

1. **Non-portable SIMD vector indexing**  
   Direct indexing of `__m128i` (e.g. `v[0] + v[1]`) is a compiler extension and is not part of the standard Intel intrinsic API.

   It is replaced with a small helper, `mm_hadd_epi64()`:

   - **x86-64:** `_mm_extract_epi64` (SSE4.1) — cheap, no stack spill  
   - **i386:** `_mm_storeu_si128` into a 2-lane array, then sum  

   That covers toolchains on both 32-bit and 64-bit targets without relying on vector subscripting (specially MSVC and possibly others).

2. **Non-standard casts between SIMD types** (`__m256` → `__m256i`)  
   Some compilers allow a direct cast between float and integer vector types; that is not universal.  
   Those casts are replaced with `_mm256_castps_si256`, a zero-cost bit reinterpretation defined by the intrinsic API.

These changes do not alter the algorithm or performance characteristics on x86-64.  
They make the AVX2/AVX-512 backend consistent across toolchains and usable on pure x86 as well.

### DIFF
--- a/libvmaf/src/feature/x86/adm_avx2.c
+++ b/libvmaf/src/feature/x86/adm_avx2.c
@@ -819,12 +819,12 @@ void adm_decouple_avx2(AdmBuffer *buf, int w, int h, int stride,
             __m256 od_inv_64 = _mm256_mul_ps(inv_64, _mm256_cvtepi32_ps(od));
             __m256 rst_d_f = _mm256_mul_ps(kd_inv_32768, od_inv_64);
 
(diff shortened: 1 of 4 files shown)


StormBytePP:
**Update**: Rebased and fixed i686 intrinsics. It is working again.

PR #1476, opened 2026-03-29
Fix UB in void* pointer arithmetic

StormBytePP:
Fix undefined behavior in `void*` pointer arithmetic and a potential memory leak in the failure path of `vif_init()`.

### Changes
- Replaced all `data += N` operations (which are UB on `void*` per C standard) with the portable and well-defined idiom `data = (char *)data + N`.
- In the `fail:` label, we now free the **original** allocation pointer (`s->public.buf.data`) instead of the advanced `data` pointer, preventing a memory leak when `vmaf_feature_name_dict_from_provided_features()` fails.

### Why
- The original code relied on a GNU extension (treating `void*` as `char*` for arithmetic), which is not guaranteed by the C standard and can break under strict conformance or aggressive optimizers.
- The leak was present even before this change; the UB fix simply made it more visible.

The memory layout and runtime behavior remain **identical** — only the correctness and portability are improved.

No functional change, only standard corrections.

### DIFF
--- a/libvmaf/src/feature/integer_vif.c
+++ b/libvmaf/src/feature/integer_vif.c
@@ -630,22 +630,22 @@ static int init(VmafFeatureExtractor *fex, enum VmafPixelFormat pix_fmt,
     if (!data) return -ENOMEM;
     memset(data, 0, data_sz);
 
-    s->public.buf.data = data; data += pad_size;
-    s->public.buf.ref = data; data += frame_size + pad_size + pad_size;
-    s->public.buf.dis = data; data += frame_size + pad_size;
-    s->public.buf.mu1 = data; data += h * s->public.buf.stride_16;
-    s->public.buf.mu2 = data; data += h * s->public.buf.stride_16;
-    s->public.buf.mu1_32 = data; data += s->public.buf.stride_32;
-    s->public.buf.mu2_32 = data; data += s->public.buf.stride_32;
-    s->public.buf.ref_sq = data; data += s->public.buf.stride_32;
-    s->public.buf.dis_sq = data; data += s->public.buf.stride_32;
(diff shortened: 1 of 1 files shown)


kylophone:
Maybe just change the original type instead of casting N times?

StormBytePP:
> Maybe just change the original type instead of casting N times?

I wanted to maintain the original format so the changes introduced are minimal.

Edit: There are some fields which still require explicit casts if original type is changed (at least in mingw) like:
```
../src/feature/integer_vif.c:636:23: error: assignment to 'uint16_t *' {aka 'short unsigned int *'} from incompatible pointer type 'char *' [-Wincompatible-pointer-types]
  636 |     s->public.buf.mu1 = data; data += h * s->public.buf.stride_16;
      |                       ^
../src/feature/integer_vif.c:637:23: error: assignment to 'uint16_t *' {aka 'short unsigned int *'} from incompatible pointer type 'char *' [-Wincompatible-pointer-types]
  637 |     s->public.buf.mu2 = data; data += h * s->public.buf.stride_16;

```

So I reverted because I think it is more readable to have all casted uniformly instead of sometimes and sometimes not. What do you think?

StormBytePP:
Rebased onto current `master`.

The fail-path leak this PR originally fixed is already handled upstream: `vif_buffer_alloc()` keeps the `aligned_malloc()` pointer in `buf->data`, and both `init`'s `fail` path and `close()` free it via `vif_buffer_free()`.

What remains is the pointer bump in `vif_buffer_alloc()`. `data += n` on a `void *` is undefined in ISO C (GNU treats it as `char *`; MSVC and other non-GNU compilers do not). Using `data = (char *)data + n` makes that placement arithmetic well-defined everywhere. `data` stays `void *` so field assignments do not need extra casts (MinGW `-Wincompatible-pointer-types`).

PR #1487, opened 2026-04-20
Fix null deref crash on prev_ref update in pure CUDA pipelines

lawrencecurtis:
#1482 introduced a segfault to libvmaf_cuda when using ffmpeg.

Pure CUDA pipelines leave `ref_host` zero-initialized. This adds a null check before `vmaf_picture_ref` to prevent crashes. 

Skipping this update seems safe because pure CUDA extractors don't appear to consume it

### DIFF
--- a/libvmaf/src/libvmaf.c
+++ b/libvmaf/src/libvmaf.c
@@ -1013,7 +1013,9 @@ int vmaf_read_pictures(VmafContext *vmaf, VmafPicture *ref, VmafPicture *dist,
 
     if (vmaf->prev_ref.ref)
         vmaf_picture_unref(&vmaf->prev_ref);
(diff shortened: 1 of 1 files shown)


kylophone:
Thanks. I tried to keep the threading changes behind the experimental flags, does this crash affect the build without any of the new threading modes enabled?

lawrencecurtis:
Yes the crash happens both with and without sadly

lawrencecurtis:
Great work on the new threading changes though. It takes 5 threads on an i5 from 22fps to 33fps. Huge gains

lawrencecurtis:
Sorry for all the noise, merged the wrong branch, then rebased. How embarrassing lol

kylophone:
Merged, thanks!

PR #1495, opened 2026-04-27
x86: move avx2+512 opts to 64-bit only

### LINKED ISSUE #1481: `libvmaf` failure to compile on 32-bit x86 GNU/Linux
Unfortunately, this PR has caused `libvmaf` to begin failing to compile with all build settings at defaults on 32-bit x86 GNU/Linux.

- https://github.com/Netflix/vmaf/pull/1452

```
FAILED: src/libx86_avx2.a.p/feature_x86_adm_avx2.c.o 
cc -Isrc/libx86_avx2.a.p -Isrc -I../src -I../src/feature/common -Iinclude -I../include -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=c11 -O3 -D_GNU_SOURCE -fPIC -mavx -mavx2 -pedantic -DOC_NEW_STYLE_INCLUDES -MD -MQ src/libx86_avx2.a.p/feature_x86_adm_avx2.c.o -MF src/libx86_avx2.a.p/feature_x86_adm_avx2.c.o.d -o src/libx86_avx2.a.p/feature_x86_adm_avx2.c.o -c ../src/feature/x86/adm_avx2.c
../src/feature/x86/adm_avx2.c: In function ‘adm_decouple_s123_avx2’:
../src/feature/x86/adm_avx2.c:1439:24: error: implicit declaration of function ‘_mm256_extract_epi64’; did you mean ‘_mm256_extract_epi8’? [-Wimplicit-function-declaration]
 1439 |             calc_angle(_mm256_extract_epi64(ot_dp_lo_epi64, 0), _mm256_extract_epi64(o_mag_sq_lo_epi64, 0), _mm256_extract_epi64(t_mag_sq_lo_epi64, 0), angle_flag[0]);
      |                        ^~~~~~~~~~~~~~~~~~~~
```

Steps to reproduce:

1. Install GNU/Linux on a 32-bit x86 computer
2. Install development packages like `git`, `build-essential` and `nasm` (possibly others also)
3. Use these commands:

```bash
git clone https://github.com/Netflix/vmaf.git
cd vmaf/libvmaf
meson setup build
meson compile -C build
```

Operating system: 32-bit x86 Debian 13 GNU/Linux

Viable workarounds I have tested:

- Workaround 1: use the last commit before that PR which worked, https://github.com/Netflix/vmaf/commit/7e16db0a2ccdd8547680b9ed0b3e52691e8ecee7 That commit compiles without errors and without changing the default build settings on 32-bit x86 GNU/Linux
- Workaround 2: use this command to configure `libvmaf`: `meson setup build -Denable_asm=false`. This disables the problematic code, allowing the newest version of `libvmaf` to be used on 32-bit x86 GNU/Linux without downgrading it, albeit I assume with a potential performance penalty.

1480c1:
This matches the compiling and linking of avx2 and avx512 code to match the checks in cpu.c, since either way, avx2 and above is disabled for 32-bit.

Closes https://github.com/Netflix/vmaf/issues/1481

### DIFF
--- a/.github/workflows/libvmaf.yml
+++ b/.github/workflows/libvmaf.yml
@@ -28,10 +28,15 @@ jobs:
           - os: ubuntu-24.04-arm
             CC: ccache clang
             CXX: ccache clang++
+          - os: ubuntu-latest
+            CC: ccache gcc
+            CXX: ccache g++
+            CROSS: ./libvmaf/test/meson-toolchains/i686-pc-linux-gnu-gcc-cross.ini
     runs-on: ${{ matrix.os }}
     env:
       CC: ${{ matrix.CC }}
       CXX: ${{ matrix.CXX }}
+      CROSS: ${{ matrix.CROSS }}
     steps:
       - name: Setup python
         uses: actions/setup-python@v6
@@ -55,6 +60,9 @@ jobs:
             ;;
           *clang) sudo -E apt-get -yq install clang nasm ;;
           esac
+          if [ -n "$CROSS" ]; then
+            sudo -E apt-get -yq install gcc-multilib g++-multilib
+          fi
           $CC --version
           meson --version
           ccache --version
@@ -71,7 +79,7 @@ jobs:
       - uses: actions/checkout@v6
       - name: Run meson
         run: |
-          meson setup libvmaf libvmaf/build --buildtype release --prefix $PWD/install -Denable_float=true
+          meson setup libvmaf libvmaf/build --buildtype release --prefix $PWD/install -Denable_float=true ${CROSS:+--cross-file $CROSS}
       - name: Run ninja
         run: |
           sudo ninja -vC libvmaf/build install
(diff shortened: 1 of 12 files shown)


kylophone:
I wonder if this can be more targeted? I think that most of these blocks build and run fine on 32-bit.

kylophone:
Maybe a 32-bit cross-compiling CI job would help?

1480c1:
> I wonder if this can be more targeted? I think that most of these blocks build and run fine on 32-bit.

It is possible, I did try a route of instead making a compat inline func to address the actual error issue and remove the limitation in cpu.c to allow using avx2 and above for 32 bit. I did confirm that that it didn't crash and produced a vmaf score equal to 64-bit vmaf, but since the intention of cpu.c copied from dav1d explicitly disabled them, I was wondered if my changes were correct and instead went with this route.

I can try again and show what the other route would look like and also add a cross compile to 32 bit job in a bit to test and show

PR #1505, opened 2026-04-29
x86/adm: fix compilation for 32-bit clang

1480c1:
Addresses the issue reported in https://github.com/Netflix/vmaf/issues/1481#issuecomment-4344041294

### DIFF
--- a/.github/workflows/libvmaf.yml
+++ b/.github/workflows/libvmaf.yml
@@ -32,6 +32,10 @@ jobs:
             CC: ccache gcc
             CXX: ccache g++
             CROSS: ./libvmaf/test/meson-toolchains/i686-pc-linux-gnu-gcc-cross.ini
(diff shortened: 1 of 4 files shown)


1480c1:
@kylophone I can confirm that the test errors for 32-bit clang still occur even with avx2 and 512 disabled, from cpu.c.

I can either go the easy route of allowing the 32-bit test for clang to fail or try to debug it, but do you think it is worth the time and effort right now? I've been seeing quite a few PRs from you regarding features being changed. Should I wait until your changes are settled?

1480c1:
It seems changing the mfpmath flag to sse to match what 64-bit gcc does by default is enough. Although curiously, 32-bit gcc is supposed to default to 387 as well, but it's not throwing an error.

kylophone:
looks good, thanks for your help here!

PR #1536, opened 2026-05-17
port(upstream): switch VIF from precomputed to on-the-fly filter (bf9ad333)

lusoris:
Ports Netflix upstream commit bf9ad333 which switches the float VIF path from precomputed Gaussian filter lookup to on-the-fly computation. Likely resolves a chunk of the 176 golden test failures (test goldens were copied from upstream with the new values).


lusoris:
Wrong target — meant for lusoris/vmaf fork. Closing.

Issue #1547, opened 2026-06-24
`Speed_chroma_feature_speed_chroma_uv_score` error with `libvmaf` and `ffmpeg` from homebrew, with v1 models


vladshcherbin:
Hi 👋 

I've installed `ffmpeg v8.1.2` via Homebrew, which pulled in `libvmaf 3.2.0` as a dependency. After installation, I tried to collect VMAF, with default model:

```sh
ffmpeg -i v-12-265-26-m.mp4 -i v-12-30.mp4 -lavfi libvmaf=log_path=output.xml -f null -
# or
ffmpeg -i v-12-265-26-m.mp4 -i v-12-30.mp4 -lavfi libvmaf=log_path=output.xml:model=version=vmaf_v0.6.1 -f null -
```

Both of them work and correctly show `VMAF score`.

Next I've tried with the newer v1.0.16 model (`vmaf_v1.0.16_3d0h`):

```sh
ffmpeg -i v-12-265-26-m.mp4 -i v-12-30.mp4 -lavfi libvmaf=log_path=output.xml:model=version=vmaf_v1.0.16_3d0h -f null -
```

This fails with:

libvmaf ERROR could not initialize feature extractor "Speed_chroma_feature_speed_chroma_uv_score"

<img width="976" height="338" alt="Image" src="https://github.com/user-attachments/assets/2814506e-509e-4790-a279-d49e13886113" />

I'm not sure where the issue can be fixed, here or in homebrew directly.

Can you please suggest the next steps?

kylophone:
> which pulled in libvmaf 3.2.0 as a dependency

Can you double check the version (`vmaf --version`)? Your error message makes it seem like the speed feature extractor was not compiled into your libvmaf. As of 4718b4f5fe58484d8b073249b81eb0cc3a3a0471 this would be part of the default build.

vladshcherbin:
@kylophone yes, I believe it's 3.2.0:

<img width="950" height="362" alt="Image" src="https://github.com/user-attachments/assets/5cf76227-5e87-4b83-8825-db5b108a069c" />

I can also see the v1.0.16 models json files, inside directory with 3.2.0 name:

https://github.com/user-attachments/assets/f8364e45-2a75-42ca-a29e-1fa344925d34

Please let me know if you need any information

ChristianFeldmann:
You must set `-Denable_float=true` in meson when compiling libvmaf. In your case (you just installed from homebrew), this is missing and the extractor will not be available.

[...]

PR #1548, opened 2026-06-27
Enable floating-point feature extractors by default

lawrencecurtis:
CAMBI and SpEED from the v1 model require float to function, so this should probably be on by default


lawrencecurtis:
This should also fix #1547

kylophone:
Thanks for the contribution. Rather than enable all float feature extractors, I will try to be a little more targeted with #1558. Closing this one for now.

PR #1477, opened 2026-03-29
Add native MSVC build support for Windows

StormBytePP:
# Add native MSVC support

This pull request introduces first-class, native MSVC support for `libvmaf` on Windows.  
The goal is to enable building, testing, and integrating `libvmaf` using the Microsoft Visual C++ toolchain without relying on MinGW/MSYS2, greatly improving compatibility with Windows-native projects and CI environments.

## Motivation

Until now, `libvmaf` could only be built on Windows through MinGW/MSYS2. While functional, this approach had several limitations:

- Native Windows projects (CMake, Visual Studio, FFmpeg builds, etc.) could not easily consume MSYS2-compiled libraries.
- MSVC is the standard toolchain in most Windows CI/CD pipelines.
- Debugging, profiling, and static analysis tools on Windows work best with MSVC binaries.
- Distributing official Windows binaries is significantly easier when built with MSVC.

This PR adds full native MSVC support, including:
- Meson build configuration compatible with MSVC
- Working static library build
- Working test suite execution
- Fully functional GitHub Actions CI for MSVC builds (both standalone library and with FFmpeg)

## Dependencies / Related PRs

This PR depends on the following PRs being merged **first**.  

Merging them first is especially important to ensure proper credit and authorship for all contributors — particularly for the two PRs that were **not authored by me** (#1410 and #1428).  

Once those PRs are merged upstream, the temporary copies of their commits (included here only for CI to pass) will be removed. This way the original authors receive full and fair attribution in the git history.

The required PRs are:

- [ ] #1475
- [ ] #1410
- [ ] #1476
- [ ] #1428

This ensures:
- CI runs correctly
- MSVC builds succeed
- The test suite passes
- No regressions are introduced

## What’s included

- Added [`pthread-win32`](https://github.com/GerHobbelt/pthread-win32) as a git submodule for MSVC builds
- Native MSVC build support via Meson
- Updated Meson configuration for full MSVC compatibility
- New Windows CI workflow using a matrix for MSYS2 + MSVC
- Automatic installation of required tools in CI (`meson`, `ninja`, `xxd`, `cmake`)
- Full test suite passing on Windows (standalone library and with FFmpeg)
- Clean separation between MSYS2 and MSVC environments
- No changes to existing Linux/macOS workflows
- Updated `Windows.md` documentation to include MSVC instructions

## CI Improvements

The new GitHub Actions workflow builds both MSYS2 and MSVC in parallel, runs the full test suite under each environment, and uploads artifacts for both toolchains.

[...]

PR #1551, opened 2026-07-02
Fix problems when compiling for/on various platforms and OSes

JensAc:
Hey there, 

I just collected a few changes required when you want to compile for/on various platforms and OSes. 
E.g. MSVC would not compile the avx accelerated code with plain casts and `[]` access in ` __m128i`s. 

The changes do not change the behavior of the code itself, but make life simpler when (cross) compiling libvmaf.

### DIFF
--- a/libvmaf/src/feature/barten_csf_tools.h
+++ b/libvmaf/src/feature/barten_csf_tools.h
@@ -20,6 +20,10 @@
 #include "common/macros.h"
 #include <errno.h>
 
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+
 #pragma once
 
 #ifndef BARTEN_CSF_TOOLS_H_
--- a/libvmaf/src/feature/ciede.c
+++ b/libvmaf/src/feature/ciede.c
@@ -52,6 +52,10 @@ SOFTWARE.
 #include "mem.h"
 #include "opt.h"
 
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+
 typedef struct CiedeState {
     VmafPicture ref;
     VmafPicture dist;
--- a/libvmaf/src/feature/integer_adm.h
+++ b/libvmaf/src/feature/integer_adm.h
@@ -171,9 +171,9 @@ struct dwt_model_params {
 
 // 0 -> Y, 1 -> Cb, 2 -> Cr
 static const struct dwt_model_params dwt_7_9_YCbCr_threshold[3] = {
-    {.a = 0.495, .k = 0.466, .f0 = 0.401, .g = {1.501, 1.0, 0.534, 1.0}},
-    {.a = 1.633, .k = 0.353, .f0 = 0.209, .g = {1.520, 1.0, 0.502, 1.0}},
-    {.a = 0.944, .k = 0.521, .f0 = 0.404, .g = {1.868, 1.0, 0.516, 1.0}}};
+    {0.495, 0.466, 0.401, {1.501, 1.0, 0.534, 1.0}},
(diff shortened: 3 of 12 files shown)


JensAc:
Ahh I have not seen that this is more or less a duplicate of #1477 which is even more complete than this version.  I am happy to close this once #1477 is merged :slightly_smiling_face: .

kylophone:
Hey, @JensAc thanks for the patience as I've been working through all these PRs. Can you tell me, is it mainly MSVC support you are looking for? So far we've only officially supported MINGW64 builds on Windows, but I'm trying understand how much value MSVC brings for Windows devs?

JensAc:
Hi @kylophone, 
thanks for the response. I'll try to tell the story behind as that might help to understand the use case. 
We integrated libvmaf into our encoders to measure video quality while encoding. We build for Windows (x64 and arm), OS X, and Linux (x64 and arm) . For that reason, we compile libvmaf with the appropriate toolchain for the target OS and CPU architecture. 

In particular, with MSVC we ran into issues, and I think these are addressed completely by #1477 , that I did not recognize when opening this PR. I did not want to propose too many changes (e.g. introducing pthreads4win), as I thought the interest might be limited due to already working MINGW64 builds. For that reason, I concentrated on changes that do not alter the functional behavior, but make life easier in case you want to compile with MSVC. From my perspective, you could close this PR, if you don't want to apply the changes or prefer #1477. The other option would be to merge it, as it should not affect the functional behavior.

StormBytePP:
Hi @JensAc, I am the original author of the PR you referenced.

First of all, thanks for the mention and for looking at my PR.

As you can see there, the PR is not monolithic and depends on other little PRs which address more atomic items needed first (like avx).

You are right onto mentioning pthreads4win, it can also be implemented in a way that it can be used externally while I bundled it for the new CI tests to pass and to help non experienced users.

Moreover, vmaf's codebase changed a bit and it needs to be rebased to work with the latest changes. Though the changes are minimal I did not wanted to touch more the PR in order to wait for @kylophone to have time to look at it as said so he does not feel pressured, a thing that of course I will do to make the PR to work again.

However, I would recommend to close this PR (and I will take the little changes for those atomic PRs too crediting you of course) because I also include doc and CI tests to ensure that MSVC build will never break with the future changes.

Cheers,

David.

JensAc:
Hi David, @StormBytePP , 

I fully agree with your comment and your recommendation is totally fair. Also, I like the effort you put into MSVC support very much. However, I will leave it up to @kylophone to decide how to proceed and not close this for the moment.

Best regards
Jens

StormBytePP:
Hi Jens, @JensAc,

Yes, I am also waiting for his review/instructions. I have just rebased and updated all the code to work with the latest master so the review can be done properly.

Specially I've fixed the problems you had here too in the CI tests (the same ones that appeared with the latest master) by changing the intrinsics to properly support plain x86 code:

```
/usr/bin/ld: adm_avx2.c:(.text+0xe5a3): undefined reference to `_mm_extract_epi64'
/usr/bin/ld: adm_avx2.c:(.text+0xe5e5): undefined reference to `_mm_extract_epi64'
/usr/bin/ld: adm_avx2.c:(.text+0xe5fb): undefined reference to `_mm_extract_epi64'
/usr/bin/ld: adm_avx2.c:(.text+0xe63d): undefined reference to `_mm_extract_epi64'
```

Additionally, I have added a new option `-Dbundled_winpthreads` (default: `true`). This allows using either the bundled pthread-win32 implementation or an external Windows pthreads library. You can find more details in the updated Windows documentation.

If you have the chance, it would be greatly appreciated if you could test the external pthreads mode (`-Dbundled_winpthreads=false`) to see whether it fits the needs of your projects. Any feedback would be very valuable.

Best regards,
David

PR #1560, opened 2026-07-10
Enable fetching scores at runtime again

JensAc:
Hey again :slightly_smiling_face: , 

In v3.2.0 fetching scores at runtime seems to be broken due to the change in flushing `integer_motion` features. It appears to me that `motionx` features are not computed in `extract()` anymore but in `flush` once all pictures are read. However, [here](https://github.com/Netflix/vmaf/blob/master/libvmaf/README.md#api-walkthrough) it is still stated that  `vmaf_score_at_index(..)` could be called to fetch a score once pictures are read. 

I suppose the current behavior was not intended and this PR provides a "hotfix". I know, it might not be the most elegant solution to poll for the `sad` feature to be available in a loop, but at least this works in local verification tests on my machine. 

If there is anything I could/should do to verify the behavior or on the implementation side, please let me know. I will do my best to make this ready for the patch release you announced [here](https://github.com/Netflix/vmaf/issues/1547#issuecomment-4928569754).

### DIFF
--- a/libvmaf/src/feature/integer_motion.c
+++ b/libvmaf/src/feature/integer_motion.c
@@ -55,6 +55,8 @@ typedef struct MotionState {
     double motion_blend_factor;
     double motion_blend_offset;
     double motion_fps_weight;
+    double prev_processed;
+    double stamp_value;
     bool motion_five_frame_window;
     bool motion_moving_average;
     bool motion_force_zero;
@@ -259,6 +261,9 @@ static int init(VmafFeatureExtractor *fex, enum VmafPixelFormat pix_fmt,
     s->h = h;
     s->bpc = bpc;
 
+    s->prev_processed = 0.0;
+    s->stamp_value = 0.0;
+
     s->feature_name_dict =
         vmaf_feature_name_dict_from_provided_features(fex->provided_features,
                 fex->options, s);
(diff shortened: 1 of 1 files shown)


kylophone:
One of the main motivators for the motion refactor was it improved the multithreading performance significantly. See the graph data in the original PR [here](https://github.com/Netflix/vmaf/pull/1483#issue-4221240982). I didn't test it yet, but I think your changes probably re-introduce a cross-frame dependency that limits the threading throughput.

I agree that this is a problem we need to fix, however I think we need to make sure we are not hurting the multithreading performance. I have some ideas on how to do this, but I am curious to know what you think?

JensAc:
Yes, I agree. These changes introduce a cross-frame dependency and block threads for some time in case the `sad` feature is not available (see [here](https://github.com/JensAc/vmaf/blob/60034a649aa6d7a17f8a5c5e658d6aa3cd30519c/libvmaf/src/feature/integer_motion.c#L319)). This is why I called it a "hotfix". Performance wise, however, this is not very different from computing the `motionx` features when flushing, as they are calculated sequentially either, aren't they?

When I measure the performance impact of these changes, the runtime difference between `v3.2.0` and this branch for a 1080p sequence of 2100 frames is minor. I measure on a AMD Ryzen 9 9950X 16-Core Processor with the command
```
 time ./libvmaf/build/tools/vmaf -r ... -d .. -w 1920 -h 1080 -p 420 -b 8 -m version=vmaf_v0.6.1neg --json -o vmaf.json --threads 32 
```
and get 
```
this branch:  62.14s user 8.66s system 891% cpu 7.938 total --> 264.6 fps
v3.2.0:       59.70s user 8.48s system 867% cpu 7.861 total --> 267.1 fps
```

Don't get me wrong, if you have ideas to solve this is a better way, let's do that. I am happy to help, but would need some more insight into your ideas. I just read your comment about an upcoming patch release and wanted to make sure that this is on the radar.

JensAc:
Hey @kylophone , 
I have observed that you just release v.3.2.1. I did not test it yet, but from the change log, it seems that this issue still remains or am I missing something?
Happy to close this one in case it is solved. Also happy to iterate, if required :slightly_smiling_face: .

PR #1575, opened 2026-08-11
Add support for Intel CET (SHSTK)

### LINKED ISSUE #1455: Libvmaf: missing support for SHSTK (Intel CET)
## Problem description

SHSTK is an x64 security feature that is part of Intel CET. In order to support SHSTK, each object of a library must have a `.note.gnu.property` section. **The problem is**, libvmaf's assembly files do not emit this section. As a result, `libvmaf.so` is not marked as being SHSTK-compatible, and this causes SHSTK not to be supported on applications that use libvmaf.

## How to reproduce

The library is not marked as being SHSTK-compatible:

```bash
user@fedora:~$ readelf -n /usr/lib64/libvmaf.so.3 | grep SHSTK
... (no results) ...
```

## Proposed fix

Libvmaf builds a single x64 assembly file, `libvmaf/src/x86/cpuid.asm`, which includes `libvmaf/src/ext/x86/x86inc.asm` for generic ASM definitions. `cpuid.asm` does not have mismatched calls/rets, and therefore can support SHSTK as-is, it only needs to emit the `.note.gnu.property` section to advertise that it is SHSTK-compatible. `x86inc.asm` comes from [VideoLan](https://code.videolan.org/videolan/x86inc.asm/-/blob/master/x86inc.asm?ref_type=heads), which in fact added support for emitting that section in [this commit from 2024](https://code.videolan.org/videolan/x86inc.asm/-/commit/c7cf926c08661e754f411722a035c3fefde313ab).

Therefore, fixing the issue is only a matter of syncing `x86inc.asm` with the upstream VideoLan version. Or in a reduced form, cherry-picking the aforementioned commit. Having done the latter locally, I can confirm that it solves the problem:

```bash
user@fedora:~$ readelf -n libvmaf.so.3 | grep SHSTK
      Properties: x86 feature: SHSTK
```

StormBytePP:
This PR adds Intel CET SHSTK support to the assembly code.

Ported from VideoLAN commit `c7cf926c08661e754f411722a035c3fefde313ab`.

Fixes #1455 (thanks to m00nbsd)

### DIFF
--- a/libvmaf/src/ext/x86/x86inc.asm
+++ b/libvmaf/src/ext/x86/x86inc.asm
@@ -815,9 +815,26 @@ BRANCH_INSTR jz, je, jnz, jne, jl, jle, jnl, jnle, jg, jge, jng, jnge, ja, jae, 
     %1: %2
 %endmacro
 
-; This is needed for ELF, otherwise the GNU linker assumes the stack is executable by default.
 %if FORMAT_ELF
+    ; The GNU linker assumes the stack is executable by default.
     [SECTION .note.GNU-stack noalloc noexec nowrite progbits]
+
+    %ifdef __NASM_VER__
+        %if __NASM_VER__ >= 0x020e0300 ; 2.14.03
+            %if ARCH_X86_64
+                ; Control-flow Enforcement Technology (CET) properties.
+                [SECTION .note.gnu.property alloc noexec nowrite note align=gprsize]
+                dd 0x00000004  ; n_namesz
+                dd gprsize + 8 ; n_descsz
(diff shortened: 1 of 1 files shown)


kylophone:
Just curious, do you have any intention to use this in any way?

StormBytePP:
> Just curious, do you have any intention to use this in any way?

It's mainly for completeness. I run Gentoo with a hardened profile and compile everything with Intel CET support for better security. Having this library compatible isn't critical for me, but it's nice to have it fully supported since it helps close off potential attack vectors.

kylophone:
OK, I think it's fine to merge anyway.

PR #1577, opened 2026-08-11
AVX512: Drop -mavx512vbmi from generic AVX-512 compile flags

### LINKED ISSUE #1555: x86 AVX512 build enables VBMI, but runtime dispatch appears to gate only on AVX512 (can cause illegal instruction on non-VBMI AVX512 CPUs)
### Summary
On x86_64, VMAF appears to compile AVX512 feature objects with VBMI-enabled compiler flags, while runtime dispatch in integer feature initialization appears to use AVX512-level gating.  This can lead to illegal instruction faults on CPUs that support base AVX512 but do not support VBMI.
 
### Impact?
Some production CPUs support AVX512F/BW/DQ/VL/CD but not AVX512VBMI.  
If code is compiled with VBMI enabled and selected at runtime via AVX512-only checks, execution hits SIGILL / illegal instruction.
 
### What I observed
1. Build path includes AVX512 feature compilation with VBMI-enabled flags.
2. Runtime selection for AVX512 code paths appears tied to AVX512 capability checks.
3. On AVX512-without-VBMI hardware, this can fail at runtime in AVX512 feature execution.
 
### Expected behavior
Either:
1. AVX512 code compiled for generic AVX512 should avoid VBMI-only instructions unless explicitly guarded, or
2. Runtime dispatch should require VBMI-capable CPU flags before selecting objects/functions compiled with VBMI.
 
### Actual behavior
AVX512 path can be selected on CPUs lacking VBMI, and those hosts may fail with illegal instruction.
 
### Repro guidance
1. Build VMAF with current x86 AVX512 path enabled.
2. Run on x86_64 CPU with AVX512 base support but no AVX512VBMI.
3. Exercise VMAF integer feature path with AVX512 dispatch active.
4. Observe illegal instruction in AVX512 path.
 
### Suggested fix options
1. Remove VBMI from default AVX512 compile flags for generic AVX512 objects, or
2. Split VBMI-dependent code into separately gated objects/functions, and dispatch them only when VBMI is present, or
3. Strengthen runtime gating so AVX512VBMI-capable CPUs are required for any VBMI-compiled AVX512 path.
 
### Temporary workaround used downstream
Disable AVX512 path selection for affected x86_64 targets to avoid runtime fault on non-VBMI AVX512 CPUs.
 
Thanks for taking a look.

StormBytePP:
## Summary

Closes #1555.

The issue is correct: the generic AVX-512 objects were compiled with `-mavx512vbmi` while runtime dispatch can select that path based only on base AVX-512 support (`VMAF_X86_CPU_FLAG_AVX512`).

AVX512_VBMI is a separate extension. CPUs with AVX-512F/BW/DQ/VL/CD but **without** VBMI are common, for example:

- Intel Skylake-X / Skylake-SP
- Intel Cascade Lake (no VBMI)
- Other server/workstation parts with “base” AVX-512 only

On those CPUs, a VBMI instruction emitted under `-mavx512vbmi` can fault with an illegal instruction (`SIGILL`).

## What I checked

A search of the AVX-512 C sources shows no explicit VBMI intrinsics (e.g. `*_permutexvar_epi8`, `vpmultishift`, `vpermb`). The only reference to VBMI was the compiler flag in `libvmaf/src/meson.build`.

So `-mavx512vbmi` was unnecessary for the current code and only widened the ISA requirement beyond what runtime gating guarantees.

## Change

Remove `-mavx512vbmi` (and a duplicated `-mavx512dq`) from the generic AVX-512 `c_args`, leaving:

- `-mavx512f`
- `-mavx512dq`
- `-mavx512bw`
- `-mavx512cd`
- `-mavx512vl`

This matches the base AVX-512 feature set expected by `VMAF_X86_CPU_FLAG_AVX512`.

If a future path truly needs VBMI, it should be compiled separately and gated with a stricter runtime flag (e.g. the existing `VMAF_X86_CPU_FLAG_AVX512ICL` / explicit VBMI), not folded into the generic AVX-512 objects.

### DIFF
--- a/libvmaf/src/meson.build
+++ b/libvmaf/src/meson.build
@@ -316,8 +316,7 @@ if is_asm_enabled
             'x86_avx512',
             x86_avx512_sources,
             include_directories : vmaf_base_include,
-            c_args : ['-mavx512f', '-mavx512dq', '-mavx512bw', '-mavx512cd', '-mavx512dq',
-                      '-mavx512vbmi', '-mavx512vl'] +
+            c_args : ['-mavx512f', '-mavx512dq', '-mavx512bw', '-mavx512cd', '-mavx512vl'] +
                      vmaf_cflags_common,
         )
 


kylophone:
It would be nice to know which compilers added VBMI instructions, and whether or not this affects performance on those systems. If there are perf wins, we should find a way to keep them. Rather than just checking for the intrinsics in the code, we should maybe check the compiled output. Or at least do a performance check with and without this flag. This is what I was getting at here: https://github.com/Netflix/vmaf/issues/1555#issuecomment-4928362335

StormBytePP:
Thanks for the feedback — I checked the **compiled output**, not only the source intrinsics.

### Method

Four release builds from **this PR’s tree** (so `meson.build` does not inject `-mavx512vbmi`), with `-Denable_avx512=true`. The same AVX-512 baseline flags were used; only VBMI and the host compiler differed. VBMI was enabled solely via `-Dc_args` in the “with VBMI” configures.

**Compilers tested**

- **GCC** 15.3.1 (Gentoo Hardened 15.3.1_p20260717)
- **Clang** 21.1.8

**Flags**

- **without** `-mavx512vbmi`: `-mavx512f -mavx512dq -mavx512bw -mavx512cd -mavx512vl`
- **with** `-mavx512vbmi`: the same set plus `-mavx512vbmi`

**Configure (example layout)**

```bash
# GCC, without VBMI
CC=gcc CXX=g++ meson setup /tmp/vmaf-no-vbmi-gcc libvmaf \
  --buildtype release --default-library static \
  -Denable_float=true -Denable_avx512=true \
  -Dc_args="-mavx512f -mavx512dq -mavx512bw -mavx512cd -mavx512vl"

# GCC, with VBMI
CC=gcc CXX=g++ meson setup /tmp/vmaf-vbmi-gcc libvmaf \
  --buildtype release --default-library static \
  -Denable_float=true -Denable_avx512=true \
  -Dc_args="-mavx512f -mavx512dq -mavx512bw -mavx512cd -mavx512vl -mavx512vbmi"

# Clang, without / with VBMI (same as above with CC=clang CXX=clang++)
```

Then `ninja -C …`, opcode counts on the `x86_avx512` objects:

```bash
objdump -d …/feature_x86_${tu}_avx512.c.o | grep -cE 'vpermt2b|vpermb|vpmultishift'
```

and a full asm diff for ADM under Clang:

```bash
objdump -d --no-show-raw-insn …/feature_x86_adm_avx512.c.o > adm_{no_,}vbmi.S
diff -u adm_no_vbmi.S adm_vbmi.S
```

### Results

| TU | GCC + VBMI | GCC − VBMI | Clang + VBMI | Clang − VBMI |
|----|------------|------------|--------------|--------------|
| `adm_avx512` | 0 | 0 | **2** (`vpermt2b`) | 0 |
| `vif_avx512` | 0 | 0 | 0 | 0 |
| `motion_avx512` | 0 | 0 | 0 | 0 |
| `speed_avx512` | 0 | 0 | 0 | 0 |

So:

- **GCC 15.3.1** did not emit any of those VBMI opcodes with or without the flag (for this code).
- **Clang 21.1.8** emitted VBMI only with the flag, only in **ADM**, and only **two** `vpermt2b` in the whole AVX-512 backend.

**On the asm diff:** I also compared `objdump -d` listings for `adm_avx512` with vs without the flag (including a pass that strips addresses so `diff` is less dominated by offset noise). The meaningful change is local: in two places Clang turns a short `vpshufb` / `vpshufb` / `vpackusdw` sequence into a single `vpermt2b`. A raw unified diff still looks large because addresses, jump targets, and register allocation shift around those sites; that does not mean the rest of ADM was rewritten. VBMI opcode count remains **2 vs 0**, and the other AVX-512 TUs have **zero** VBMI opcodes either way.

### Performance vs portability

~~I do **not** have AVX-512 hardware available, so I could not run an on-device A/B performance comparison.~~ *(see follow-up comment below)* From the asm alone, the only compiler-driven VBMI use is those two instructions in ADM under Clang. That *could* be a small local win on VBMI-capable CPUs (fewer ops in that hotspot versus the shuffle/pack sequence), but it is still a tiny fraction of the AVX-512 backend and of end-to-end VMAF — any overall effect is likely small. ~~That is an inference from codegen, not a measured result; a benchmark on suitable hardware would be definitive.~~ *(measured on-device — see follow-up comment below)*

Keeping `-mavx512vbmi` on the shared AVX-512 objects still forces a VBMI **CPUID** requirement (and, under Clang, those two ops) on **all** AVX-512 builds. On machines with AVX-512 but **without** VBMI (e.g. Skylake-SP / Cascade Lake — #1555), that leads to **SIGILL** at runtime.

### Suggestion

I think we should **keep this PR as-is** (drop `-mavx512vbmi` from the common AVX-512 flags):

- Restores correct behavior on AVX-512 CPUs that lack VBMI.
- Codegen impact is limited to a local Clang change in ADM (GCC showed no VBMI opcodes in this test).
- Splitting a separate VBMI translation unit, optional Meson flag, or runtime dispatch **for two instructions** would be overkill unless profiling later shows a clear, measurable win on VBMI-capable hardware. Until then, the complexity is not justified.

Happy to add a short comment in `meson.build` documenting this tradeoff if useful.

StormBytePP:
### Update: on-device performance check (Ryzen 9 9900X)

I was able to run A/B benchmarks on hardware that **does** support AVX-512 VBMI (`avx512vbmi` / `avx512_vbmi2` in `/proc/cpuinfo`).

**Setup**

- CPU: AMD Ryzen 9 9900X (Zen 5), 12 cores / 24 threads
- Builds: this PR’s tree, Clang, `-Denable_avx512=true`
  - **no-VBMI:** baseline AVX-512 flags only
  - **with-VBMI:** same flags plus `-mavx512vbmi` (via `-Dc_args`)
- Confirmed VBMI executes without SIGILL; `objdump` on the binaries: **with-VBMI** contains `vpermt2b`/`vpermb`, **no-VBMI** contains none
- Content: 30 s mid-film clip, 1920×1080 yuv420p 8-bit, 720 frames, `vmaf_v0.6.1`
- Tool: `hyperfine` (VBMI A/B); `time` (path check)

**SIMD path**

libvmaf’s `--cpumask` is a **disable** mask (`vmaf_set_cpu_flags_mask(~cfg.cpumask)`): bit 16 disables AVX-512, bit 32 disables AVX512ICL, etc. Default (`cpumask=0`) leaves full detected flags enabled.

On this CPU, default includes AVX-512. Check at `--threads 1`:

| Mode | Wall time | FPS |
|------|----------:|----:|
| default (full, includes AVX-512) | 12.147 s | 58.54 |
| `--cpumask=48` (disable AVX-512 + AVX512ICL → AVX2 only) | 19.724 s | 36.27 |

So the default CLI path is the **AVX-512** path (and it is faster than AVX2-only on this workload). Same score in both cases: **vmaf_v0.6.1 = 87.361598**.

**VBMI A/B on default path**

`hyperfine` warmup 1, 5 runs, `--threads 24`, no `--cpumask` (AVX-512 path):

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| no-VBMI, default, threads=24 | 2.346 ± 0.015 | 2.328 | 2.363 | 1.00 |
| with-VBMI, default, threads=24 | 2.357 ± 0.009 | 2.344 | 2.365 | 1.00 ± 0.01 |

No meaningful end-to-end difference (noise level). Same score: **vmaf_v0.6.1 = 87.361598**.

**Takeaway**

- Asm: only a couple of Clang-emitted `vpermt2b` in ADM under `-mavx512vbmi` (GCC: none in earlier tests).
- On-device, on the default AVX-512 path, enabling `-mavx512vbmi` does not help end-to-end VMAF.
- Keeping `-mavx512vbmi` on the generic AVX-512 objects still risks **SIGILL** on AVX-512 CPUs without VBMI (#1555).

I still recommend **keeping this PR as-is**: drop `-mavx512vbmi` from the common AVX-512 flags.

kylophone:
Great, thanks for checking.

PR #1582, opened 2026-08-17
Fix out-of-bounds boundary reflection for tiny frames in motion feature

### LINKED ISSUE #1580: mirror()-style boundary reflection reads out-of-bounds for width/height < 3 (motion feature, all backends)
\`mirror(int idx, int size)\` reflects row/column indices at frame boundaries
for the 5-tap motion filter (radius 2):

    static inline int mirror(int idx, int size)
    {
        if (idx < 0) return -idx;
        if (idx >= size) return 2 * size - idx - 2;
        return idx;
    }

For `size < 3` this returns indices outside `[0, size)`:

- `mirror(-2, 1)` → `2` (valid range is only `{0}`)
- `mirror(-2, 2)` → `2` (valid range is `{0, 1}`)
- `mirror(3, 2)`  → `-1`

`size >= 3` is safe — I checked every tap offset (`i-2+k` for `i` in
`[0,size)`, `k` in `[0,5)`) at `size=3` and all outputs land in `[0,3)`.

This affects motion feature extraction on frames with width or height 1 or
2. The identical `mirror()` is independently duplicated in:

- `libvmaf/src/feature/integer_motion.c` (scalar reference)
- `libvmaf/src/feature/x86/motion_avx2.c`
- `libvmaf/src/feature/x86/motion_avx512.c`
- `libvmaf/src/feature/arm64/motion_neon.c`

Two related paths have the same underlying sub-3 problem but aren't
identical copies, so flagging separately rather than lumping them in:

- `libvmaf/src/feature/cuda/integer_motion/motion_score.cu` has its own,
  differently-structured mirror helper — only unsafe at size 1, not size 2.
- The float motion path's convolution boundary logic
  (`common/convolution_internal.h`) uses the same reflection formula and
  has the same sub-3 issue.

Found while adding the NEON motion implementation (#1570) — a parity test
comparing scalar vs. NEON output was occasionally flaky at width/height
1-2. My working theory (not something I've proven from the allocator
internals) is that the two independently-allocated picture buffers used by
the test happened to read different out-of-bounds memory at those sizes;
excluding those two sizes from the test made the flakiness go away. Either
way, scalar and NEON already do the identical out-of-bounds read, so this
isn't a NEON-specific bug and wasn't fixed as part of that PR.

Not proposing a specific fix here — a naive clamp could change the
intended reflection semantics rather than just guard the bounds, so it
probably needs some care. Happy to help investigate if useful.

kjg0724:
Fixes #1580.

## Bug

`mirror(int idx, int size)` reflects row/column indices at frame boundaries for the
5-tap (radius-2) motion filter. It's a single-bounce reflect-101 implementation:

```c
static inline int mirror(int idx, int size)
{
    if (idx < 0) return -idx;
    if (idx >= size) return 2 * size - idx - 2;
    return idx;
}
```

For `size < 3` this returns indices outside `[0, size)` — e.g. `mirror(-2, 1) = 2`,
out of range for a 1-row/1-column plane. The identical function was independently
duplicated in 4 places (`integer_motion.c`, `x86/motion_avx2.c`, `x86/motion_avx512.c`,
`arm64/motion_neon.c`), plus a differently-structured CUDA version, plus the shared
float convolution helpers used by `float_motion` and (via `convolution_f32_avx_*`)
`vif_tools.c`.

The float convolution path additionally has an out-of-bounds **write**: with radius 2,
`convolution_x_c_s`'s `borders_right = width - 3` goes negative for `width < 3`, so the
trailing edge loop starts before `dst[0]`. The equivalent bug exists in
`convolution_avx.c`'s vertical pass (`i_vec_end = height - radius` goes negative,
writing `tmp[-tmp_stride + j]`).

This isn't only a 1×1-frame issue. `vif_tools.c` calls `convolution_f32_avx_s` with
filter widths up to 17 (radius 8), so on any AVX2 host the same OOB write triggers for
**any dimension below 8**, at every VIF scale — not just degenerate frames.

## Fix

- `motion_mirror()` (new, replacing 4 duplicated `mirror()` copies): reflects
  *repeatedly* instead of once, so any `size >= 1` is safe. `size == 1`/`size <= 0`
  degenerate to `0` (the unique valid extension of a 1-sample signal — matches
  OpenCV's `BORDER_REFLECT_101` and scipy's `mode='mirror'`).
- Renamed from `mirror` to `motion_mirror` and consolidated into `integer_motion.h`
  (all 4 CPU files already include it). The rename is required, not stylistic: the CUDA
  translation unit also includes this header and already owns a `mirror` symbol with
  different (intentionally unrelated, untouched) reflect-even semantics — reusing the
  bare name would collide under `nvcc`.
- `convolution_mirror()` (new, in `convolution_internal.h`): same fix, kept separate
  from `motion_mirror` since this header serves other float features with their own
  style and no existing dependency on `integer_motion.h`.
- CUDA's own `mirror()`: added `if (sup == 1) return 0;` — one line, its `sup >= 2`
  formula is untouched.
- `convolution.c` / `convolution_avx.c`: clamped the edge-loop bounds so tiny
  width/height can't write out of bounds, without changing anything for normal sizes.

**Non-regression is structural, not just tested.** For the real call contract
(`idx` in `[-2, size+1]`, 5-tap/radius-2), a negative overshoot lands in `[1,2]` and a
positive one in `[size-3, size-1]` whenever `size >= 3` — both already valid, so the
new multi-bounce loop takes exactly one iteration and computes the *exact same
expression* as the old code. This is proven exhaustively in tests (below), not just
argued in the abstract.

## Testing

- **Exhaustive old-vs-new formula proofs** for all three reflection implementations
  (`motion_mirror`, `convolution_mirror`, and a host-side transcription of the CUDA
  formula), comparing every `size` in `[3, 8192]` against every in-contract offset —
  mechanically proves zero behavior change for `size >= 3`.
- **Pipeline tests** at 8 tiny sizes (`1×1` through `8×2`) across multiple bit depths,
  run through the real `motion`/`float_motion` feature extractors, asserting success,
  finite scores, and — for each size — that the score obtained via forced-scalar
  dispatch matches the score obtained via best-available SIMD dispatch (bit-exact for
  the integer path; a documented, generously-tight tolerance for the float path, since
  AVX's `-mfma` can contract the shared edge-filter's multiply-add into an FMA and
  change the last bit or two versus the non-FMA scalar build — unrelated to this fix).
  This is the only runtime coverage the AVX2/AVX512 motion kernels get from this test
  suite, since no x86 hardware was available anywhere in developing this fix (see
  below).
- **Canary-buffer bounds tests** for the OOB-*write* fix: guard regions before/after
  the real output buffer, asserted intact after the call, for the scalar convolution
  path (verified with a negative control — temporarily removing the clamps reproduces
  the failure) and, `#if ARCH_X86`-gated, the AVX path.
- **Live ASan red-then-green demonstration**, reproduced independently on two hosts/
  toolchains (macOS/Apple Clang arm64, Ubuntu 22.04/GCC 11.4.0 aarch64): built with
  `-Db_sanitize=address,undefined`, confirmed clean on the fix, then temporarily
  reverted just the two `mirror` function bodies to the old formula and confirmed a
  real `heap-buffer-overflow` abort in both the integer path
  (`motion_score_pipeline_8`) and the float path (`convolution_y_c_s`) — then restored
  and reconfirmed clean.
- Full `meson test` suite: 25/25 on both hosts.
- Python end-to-end golden-score regression suite (`quality_runner_test.py`): 67/67,
  run twice, on Ubuntu/GCC — no VMAF score on real video content moved.

## What's *not* verified at runtime

No x86 or CUDA hardware was available anywhere in developing this fix.

- **x86 (AVX2/AVX512):** verified only via cross-target `clang -target x86_64-... 
  -mavx2/-mavx512f -fsyntax-only`/`-c` checks (proves it compiles and type-checks, not
  that it's correct at runtime) plus the exhaustive formula proofs and mechanical
  rename review. This PR's own test suite gives these kernels real coverage the moment
  CI runs it on an x86 runner.
- **CUDA:** the one-line guard is exhaustively proven inert for `sup >= 2` via a
  host-side transcription, but the actual `.cu` file has never been compiled. This
  repo's CI doesn't build the CUDA path, so a maintainer with CUDA access verifying the
  guard would be appreciated.

### DIFF
--- a/libvmaf/src/feature/arm64/motion_neon.c
+++ b/libvmaf/src/feature/arm64/motion_neon.c
@@ -5,13 +5,6 @@
 
 #include "feature/integer_motion.h"
 
-static inline int mirror(int idx, int size)
-{
-    if (idx < 0) return -idx;
-    if (idx >= size) return 2 * size - idx - 2;
-    return idx;
-}
-
 // Phase 2: x_conv + abs + SAD for one row of int32 y_row.
 // Processes 4 int32 columns at a time via widening s32->s64 multiply-accumulate.
 static inline uint32_t
@@ -24,7 +17,7 @@ x_conv_row_sad_neon(const int32_t *y_row, unsigned w)
     for (j = 0; j < 2 && j < w; j++) {
         int64_t accum = 0;
         for (int k = 0; k < 5; k++) {
-            int col = mirror((int)j - 2 + k, (int)w);
+            int col = motion_mirror((int)j - 2 + k, (int)w);
             accum += (int64_t)filter[k] * y_row[col];
         }
         int32_t val = (int32_t)((accum + (1 << 15)) >> 16);
@@ -68,7 +61,7 @@ x_conv_row_sad_neon(const int32_t *y_row, unsigned w)
     for (; j < w; j++) {
         int64_t accum = 0;
         for (int k = 0; k < 5; k++) {
-            int col = mirror((int)j - 2 + k, (int)w);
+            int col = motion_mirror((int)j - 2 + k, (int)w);
             accum += (int64_t)filter[k] * y_row[col];
         }
         int32_t val = (int32_t)((accum + (1 << 15)) >> 16);
@@ -89,7 +82,7 @@ uint64_t motion_score_pipeline_8_neon(const uint8_t *prev, ptrdiff_t prev_stride
     for (unsigned i = 0; i < h; i++) {
         const uint8_t *p[5], *c[5];
         for (int k = 0; k < 5; k++) {
-            int r = mirror((int)i - 2 + k, (int)h);
+            int r = motion_mirror((int)i - 2 + k, (int)h);
             p[k] = prev + r * prev_stride;
             c[k] = cur + r * cur_stride;
         }
--- a/libvmaf/src/feature/common/convolution.c
+++ b/libvmaf/src/feature/common/convolution.c
@@ -25,11 +25,20 @@
 extern int vmaf_floorn(int, int);
 extern int vmaf_ceiln(int, int);
 
+#define MIN(x, y) (((x) < (y)) ? (x) : (y))
+#define MAX(x, y) (((x) > (y)) ? (x) : (y))
+
 void convolution_x_c_s(const float *filter, int filter_width, const float *src, float *dst, int width, int height, int src_stride, int dst_stride, int step)
 {
 	int radius = filter_width / 2;
 	int borders_left = vmaf_ceiln(radius, step);
 	int borders_right = vmaf_floorn(width - (filter_width - radius), step);
+
+	// Clamp so tiny widths (< filter_width) cannot produce negative loop bounds,
+	// which would make the trailing edge loop start at a negative index and write
+	// outside dst. No-op for width >= filter_width.
+	borders_left = MIN(borders_left, width);
+	borders_right = MAX(borders_right, borders_left);
 
 	for (int i = 0; i < height; ++i) {
 		for (int j = 0; j < borders_left; j += step) {
@@ -55,6 +64,12 @@ void convolution_y_c_s(const float *filter, int filter_width, const float *src, 
 	int radius = filter_width / 2;
 	int borders_top = vmaf_ceiln(radius, step);
 	int borders_bottom = vmaf_floorn(height - (filter_width - radius), step);
+
+	// Clamp so tiny heights (< filter_width) cannot produce negative loop bounds,
+	// which would make the trailing edge loop start at a negative index and write
+	// outside dst. No-op for height >= filter_width.
+	borders_top = MIN(borders_top, height);
+	borders_bottom = MAX(borders_bottom, borders_top);
 
 	for (int i = 0; i < borders_top; i += step) {
 		for (int j = 0; j < width; ++j) {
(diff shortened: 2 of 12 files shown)


lusoris:
I filled in part of the x86 runtime-verification gap on the current PR head, `e06467aa994081c0cd808bfee16154b663789747`.

On an AMD Ryzen 9 9950X3D (x86-64, AVX2/AVX-512 available), Linux, GCC 15.2, Meson 1.10.1, with assembly and floating-point features enabled:

```sh
meson setup build-sanitize libvmaf \
  -Denable_cuda=false -Denable_docs=false -Denable_float=true \
  --buildtype=debug -Db_sanitize=address,undefined
ninja -C build-sanitize
meson test -C build-sanitize \
  test_motion_mirror test_motion_blend test_vif_tools --print-errorlogs
```

All three targets passed under ASan/UBSan/LSan. `test_motion_mirror` reported all eight cases passing, including the scalar-vs-best-available motion comparisons and `test_convolution_avx_write_bounds`; the latter's `ARCH_X86` body was compiled and executed here. I used this PR's source directly, not our modified VMAFx motion implementation.

This is targeted CPU validation only: it does not validate the CUDA device guard, replace a full model/golden regression run, or establish a performance result. In our fork we also guard the smallest unsupported convolution-plane sizes at initialization, so that policy is separate from this PR's goal of scoring tiny inputs.

kjg0724:
Thank you for verifying this.

PR #1583, opened 2026-08-19
cuda: fix end-of-stream double flush and unordered motion SAD reset

### LINKED ISSUE #1180: Strange issue observed calculating VMAF metric values using libvmaf in CUDA mode
I'm developing SW library and using **libvmaf** successfully for vmaf metric calculation.
Unlike **vmaf** tool application, in my case I request a calculated vmaf metric value for frame N after submitting data for frame N+1 using a call to **vmaf_score_at_index()** routine.
All is working fine. Sample output in CPU mode:

$ ~/yama/build/calc_metrics  -n 10 -m vmaf -w 832 -h 480 ~/0/PartyScene_832x480_50.yuv ~/0/PartyScene_832x480.x264.VS.900.264.yuv
VMAF;VMAF_avg;
91.549303;91.549303;
91.563187;91.556245;
91.568763;91.560418;
91.072508;91.438440;
91.768058;91.504364;
90.827978;91.391633;
93.576076;91.703696;
89.840477;91.470794;
90.957639;91.413777;
90.588053;91.331204;

Recently I added code to enable great new feature added to **libvmaf** (support for CUDA GPU offloading).
Using vmaf tool sources as an example I added only a call to **vmaf_cuda_init()** routine and found the strange issue:

$ ~/yama/build/calc_metrics  -n 10 --vmaf-cuda -m vmaf -w 832 -h 480 ~/0/PartyScene_832x480_50.yuv ~/0/PartyScene_832x480.x264.VS.900.264.yuv
VMAF;VMAF_avg;
extract_fex_cuda() finished for frame 0 // it's output of printf added immediately before return from **extract_fex_cuda()**
extract_fex_cuda() finished for frame 1
91.549303;91.549303;
extract_fex_cuda() finished for frame 2
libvmaf ERROR vmaf_predict_score_at_index(): no feature 'VMAF_integer_feature_motion2_score' at index 1

Just for test, I added 1 ms delay before the call to **vmaf_score_at_index()** in my code and all results were calculated correctly without the errors.

$ ~/yama/build/calc_metrics  -n 10 --vmaf-cuda -c ccode -m vmaf -w 832 -h 480   ~/0/PartyScene_832x480_50.yuv ~/0/PartyScene_832x480.x264.VS.900.264.yuv
VMAF;VMAF_avg;
extract_fex_cuda finished for frame 0
extract_fex_cuda finished for frame 1
91.549303;91.549303;
extract_fex_cuda finished for frame 2
91.563187;91.556245;
extract_fex_cuda finished for frame 3
91.568763;91.560418;
extract_fex_cuda finished for frame 4
91.072508;91.438440;
extract_fex_cuda finished for frame 5
91.768058;91.504364;
extract_fex_cuda finished for frame 6
90.827978;91.391633;
extract_fex_cuda finished for frame 7
93.576076;91.703696;
extract_fex_cuda finished for frame 8
89.840477;91.470794;
extract_fex_cuda finished for frame 9
90.957639;91.413777;
90.588053;91.331204;

How should I get calculated vmaf metric values ASAP and fix the issue in CUDA mode?
Please advice me.

My configuration:
Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz
NVIDIA GeForce GT 1030: 1 x 3 @ 1518Mhz (OpenCL 3.0 CUDA)
x86-64 Ubuntu 22.04.1 Linux
CUTA toolkit 12.1
vmaf:
commit bf80018c3f21151b7c90324da95ef2de80f49297 (HEAD -> master, origin/master, origin/HEAD)
Author: Maximilian Müller <maximilianm@nvidia.com>
Date:   Tue Mar 14 19:10:20 2023 +0100

    reenable CPU mutlithreading in combination with CUDA

jmsether:
Two independent ordering bugs in the CUDA motion extractor. Both are in the
temporal path, both are old, and neither has an upstream fix. The first is
deterministic and stops every threaded CUDA run at end of stream; the second is
a race that silently lowers motion scores. They are separate commits and can be
taken separately, though the second is hard to observe without the first.

Line numbers below are at `f85a8536`.

### 1. CUDA feature extractors are flushed twice (`libvmaf/src/libvmaf.c`)

`flush_context_threaded()` has two loops. The first (`:649–654`) flushes every
extractor carrying `VMAF_FEATURE_EXTRACTOR_TEMPORAL`. The second (`:656–666`)
flushes the rest, and already excludes CUDA extractors at `:660`. The
non-threaded branch of `flush_context()` also already excludes them (`:681`),
because the `#ifdef HAVE_CUDA` block at `:687–694` is what flushes CUDA
extractors — after the CUDA context is synchronized.

`motion_cuda` carries both flags (`integer_motion_cuda.c:369`), so it matches
the first loop's condition and is flushed there *and* again in the `HAVE_CUDA`
block. The second flush re-appends the final frame's
`VMAF_integer_feature_motion2_score` at an index that already has a value;
`feature_vector_append()` (`libvmaf/src/feature/feature_collector.c:192`)
refuses the overwrite and returns `-EINVAL`.

Only the first loop is missing the guard the other three sites have. The fix
clones it.

The failure is deterministic on any threaded CUDA run:

```
libvmaf WARNING feature "VMAF_integer_feature_motion2_score" cannot be overwritten at index 47
libvmaf ERROR context could not be synchronized
problem flushing context
```

The middle line is misleading and worth noting if anyone has chased this
before: `err` is accumulated with `|=` from the double flush before the
`HAVE_CUDA` block runs its sync calls, so the block reports a synchronization
failure at `:707` for an error that has nothing to do with synchronization.

### 2. The motion SAD accumulator is reset on the wrong stream (`libvmaf/src/feature/cuda/integer_motion_cuda.c`)

In `extract_fex_cuda()`:

```c
280:  CHECK_CUDA(cu_f, cuMemsetD8Async(s->sad->data, 0, sizeof(uint64_t), s->str));
...
284:  s->calculate_motion_score(ref_pic, ..., s->sad, ...,
286:          cu_f, vmaf_cuda_picture_get_stream(ref_pic));
```

`:280` zeroes `s->sad` on `s->str`. `:284` launches the kernel that
`atomicAdd`s into `s->sad`
(`libvmaf/src/feature/cuda/integer_motion/motion_score.cu:79`, and `:123` for
16-bpc) on `vmaf_cuda_picture_get_stream(ref_pic)`. **There is no event, no
wait, and no implicit ordering between those two streams.**

They are unconditionally different streams, and structurally so:

* `s->str` is created per extractor at `:142` with `CU_STREAM_NON_BLOCKING`;
* every CUDA picture carries its own stream — `cuda/picture_cuda.c:192` creates
  one per picture in the pool — so the kernel also runs on a *different* stream
  from frame to frame.

Non-blocking streams do not serialize against each other. What usually saves
the memset is wall clock, not ordering: `:265` drains `s->str`, so the 8-byte
memset is submitted into an empty stream and has the host-side work of
`:283–:286` (a `cuStreamWaitEvent` and a `cuLaunchKernel`, plus context
push/pop) to complete in. When driver submission is delayed — queue depth,
another process on the device, a scheduler hiccup, WDDM batching on Windows —
the memset lands *after* the kernel has begun its `atomicAdd`s and wipes the
contributions that already arrived.

The signature is what an accumulator zeroed mid-flight looks like: the reported
score is the correct score minus an arbitrary fraction. Across six captured
outliers the loss ranged from 0.015 % to 3.9 %, and **all six were low** — the
error is one-sided. None of the corrupt values equalled any other frame's
score, which rules out a frame-pairing slip.

About half of the hits never reach `vmaf`: `motion2` is `min(prev, cur)`
(`:252`), so a downward corruption of the larger member of a pair is discarded
by the `min`. That is masking, not safety — the same defect, half of it silent.

**The fix** is to issue the reset on the stream that consumes it. One argument,
no new events, no new allocations, no special case for the first frame. The
ordering that replaces the race is CUDA's own program order within a single
stream. The move is safe against the surrounding code: the reset stays behind
the *previous* frame's device-to-host read of `s->sad` (that copy is on
`s->str`, which `:265` drains before anything here is enqueued), and enqueuing
it ahead of `:283`'s wait still orders it ahead of the kernel on the same
stream.

### Deterministic falsifier

The race is timing-dependent, so here is a scratch edit that makes it fire
every time — a maintainer can run it to confirm the missing edge in one build.
It queues ~8 GiB of memset work (~10 ms of GPU time) on `s->str` immediately
ahead of the 8-byte SAD reset. **It reorders nothing and deletes nothing; it
only makes `s->str` slow.** If the reset and the kernel were ordered, the
scores are unchanged. Apply on top of commit 1 (otherwise the run aborts at end
of stream before writing output) and *without* commit 2:

```c
/* in MotionStateCuda */
    VmafCudaBuffer* amp;

/* in init_fex_cuda(), beside the s->sad allocation */
    ret |= vmaf_cuda_buffer_alloc(fex->cu_state, &s->amp, 1024u * 1024u * 1024u);
    if (ret) goto free_ref;

/* in extract_fex_cuda(), immediately before the "Reset device SAD" memset */
    for (int amp_i = 0; amp_i < 8; amp_i++) {
        CHECK_CUDA(cu_f, cuMemsetD8Async(s->amp->data, 0,
                                         1024u * 1024u * 1024u, s->str));
    }
```

Result at `f85a8536` on an RTX 3090, 48-frame 1080p 10-bit pair, three runs of
three: every frame's `integer_motion` is exactly `0.0` — 48/48 frames, every
run — and `integer_motion2`, being a `min()` over consecutive motion values,
follows. A 10 ms delay applied to one of the two operations
annihilates the result, which is only possible if the other is not waiting for
it.

### Natural rate

Unpatched, scoring a fixed 48-frame 1080p 10-bit pair repeatedly with
`--gpumask 0 --threads 8` and comparing full per-frame score vectors between
runs:

| host | runs | runs with a corrupted motion score |
|---|---|---|
| Linux x86-64 (WSL2, CUDA 12.9), RTX 3090 / RTX 4090 | 117 | 9 (7.7 %) |
| Windows x86-64 (native, CUDA 13.0), RTX 3090 | 53 | 1 (1.9 %) |

That is ~1.6e-3 per scored frame pair on x86-64. Worst observed single-frame
error on the final `vmaf` score was 1.0e-01.

**We did not observe it on aarch64.** On two GB10 devices, unpatched, the same
instrument produced **zero** events in **14 190 scored frame pairs** (200 runs
of the 48-frame pair, plus 10 runs of a 479-pair span). At the x86-64 rate that
predicts ~23 events; P(0 | x86 rate) ≈ 1.4e-10. We are stating this as a
measured rate difference, not an explanation — we did not determine the cause,
and the missing edge is present in the same source on every platform, so we
would not read this as aarch64 being safe. Plausible contributors we could not
separate: no WDDM batching on Linux, single-tenant headless boxes, and an
integrated part with a unified memory controller where an 8-byte device memset
is a very different operation than across a discrete PCIe card.

If nothing else, it may explain why intermittent CUDA-mode reports have been
hard to reproduce.

## Validation

* **300 runs of the patched build, zero mismatches**, across four physical GPUs
  (RTX 3090, RTX 4090, two GB10) and three platforms (Linux x86-64 under WSL2,
  native Windows x86-64, Linux aarch64), against 9 in 117 for the unpatched
  control on the same host and content. Plus 20 patched runs of a 479-pair
  span — 9 580 scored pairs, zero disagreements.
* **The scores do not move.** On runs that did not hit the race, patched and
  unpatched binaries produce byte-identical per-frame score vectors, and the
  patched build at `f85a8536` reproduces the same vector we recorded on earlier
  revisions. Method: SHA-256 over the JSON `frames` array canonicalized as
  LF-terminated `<frameNum>\t<metric>\t<repr(value)>` lines, all frames in
  ascending order and metric names sorted, with `fps`, `version` and
  `pooled_metrics` excluded.
* **Blast radius.** The same binary run with `--gpumask -1` produces an
  unchanged score vector, confirming nothing outside the CUDA motion extractor
  moved.
* **Error direction.** The corrupted SAD is always low, never high, so
  `motion`, `motion2` and therefore `vmaf` are biased *downward* under the
  race — a distorted file looks slightly worse than it is, never better.
* **Test suite.** `ninja -C libvmaf/build test` gives identical results with
  and without these patches on our machine. One test fails in both cases on a
  CUDA-enabled build here — `test_cuda_pic_preallocation`, SIGSEGV — which
  reproduces unpatched at `f85a8536` and is untouched by this PR.

## Scope

No API change, no ABI change, no score change, no new allocations, events or
options. Commit 2 touches a CUDA-only translation unit. Commit 1 adds one flag
test in common code that is a no-op when no CUDA extractor is registered, so
non-CUDA builds are unaffected.

No new files, so no new license headers; both modified files already carry the
BSD+Patent header. This contribution is licensed under BSD+Patent per
`CONTRIBUTING.md`.

## Related, not fixed

#1180 reports a different CUDA-mode synchronization failure —
`vmaf_predict_score_at_index(): no feature 'VMAF_integer_feature_motion2_score'
at index 1` when a caller reads a score immediately after submitting the next
picture, worked around with a 1 ms sleep before the read. That is a *missing*
feature at read time: `motion_cuda` publishes motion2 from a `cuLaunchHostFunc`
callback on `s->host_stream` (`integer_motion_cuda.c:321`), so the value need
not be in the collector when the caller asks for it. The bug fixed here
produces a *wrong* value, never a missing one. **These commits do not close
#1180**; we mention it because it is the only existing report of unsynchronized
behaviour in this extractor and the two are neighbours.

One more thing we noticed and deliberately left alone: `:283` waits the
*reference* picture's stream on the *distorted* picture's ready event, while
the kernel at `:284` reads the reference. That is benign today only because
each picture's upload is issued on its own stream, so the reference upload is
implicitly ordered ahead of a kernel on that same stream. It looks like a typo
rather than intent — happy to send a follow-up if you agree.


jmsether:
A third instance of this PR's defect class, in the neighbouring extractor, and it is present at this PR's own base.

`libvmaf/src/feature/cuda/integer_vif_cuda.c`, `extract_fex_cuda()` at `f85a8536`:

```c
458:  CHECK_CUDA(cu_f, cuMemsetD8Async(s->buf.accum_data->data, 0, sizeof(vif_accums) * 4, s->str));
...
466:  filter1d_8 (..., vmaf_cuda_picture_get_stream(ref_pic));   // scale 0, 8 bpc
468:  filter1d_16(..., vmaf_cuda_picture_get_stream(ref_pic));   // scale 0, 16 bpc
470:  filter1d_16(..., s->str);                                  // scales 1..3
```

`:458` zeroes all four scales' accumulators on `s->str`. `:466`/`:468` launch the kernels that `atomicAdd` into `accum[0]` (`integer_vif/filter1d.cu:423`: warp-reduce, then one atomic per warp) on the picture's stream. Same missing edge as patch 2 of this PR: `s->str` is per-extractor `CU_STREAM_NON_BLOCKING` (`:101`), every picture carries its own stream (`cuda/picture_cuda.c:192`), and non-blocking streams do not serialize against each other.

**Scales 1-3 are the built-in control.** They are launched on `s->str` at `:470`, program-ordered behind the reset, and they never corrupt. The damage is confined to exactly the scale the code says is unordered.

**The symptom is louder than patch 2's.** VIF's `write_scores` divides bare at `:367-381` (`vif.scale[0].num / vif.scale[0].den`). A total wipe leaves `num == den == 0`, so this is a literal `0.0f/0.0f` -> NaN. `vmaf_v0.6.1`'s `score_clip` cannot clip a NaN (every comparison against NaN is false), so it reaches `output.c:224` and the writer spells the per-frame score `null`. With two `--model` flags the exposure is deterministic and lands on whichever registers first: swapping the flags moves the nulls to the other model, 16/16 runs. About a fifth of hits are *partial* wipes instead: finite, one-sided low, 0.0003 % to 0.64 %, and those pass every parser.

**Falsification.** `CUDA_DEVICE_MAX_CONNECTIONS=1` collapses independent non-blocking streams onto one hardware queue, so they execute in issue order, and the memset is issued first. If the diagnosis is right this must remove the corruption entirely; if it survives, the window is elsewhere. Measured on an RTX 4090 + 3090 box, 700 frames of 1080p10 identity per run:

| configuration | free-running | `CUDA_DEVICE_MAX_CONNECTIONS=1` |
|---|---|---|
| two `--model` flags | 1 241 corrupt frames / 8 400 (14.8 %) | **0 / 2 800** |
| + 3 extra `vif_cuda` instances | 1 227 / 2 100 (58.4 %) | **0 / 2 100** |

**The fix** is this PR's own, one file over: issue the reset on the stream that consumes it.

```c
CHECK_CUDA(cu_f, cuMemsetD8Async(s->buf.accum_data->data, 0,
            sizeof(vif_accums) * 4, vmaf_cuda_picture_get_stream(ref_pic)));
```

Safe against the surrounding order: scales 1-3 stay behind the reset through the existing `s->event` edge (`:475` records on the picture stream after the scale-0 kernels, `:476` waits on `s->str`); the DtoH read at `:487` stays behind every accumulation through that same event; and the previous frame's read of `accum_data` is already drained by `:455`'s `cuStreamSynchronize(s->str)`.

With the patch, same box and same runs: **0 corrupt frames in 21 000 two-device frames** across both configurations above, against ~4 900 predicted by the pre-fix rate. Per-frame scores are bit-identical to the unpatched engine under a single-GPU pin, where the unpatched engine is clean (17 500 values per run, on both cards), so this removes the race and not the answer. It applies to `f85a8536` with `git apply` clean.

Two notes, offered rather than assumed:

- **A `den == 0` guard would be the wrong fix here**, even though `integer_adm_cuda.c:679` has one. It would turn every total wipe into a *perfect* VIF score of 1.0, converting the loud half of this defect into the silent half. The same function already guards at `:393-394`, but only for the aggregate `integer_vif`, behind the `if (!s->debug) return err;` early-out at `:383`, i.e. on the branch that does not ship.
- Unrelated and cosmetic: the `null` branch at `output.c:224` omits the `\n` its sibling cases print, so a corrupted frame's metrics run together on one line. Valid JSON, but it makes line-oriented reading of the output misleading.

Happy to send this as a third commit on this PR, or separately, whichever you prefer.

PR #1584, opened 2026-08-22
avx2: vectorize a couple scalar parts

anematode:
Hi :) Hoping to make some performance improvements to the kernels. On my Zen 5 machine I get

4229894.000000 us/4316783.000000 us = 2.0% speedup


lusoris:
**Verdict:** the logic is right and it's a real win. I verified it; the PR should carry that verification itself. Fix the nits below, add the evidence, squash, and this is mergeable.

**What I verified (should have been in the description):**

- AVX2 vs pure C (`--cpumask`) on a synthetic 576x324 clip with flat bands, 0/255 stripes at periods 4..64 and a noise patch: every `integer_adm*` / `integer_vif*` value identical, and identical to master's AVX2.
- checkasm with `fill_band_i32` widened from ±8000 to ±1,000,000, so that `abs >= 32768` actually occurs and `get_best15_from32_256` is exercised: 20 seeds, all adm tests pass. **With the stock inputs this function is dead on every lane**, so the CI badge says nothing about it.
- checkasm bench, i9-12900K P-core, 64x48, cycles:

| function | master avx2 | this PR | delta |
|---|---|---|---|
| `adm_decouple_64x48` | 10321 (1.05x vs C) | 8199 (1.33x) | −21% |
| `adm_decouple_s123_64x48` | 24998 (**0.69x** vs C) | 18324 (0.93x) | −27% |

Note that master's s123 AVX2 path is *slower than scalar* on this CPU and your PR only brings it to 0.93x. Not your bug, but "2.0% speedup" undersells the kernel gain and hides that the kernel is still net-negative vs C. Put the kernel numbers in the description.

**Correctness reasoning that belongs in the code, not in a reviewer's head:**

- `SPLIT_TO_DOUBLES` goes int32 → float → double. That isn't sloppy, it is *required*: the scalar does `(float)ot_dp / 4096.0`, so it rounds to float first, and `/4096.0` is an exact power-of-two scale that can be dropped. Write that in a comment, or the next person "fixes" it to `_mm256_cvtepi32_pd` and breaks bit-exactness.
- `get_best15_from32_256`: the `andnot(srli(temp, 8), temp)` trick works because it guarantees bit `msb-8` is zero, which lies inside the 24-bit float mantissa, so round-to-nearest-even cannot carry into the next exponent. Your comment says "clears enough of them"; state the invariant. The precondition "sign bit never set" needs one sentence on what bounds `|band|` at s123. Also say explicitly that lanes with `temp < 32768` compute garbage that the following `blend` discards, so `temp != 0` is not actually required.
- `double_to_64` (magic 2^52+2^51) is valid for |v| < 2^51. Say why the input satisfies it (`|rst| < 2^31` times the gain limit).

**Nits:**

- `_mm256_cmp_pd(..., 5)`: use `_CMP_GE_OQ` (or at least spell `_CMP_NLT_US`). The file already uses `_CMP_GT_OS` by name thirty lines later.
- `(__m256i)_mm256_cmp_pd(...)` C-style vector cast: use `_mm256_castpd_si256`, as the code at the `mask_gt_h_lo_epi64` site already does. Don't add more of the non-portable style.
- Function-local `#define CALC_ANGLE_256` / `SPLIT_TO_DOUBLES` are never `#undef`'d, and `SPLIT_TO_DOUBLES` ends with a stray `\` continuing onto a blank line. Make them `static inline` functions.
- The PR converts a few unrelated tab-indented lines to spaces, then adds four new tab-indented lines in `merge_64_to_32`. Pick one.
- `merge_64_to_32` and `double_to_64` are duplicated under different names in #1585's `vif_avx2.c` (`merge_64_to_32`, `cvttpd_epi64_256`). One shared x86 helper header.
- Two commits ("vectorize some scalar parts", "vectorize 64-bit to int32 and double conversions"). Squash; subject along the lines of `feature/adm/x86: vectorize calc_angle, get_best15_from32 and 64<->32/double conversions in adm_decouple*_avx2`.
- Textual conflict with #1609 (both edit around `blend`). Say which merges first.

PR #1585, opened 2026-08-22
avx2: vectorize more of vif_statistic_8_avx2/vif_statistic_16_avx2

anematode:
3478896 us/4192474 us -> 17% speedup

With some tricks we can vectorize the residual calculation as well. Should be exactly the same as scalar


lusoris:
**Verdict:** correct as far as I can establish, and a real win on Intel too. But "Should be exactly the same as scalar" was asserted, not shown, and the paths where it could differ are exactly the ones CI never runs. One change requested (item 1), the rest is documentation and a benchmark hook.

**What I verified:**

- Stock checkasm fills ref/dis with uniform random noise, so `sigma1_sq >= 2*65536` on every pixel and the `accum_num_non_log` branch **never executes** in CI. I rebuilt checkasm with `check_vif_statistic_8` filling 32x32 flat blocks plus sparse ±1 perturbations: 20 seeds, passes.
- AVX2 vs pure C on a synthetic clip with flat bands (guaranteed `sigma1_sq < 2*65536`), stripes and noise: all `integer_vif_scale*` identical, and identical to master.
- Wall-clock, `--feature vif` only, 1920x1080x4, `--threads 1`, i9-12900K P-core, 5 interleaved runs: master 156-158 ms, this PR 129-136 ms, about **−15%**. So `vpgatherdd` is not a problem on a GDS-mitigated Intel core; good to know, and it should be in the description.
- checkasm cannot benchmark this function: `check_vif.c` has **zero `bench_new()` calls**. Add them in this PR. That is the number Kyle asked for, in the form he asked for it.

**Issues:**

1. `accum_num_non_log`: the reference adds `int32_t sigma2_sq` **sign-extended**; you use `_mm256_cvtepu32_epi64` (zero-extend). Equivalence therefore rests on `sigma2_sq >= 0` in that branch, which nobody has proven; the reference guards `sigma2_sq > 0` elsewhere precisely because it can be ≤ 0. I could not produce a negative one on 8-bit content, but `_mm256_cvtepi32_epi64` costs nothing and makes the equivalence unconditional. Please change it.
2. `log2_index_32_256` silently requires `0 < temp < 2^31`: with the sign bit set, `cvtepi32_ps` yields a negative float and the exponent extraction returns garbage. The scalar `log2_32` handled the full `uint32_t` range via `__builtin_clz`. `sigma1_sq + 2^17` and `sv_sq + 2^17` are int32 adds that can wrap for large variances. Either document why the inputs are bounded (they appear to be for 8-bit; the 16-bit path with 12-bit input deserves a sentence) or handle it. You did write "Requires 0 < temp < 2**52" on the 64-bit version; the 32-bit one deserves the same.
3. `log2_table_gather_256` does 4-byte gathers at byte offset `index*2`; for `index == 65535` it reads `log2_table[65536]`, which exists only because the table is declared `[65537]`. Make that dependency explicit in a comment.
4. Per-lane int32 accumulators for `den_log` / `num_log` / `den_non_log`, reset per row: worst case ≈ 30720 × w/8 per lane, overflow around w > 500k. Fine, but say so in a comment; the scalar used int64 and someone will wonder.
5. `uint32_t xx[static 16]`: `[static N]` is C99-only and MSVC rejects it. Not a CI blocker, but nothing else in the file uses it.
6. `merge_64_to_32` and `cvttpd_epi64_256` duplicate #1584's `merge_64_to_32` / `double_to_64`. One shared header.
7. Description: "3478896 us/4192474 us -> 17% speedup" of *what*? Whole run or vif only, which clip, bit depth, thread count, compiler? Put the command line in.
8. Commit subject: `feature/vif/x86: vectorize the per-block log/accumulate loop in vif_statistic_{8,16}_avx2`.

PR #1586, opened 2026-08-22
avx2: use vpabsd

anematode:
Instead of sign-extending to 64-bit and then taking the absolute value, we can zero extend first and take 32-bit absolute values (or the other way around, doesn't matter)


kylophone:
Do you have benchmarks?

anematode:
Not immediately, but seems like a pretty straightforward improvement to me?

kylophone:
For this and all of your other SIMD PRs, could you please integrate with checkasm (https://github.com/Netflix/vmaf/pull/1592)? This will give us correctness checks and benchmarks. Once those tests pass, we should be able to merge these PRs. Thanks for your patience!

anematode:
@kylophone Hi! I've rebased the PRs, is there something else that I ought to do?

lusoris:
Answering "is there something else that I ought to do?" — this applies to #1584, #1585, #1586 and #1609.

Yes, and it's the same thing on all four: **evidence**. I went through all four diffs, rebuilt each branch, ran checkasm with inputs that actually reach the changed code, diffed AVX2 against pure C on synthetic content, and benchmarked on an Intel core. The code is essentially correct (details on each PR). None of that verification is in the PRs, and a reviewer shouldn't have to redo it to click approve.

Concretely, what "integrate with checkasm" means for these PRs:

1. **Run checkasm locally and paste the result.** Today the only proof is a green CI badge, and that badge is weaker than it looks (next point).
2. **The existing checkasm inputs never reach the code you changed.** `check_adm.c` fills the i32 bands with values in ±8000, so `abs >= 32768` in `adm_decouple_s123_avx2`, i.e. the whole `get_best15_from32_256` path in #1584, is masked out on every lane of every CI run. `check_vif.c` fills with uniform random noise, so `sigma1_sq < 2*65536` never happens and the `accum_num_non_log` path in #1585 is never executed. Widen the inputs in the same PR (`fill_band_i32` to ±2^20; a flat-block pattern for vif). I did both locally and your code passes 20 seeds each, so this costs you nothing but proves the thing that matters.
3. **`check_vif.c` has zero `bench_new()` calls**, so checkasm currently cannot benchmark `vif_statistic_*` at all. #1585 should add them. That is what turns "3478896 us/4192474 us" into a number someone else can reproduce.
4. **PR descriptions need a command line.** Content, resolution, bit depth, features enabled, threads, compiler, run count. "On my Zen 5 machine" is also the one x86 microarchitecture where every one of these changes is cheapest (1-uop `vpblendvb`, fast gathers). On an Alder Lake P-core, #1584 and #1585 are real wins; #1586 and #1609 are inside the noise floor. Numbers on each PR.
5. **Commit subjects.** Upstream uses `feature/adm/x86: <what, and why it is safe>`. "vectorize some scalar parts", "use vpabsd", "Fix indentation" don't say what changed or why the result is bit-exact.
6. **Ordering.** All four edit `adm_avx2.c` / `vif_avx2.c` independently. #1584 and #1609 both touch the `blend` helper region and will conflict. Stack them or say which merges first. #1609 is also based on `f85a85369`, which predates checkasm (so it can't run CI) and predates 03b5562c5, the out-of-bounds fix in the very function it edits.

None of this is a big ask. It's the difference between a reviewer approving in ten minutes and a reviewer spending an afternoon rebuilding your evidence.

lusoris:
**Verdict:** correct, and equivalent to the old code including the `INT32_MIN` corner (zero-extend then `abs_epi32` leaves `+2^31` in the 64-bit lane, same as the old sign-extend-and-negate path). Worth taking as a cleanup. But "seems like a pretty straightforward improvement to me?" is not an answer to "do you have benchmarks?", so here is one.

**Bench,** checkasm, i9-12900K P-core, `adm_csf_den_s123_64x48_scale1_avx2`, cycles:

| | seed 1 | seed 2 |
|---|---|---|
| master | 5603 | (noise floor ±2%) |
| this PR | 5582 | 5599 |

No measurable change. Twelve uops fewer per iteration doesn't matter because the loop isn't bound on them. That is fine, but then describe it as a cleanup (−9 lines, clearer), not a performance change, unless your Zen 5 numbers say otherwise, in which case put them in the description with the command line.

**Verified:** stock checkasm passes (this function's inputs do reach the abs path); AVX2 vs pure C identical on synthetic content, identical to master.

**Nits:**

- `_mm_abs_epi32` on the 128-bit load *before* `_mm256_cvtepu32_epi64` is the same op count but reads as "abs, then widen", which is what the scalar does. Either is fine.
- Commit subject: `feature/adm/x86: use vpabsd for |src| in adm_csf_den_s123_avx2`.
- This file is also edited by #1584 and #1609. No conflict with this one, but say the intended merge order somewhere.

Issue #818, opened 2021-02-01
libvmaf API does not support `perc*` pool methods


cosmin:
The libvmaf API used by ffmpeg does not support the `perc5, perc10, perc20` pooling methods. Furthermore attempting to pass one of these values will silently fallback to mean as the pooling method due to the implementation in `pool_method_map`, leading to a confusing user experience (even though the documentation mentions that only `min`, `mean` and `harmonic_mean` are supported).

It would be great to either unify the pooling methods such that `perc*` works, or to perhaps consider returning an error if an unsupported pooling method is requested.

HunterAP23:
I'm pretty sure the main reason for this is that FFmpeg uses the `compute_vmaf.c` code which is set to be deprecated on the VMAF library. That file only has `min`, `mean` and `harmonic_mean` as per this section:

https://github.com/Netflix/vmaf/blob/bf5e496c4c68875826a9da967f036ddcfc128142/libvmaf/src/compute_vmaf.c#L23-L32

the only other mention of `percX` pooling methods I could find were only present in the python library which uses numpy to get those percentile points. `percX` doesn't seem to be implemented in C/C++ for FFmpeg to import.

cosmin:
Yes, it would require adding percentile support to `vmaf_feature_score_pooled` in `libvmaf.c` but it seems like it would be fairly straightforward.

HunterAP23:
Not trying to plug my code, but I've been working on a python script that takes VMAF report files and turns them into a chart image and a video that graphs the different VMAF values over time. One of the items I provide on the legend is the 1% and 0.1%, and 0.001% lowest percentiles. It's still a work in progress and not yet complete, but if you're willing to use an external program to calculate different percentiles of the different metrics (like PSNR and SSIM) then you can check it out: https://github.com/HunterAP23/VMAF-Over-Time

lusoris:
I submitted the C API portion as #1589, adapted from VMAFx's merged implementation. It adds `VMAF_POOL_METHOD_MEDIAN`, `PERC5`, `PERC10`, and `PERC20`, with linear interpolation matching the Python harness's percentile rule and the existing interval/subsampling selection.

The patch preserves the existing pooling methods and includes five regression cases; the full upstream release suite passes 24/24 and the new cases pass ASan/UBSan/LSan. It is based on current upstream master, not our fork's newer C++/dependency configuration.

[...]

PR #1589, opened 2026-09-08
libvmaf: add median and percentile C API pooling

lusoris:
The C pooling API cannot currently select the median or lower percentiles requested in #818. This adapts [VMAFx #1340](https://github.com/VMAFx/vmafx/pull/1340) to current upstream C11: append `MEDIAN`, `PERC5`, `PERC10`, and `PERC20`, gather the scores selected by the inclusive interval and `n_subsample`, then sort and linearly interpolate at `percentile * (n - 1) / 100`. Existing pooling values, accumulation order, and allocation behavior remain unchanged. The README and public header document the semantics and memory cost.

The implementation explicitly rejects non-finite samples because current upstream permits importing them (unlike the fork's collector). It propagates missing-score errors, rejects an empty selected interval, checks allocation size, and uses 64-bit interval arithmetic to avoid wrapping at `UINT_MAX`. The feature-pooling output is unchanged on failure.

This adds the C API functionality. FFmpeg's filter option mapping needs a separate FFmpeg-side change, so this does not claim that its `pool` option immediately accepts the new values. No CLI option, model, bootstrap-confidence-interval formula, or golden assertion changes.

Correction, 2026-09-19: as submitted until today (`fe73c3fd`, rebased as `e2c17610`) this PR broke the XML and JSON reports. Their writers in `output.c` loop over every method below `VMAF_POOL_METHOD_NB` and index a name table with four entries, so with the four new enum values they read past the table and printed garbage keys in `pooled_metrics`; ASan reports a global-buffer-overflow at `output.c:131`. The writers now stop at the end of the name table. Reports keep listing `min`, `max`, `mean` and `harmonic_mean` only, the percentile methods are reachable through the API, and XML, JSON, CSV and subtitle output of the `vmaf` tool is identical to upstream's apart from the version string and fps. A sixth regression case writes a JSON report through `vmaf_write_output()` and requires exactly those four pooled entries; it fails against the earlier state of this branch with “pooled psnr_y does not have exactly four entries”.

Validation on upstream `86da14d0306a138fd3f01319860b905169746516`, x86-64 Linux, GCC 16.2.1 / C11, Meson 1.12.0 (rebased on 2026-09-19 from `f85a853692a8c730d0270cd733c8bb30b5b93b7c` to resolve a conflict with the checkasm block in `libvmaf/test/meson.build`; the test's Meson block now sits next to `test_feature_collector` so that this PR and #1590/#1591 no longer conflict with one another; apart from that and the correction above the patch is unchanged, and the results below were re-run, including the negative control):

```sh
meson setup build libvmaf -Denable_cuda=false -Denable_docs=false -Denable_float=true --buildtype=release
ninja -C build
meson test -C build --print-errorlogs
# 24/24 passed

meson setup build-sanitize libvmaf -Denable_cuda=false -Denable_docs=false -Denable_float=true --buildtype=debug -Db_sanitize=address,undefined
ninja -C build-sanitize
meson test -C build-sanitize test_pool_percentile --print-errorlogs
# 6 regression cases passed with ASan/UBSan/LSan
```

Cases cover interpolation on unsorted scores without mutating their stored order, singleton/tied/negative values, subsampling and partial intervals, missing/non-finite scores, invalid arguments, `UINT_MAX` endpoints, and delegation from cached model scores. The new interpolation test fails when only `libvmaf.c` is replaced by the original upstream file. Sanitizer coverage claimed here is the targeted test; the full upstream baseline has unrelated LeakSanitizer failures in `test_predict` and `test_pic_preallocation`, recorded while validating #1588.


lusoris:
Correction to this PR: as submitted it corrupted the `pooled_metrics` section of XML and JSON reports. The report writers in `output.c` loop over every method below `VMAF_POOL_METHOD_NB` and index a name table with four entries, so the four new enum values made them read past the table. The `vmaf` tool printed garbage keys after `harmonic_mean`, and ASan reports a global-buffer-overflow at `output.c:131`. The earlier validation ran the unit tests but never wrote a report, which is why it was not caught.

Fixed in `095b4c20`: the writers stop at the end of the name table, so reports list `min`, `max`, `mean` and `harmonic_mean` as before and the percentile methods stay API-only. XML, JSON, CSV and subtitle output is identical to upstream `86da14d0` apart from the version string and fps, and a sixth test case writes a JSON report and requires exactly four pooled entries (it fails against the earlier state of the branch). The branch was also rebased onto `86da14d0`; the description has the details.

PR #1595, opened 2026-09-17
cuda: keep kernel parameter structs and accumulators out of local memory

birkdev:
The __device__ helper kernels take AdmBufferCuda, AdmFixedParametersCuda, VifBufferCuda and filter_table_stuct by value. nvcc forwards those copies to param space, but clang's NVPTX backend copies the 584 and 248 byte structs into a local memory depot with a byte loop as soon as they are indexed with a runtime value like blockIdx.z. The filter1d horizontal kernels have the same problem with their seven int64 accumulators: nvcc unrolls that loop on its own, clang doesn't, and the array lands in local memory.

Passing the structs by const reference and adding #pragma unroll on those loops fixes both. No kernel logic changes. nvcc output is unchanged (same local memory footprint, integer_adm and integer_vif bit identical before and after). With clang, adm_csf went from 57x slower than nvcc to parity and total kernel time is within 3% of nvcc.

This is the first half of making -Denable_nvcc=false usable without the CUDA toolkit headers, see #1596.


birkdev:
The second part, the clang build path this prepares for, is #1596.

PR #1596, opened 2026-09-17
cuda: build the kernels with clang without the CUDA toolkit headers

birkdev:
Builds on #1595.

#1436 added -Denable_nvcc=false to compile the kernels with clang, but -nocudainc was left commented out because device intrinsics broke without the toolkit headers. This finishes that: a small header (src/cuda/cuda_runtime_compat.h) includes the self-contained CUDA wrappers that clang ships, the way clang's own runtime wrapper does, and adds the few things that live in toolkit headers (qualifiers, vector types, __shfl_down_sync, atomicAdd, double min/max). The kernels are emitted as PTX and JIT compiled by the driver. The toolkit is still needed for libdevice and bin2c, nothing else. Device and host code both use nv-codec-headers only.

The reason: the toolkit headers don't support a MinGW-w64 host (crt/math_functions.hpp takes a pre-2013 MSVC code path when _WIN32 is defined without _MSC_VER), and nvcc can't be configured there either, so today there is no way to build CUDA support with MSYS2 at all (#1154). With this the normal MinGW toolchain is enough.

Tested on Windows 11 (MSYS2 clang 22, CUDA 13.4) and in an Ubuntu 24.04 container (clang 18, libstdc++ 13, CUDA 12.8 with nvcc for comparison), both on an RTX 5090. integer_adm and integer_vif are bit identical between clang and nvcc on every frame on both platforms, and kernel time is within 3% of nvcc. meson test gives the same results as the nvcc build.

The header only covers what the kernels use today; a new intrinsic would fail the clang build with an undeclared identifier, which is intentional. The toolkit root is taken from the location of bin2c, with -Dcudatoolkit_path as an override.


birkdev:
@gedoensmax this finishes the `-nocudainc` path you left commented out in #1436; would you have a look?

gedoensmax:
Did you try to run some 8bit and 10/12 bit footage to check if VMAF scores are correct ? You mention some meson tests, but i think the coverage is subpar.

birkdev:
Only 8-bit before your comment, fair point. I now ran 8, 10 and 12-bit 1080p clips (120 frames each, x265 re-encodes against y4m references) through the clang build on Windows and Linux, the nvcc build on Linux, and the CPU path. integer_adm and integer_vif are bit-identical across all of them at every depth. At 10 and 12-bit the full VMAF score matches too (90.5252 and 88.5416 on every path); at 8-bit only integer_motion differs by a little between runs, and it does that between two runs of the same nvcc binary as well, so it looks like an existing race in the motion extractor rather than anything in this PR. I'll open a separate issue for that. Happy to run any other footage you have in mind.

gedoensmax:
@kylophone could you take a look at this ? This would enable compilation without any CUDA Toolkit dependencies from what i understand. Basically enabling regular builds with ffmpeg i assume.

PR #1598, opened 2026-09-17
.github/ci: compare checkasm benchmarks against the PR's base ref

kylophone:


kylophone:
The Github CI runners seem way too inconsistent to be an accurate benchmark...

anematode:
@kylophone Thoughts on comparing the ratio optimized/generic rather than the absolute number ? Of course assuming that the generic impl doesn't change much, which will be true in practice. The ratio should be less noisy methinks

PR #1602, opened 2026-09-19
feature/adm: keep contrast masking thresholds in range

lusoris:
`adm_cm()` used to narrow the scale-0 centre tap through `int16_t` in the scalar, x86 edge, and CUDA paths, while the AVX2 and AVX-512 interior kept the full 32-bit value. For bounded scale-0 DWT coefficients above 15360, that made the threshold wrap in some paths and not others.

The original revision made the SIMD interior reproduce that wrap. Review correctly pointed out that the wrap itself was wrong. Commit `21e6178` fixes the source of the mismatch instead:

- keep the centre tap in 32 bits in scalar, x86 edge, and CUDA code;
- calculate scalar/CUDA threshold subtraction in 64 bits and clamp the post-threshold value to the representable 32-bit range;
- explicitly zero AVX2/AVX-512 lanes whose shifted threshold would exceed `INT32_MAX`, preserving `max(abs(x) - threshold, 0)` without signed-shift overflow.

The regression test fills `decouple_r` and `decouple_a` with bounded scale-0 coefficients, runs `adm_csf()`, and compares scalar, AVX2, and AVX-512 results across multiple frame widths. It also covers values that make the rescaled threshold exceed `INT32_MAX`.

Validation against the current upstream base:

```sh
meson setup build-release libvmaf --buildtype=release -Denable_cuda=false -Denable_float=true -Denable_docs=false
ninja -C build-release
meson test -C build-release --print-errorlogs

meson setup build-sanitize libvmaf --buildtype=debug -Db_sanitize=address,undefined -Denable_cuda=false -Denable_float=true -Denable_docs=false
ninja -C build-sanitize
ASAN_OPTIONS=detect_leaks=1:halt_on_error=1 UBSAN_OPTIONS=halt_on_error=1 \
  meson test -C build-sanitize test_adm_cm_large_coeffs test_feature_extractor --print-errorlogs
```

Both suites pass. No golden assertions are changed.


anematode:
Presumably the wrapping is wrong, and the other impls should be fixed?

Either way it can be folded into the preceding `_mm256_srai_epi32` to save some shifts

lusoris:
Agreed. I replaced the wrap-matching patch in `21e6178`: centre taps now stay 32-bit across scalar, x86 edge, and CUDA; subtraction is 64-bit; SIMD lanes beyond `INT32_MAX` are explicitly zeroed; and the sanitizer test covers bounded scale-0 coefficients. I have updated the title and body to describe the revised fix.

PR #1605, opened 2026-09-19
feature/adm: keep get_best15_from32() inside its defined range on AVX2

lusoris:
`get_best15_from32()` is defined only for values of 32768 and above. Below that, `17 - __builtin_clz(temp)` is zero or negative, so both shifts in

```c
temp = (temp + (1 << (k - 1))) >> k;
```

are undefined, and for `temp == 0` `__builtin_clz` has no defined result at all.

The scalar `adm_decouple_s123()` in `integer_adm.c` and the AVX-512 one in `adm_avx512.c` respect that with a guard at the call site:

```c
uint16_t kh_msb = (abs_oh < (32768) ? abs_oh : get_best15_from32(abs_oh, &kh_shift));
```

`adm_decouple_s123_avx2()` cannot branch per lane, so it runs the function on all eight lanes of `abs_oh`, `abs_ov` and `abs_od` and selects afterwards with `_mm256_cmpgt_epi32(const_32768_epi32, abs_oh_epi32)`. The lanes below 32768 are discarded, but they are evaluated first. Real content reaches this on every frame:

```sh
head -c 279936 python/test/resource/yuv/src01_hrc00_576x324.yuv > s1.yuv
head -c 279936 python/test/resource/yuv/src01_hrc01_576x324.yuv > s2.yuv
meson setup build-ubsan libvmaf --buildtype=debug -Db_sanitize=undefined \
  -Denable_cuda=false -Denable_float=true -Denable_docs=false
ninja -C build-ubsan
UBSAN_OPTIONS=print_stacktrace=1 build-ubsan/tools/vmaf -r s1.yuv -d s2.yuv \
  -w 576 -h 324 -p 420 -b 8 --feature adm --no_prediction --cpumask 16 -q -o /dev/null
# adm_avx2.c:1339:23: runtime error: shift exponent -1 is negative
#     #0 get_best15_from32 ../libvmaf/src/feature/x86/adm_avx2.c:1339
#     #1 adm_decouple_s123_avx2 ../libvmaf/src/feature/x86/adm_avx2.c:1532
#     #2 integer_compute_adm ../libvmaf/src/feature/integer_adm.c:2954
# adm_avx2.c:1339:36: runtime error: shift exponent -4 is negative
```

`--cpumask 16` masks AVX-512 off so that the AVX2 kernels are dispatched; without it the AVX-512 version runs and its call-site guard keeps the function in range.

The fix returns, for those values, what the select already picks for them: the value itself and a shift of 0. `blend(a, b, mask)` is `(mask & a) | (~mask & b)`, the mask is set where `abs_oh` is below 32768, and the two calls there are `blend(abs_oh_epi32, tmp_kh_msb_epi32, mask)` and `blend(_mm256_setzero_si256(), kh_shift_epi32, mask)`, so the guarded lanes carry the same bits they carried before. Doing it inside the function covers all 24 call sites without changing the vector code.

No score changes. Validation against upstream `86da14d0306a138fd3f01319860b905169746516`, x86-64 Linux (AVX-512 host), GCC 16.2.1, Meson 1.12.0:

```sh
meson test -C build-release --print-errorlogs
# 23/23 passed
./build-checkasm/test/checkasm/checkasm --test=adm
# checkasm: all 265 tests passed
```

The three reference pairs (`src01_hrc00`/`src01_hrc01` at 576x324, and `checkerboard_1920_1080_10_3_0_0` against `_1_0` and `_10_0`) give JSON identical to the unpatched build apart from the `fps` field, under the default dispatch, `--cpumask 16` and `--cpumask -1`; pooled `vmaf_v0.6.1` means are 76.667831, 35.068667 and 7.985899 in all nine runs. On the clip above the AVX2 and the scalar `integer_adm*` values are already equal before the change, which is the same thing said per-feature.

After the change the run above reports nothing from `adm_avx2.c`. It still reports `variable length array bound evaluates to non-positive value 0` at `vmaf.c:220`, which comes from `--no_prediction` and is unrelated.

Overlap: #1494 rewrites `adm_cm()` and its AVX2 and AVX-512 versions in the same three files. It does not touch `get_best15_from32()` or `adm_decouple_s123_avx2()`, so the two should not collide, though both change `adm_avx2.c`.


anematode:
btw this will be fixed by https://github.com/Netflix/vmaf/pull/1584

lusoris:
Thanks for the pointer. I confirmed #1584 covers this undefined-shift path, so this PR is superseded and I am closing it to avoid duplicate review work.

PR #1609, opened 2026-09-21
avx2: improve blend operations in adm

anematode:
Refactor more operations to use `blend` and use `_mm256_blendv_epi8` for that.

### DIFF
--- a/libvmaf/src/feature/x86/adm_avx2.c
+++ b/libvmaf/src/feature/x86/adm_avx2.c
@@ -684,6 +684,11 @@ angle_flag = ((((float)ot_dp / 4096.0) >= 0.0f) && \
                     cos_1deg_sq * ((float)o_mag_sq / 4096.0) * ((float)t_mag_sq / 4096.0))); \
 } \
 
+static inline __m256i blend(__m256i a, __m256i b, __m256i mask)
+{
+    return _mm256_blendv_epi8(b, a, mask);
+}
+
 void adm_decouple_avx2(AdmBuffer *buf, int w, int h, int stride,
                          double adm_enhn_gain_limit, int32_t* adm_div_lookup)
 {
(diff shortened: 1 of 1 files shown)


lusoris:
**Verdict:** equivalent, but stale base, no evidence, and no measurable gain where I measured. Rebase, benchmark on the machine where it helps, and lead with the part of the diff that is actually good.

**Equivalence:** I checked every `blend()` call site in the file. All masks are full-lane compare results (`cmpgt_epi32/64`, `cmpeq` xor −1, `cmp_ps`/`cmp_pd` casts), so switching the helper to `_mm256_blendv_epi8(b, a, mask)` is safe. `_mm256_blend_epi32(lo, slli_epi64(hi, 32), 0xAA)` for the lo/hi interleave is right. Dropping the redundant `& mask_rst_*` before the blend is right.

**Problems:**

1. **Base is `f85a85369`**, eight commits behind master, from before checkasm was integrated (which is why the branch shows "no checks") and before 03b5562c5 fixed an out-of-bounds write in `adm_decouple_avx2`, the function this PR edits. Rebase. I cherry-picked both commits onto current master to test: builds, stock checkasm passes (the ±8000 int16 bands do cover these paths), AVX2 vs pure C identical, identical to master.
2. **Bench,** rebased, i9-12900K P-core, `adm_decouple_64x48_avx2`, cycles: master 10321 / 10333 (two seeds), this PR 10526 / 10152. Inside the ±2% noise. `vpblendvb ymm` is 2 uops on most Intel cores and 1 uop on Zen, so this may well be a Zen-only win. Measure it there and put the number in, or call the PR a readability change.
3. The genuinely good part is the `and(0xFFFFFFFF) | slli` → `blend_epi32 0xAA` replacement (3 ops → 1). Lead with that in the description.
4. Textual conflict with #1584 (both edit around the `blend` helper). Say which merges first.
5. Two commits, one is "Fix indentation". Squash. Subject: `feature/adm/x86: use blendv/blend_epi32 for mask selects in adm_decouple_avx2`.

